PR #954, opened 2014-05-17
Add asciimath support for docbook5

pepijnve:
Add an Asciimath parser and MathML generator. This enables conversion of asciimath expressions to MathML in the docbook5 converter.

### DIFF
--- a/lib/asciidoctor/converter/docbook5.rb
+++ b/lib/asciidoctor/converter/docbook5.rb
@@ -256,9 +256,11 @@ module Asciidoctor
       if node.style == 'latexmath'
         equation_data = %(<alt><![CDATA[#{equation}]]></alt>
 <mediaobject><textobject><phrase></phrase></textobject></mediaobject>)
-      # asciimath
+      elsif node.style == 'asciimath' && (!(defined? @asciimath_loaded) ?
+          (@asciimath_loaded = Helpers.require_library 'asciimath', true, :warn) : @asciimath_loaded)
+        equation_data = ::AsciiMath.parse(equation).to_mathml 'mml:', 'xmlns:mml' => 'http://www.w3.org/1998/Math/MathML'
       else
-        # DocBook backends can't handle AsciiMath, so output raw expression in text object
+        # Unsupported math style, so output raw expression in text object
         equation_data = %(<mediaobject><textobject><phrase><![CDATA[#{equation}]]></phrase></textobject></mediaobject>)
       end
       if node.title?
--- a/test/blocks_test.rb
+++ b/test/blocks_test.rb
@@ -1285,11 +1285,9 @@ x+b/(2a)<+-sqrt((b^2)/(4a^2)-c/a)
 ++++
       EOS
 
-      expect = <<-'EOS'
-<informalequation>
-<mediaobject><textobject><phrase><![CDATA[x+b/(2a)<+-sqrt((b^2)/(4a^2)-c/a)]]></phrase></textobject></mediaobject>
-</informalequation>
-      EOS
+      expect = %(<informalequation>
+<mml:math xmlns:mml="http://www.w3.org/1998/Math/MathML"><mml:mi>x</mml:mi><mml:mo>+</mml:mo><mml:mfrac><mml:mi>b</mml:mi><mml:mrow><mml:mn>2</mml:mn><mml:mi>a</mml:mi></mml:mrow></mml:mfrac><mml:mo>&#x003C;</mml:mo><mml:mo>&#x00B1;</mml:mo><mml:msqrt><mml:mrow><mml:mfrac><mml:msup><mml:mi>b</mml:mi><mml:mn>2</mml:mn></mml:msup><mml:mrow><mml:mn>4</mml:mn><mml:msup><mml:mi>a</mml:mi><mml:mn>2</mml:mn></mml:msup></mml:mrow></mml:mfrac><mml:mo>&#x2212;</mml:mo><mml:mfrac><mml:mi>c</mml:mi><mml:mi>a</mml:mi></mml:mfrac></mml:mrow></mml:msqrt></mml:math>
+</informalequation>)
 
       output = render_embedded_string input, :backend => :docbook
       assert_equal expect.strip, output.strip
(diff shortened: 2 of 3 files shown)


jackrabbit128:
Is there any progress on this?
Would you like some help?

mojavelinux:
@jackrabbit128 sure! There's now a release of asciimath. So the next step is simply calling it from the converter code as we had discussed. See if you can get it to run and we'll sort out the best way to merge it into core.

pepijnve:
@jackrabbit128 the asciimath parser and mathml generator could use more test cases (see https://github.com/pepijnve/asciimath/blob/master/test/parser_spec.rb). I seem to remember there was still an issue in matrix recognition in the parser somewhere but maybe I fixed that already in the meantime.

I only glanced at the changes in this PR briefly just now. Looks like this should still work as is.

So basically what's left to be done is testing. If you have some time to spare and can integrate this PR locally and try it out with your documents that would be really helpful. If you run into any asciimath constructs that aren't converted correctly please log issues in https://github.com/pepijnve/asciimath.

@mojavelinux I seem to remember we talked about putting asciimath under the asciidoctor umbrella as well. Do you still want to do that or are you ok with it being a 'third-party' dependency.

mojavelinux:
> @mojavelinux I seem to remember we talked about putting asciimath under the asciidoctor umbrella as well. Do you still want to do that or are you ok with it being a 'third-party' dependency.

I'm comfortable either way. If you would like it inside of asciidoctor, I'd be happy to move it there. But that's not a requirement in order to make use of it. Just let me know what you prefer.

jackrabbit128:
I am way too uncomfortable with git to be sure that I did everything correctly. However, it looks like it *almost* works.

`lib/asciidoctor/converter/docbook5.rb` line 265 invokes `Asciimath.parse` which does not exist as the class is now called `AsciiMath`. After making that change, the generated DocBook 5 looks good.

pepijnve:
@jackrabbit128 thanks for trying it out. Class name spelling has been corrected in the PR.

test/blocks_test.rb
```diff
@@ -1256,11 +1256,20 @@ def names
 ++++
       EOS
 
-      expect = <<-'EOS'
+      expect = begin
+        gem 'asciimath', '~> 1.0.0.a'
```
mojavelinux: Can you upgrade this to 1.0.0.preview.1? Though I'd be okay with adding it as a development dependency so we can run the test without the begin/rescue.

lib/asciidoctor/converter/docbook5.rb
```diff
@@ -245,6 +245,13 @@ def literal node
       end
     end
 
+    begin
+      require 'asciimath'
```
mojavelinux: Could you change this line to:

```
require 'asciimath' unless defined? ::AsciiMath
```

That way it doesn't attempt to go down the require hole when we've already got it.

pepijnve:
Review comments applied

mojavelinux:
@pepijnve I sent a pull request to your pull request which updates the following:

* lazy load the asciimath gem (this took awhile to find the optimal logic, but I think I finally got it)
* add the asciimath gem as a development dependency

Aside from that, there are just a few other things to do before we can merge.

1. Release a non-prerelease of the gem (this will also publish the name change from Asciimath to AsciiMath); perhaps just call it 1.0.0?
2. Add a git tag that corresponds to the asciimath release :)
2. Squash the commits (note that your last few commits don't have your GitHub e-mail address, so you may want to fix your git configuration)

mojavelinux:
Btw, while this PR has been going on, `Helpers.require_library` got a "try" feature, so we're able to use that now.

pepijnve:
@mojavelinux I'll try to make some time to clean this up this evening. Does GitHub handle history rewrites well? In other words, if I shuffle things around, but keep the pepijnve/asciimath branch, will the PR update itself even if the new branch ends up being unrelated to the previous one? In other words, can I squash and rebase without making things more complicated for you?

mojavelinux:
The trick (at least what I do) is just force push to the same branch over and over, which stays in sync with the PR (because the PR is just a pointer to the branch).

mojavelinux:
That's the reason I always name my branches `issue-<xyz>`. Then I know that branch is tied directly to that PR (though the PR gets its own number, but that's a different story).

mojavelinux:
We have to wait on preview.2 because preview.1 uses the classname Asciimath. That's why the test fails. But I'm okay to wait until you've got that published. I still have a few other things to wrap up for 1.5.3, so it's going to take a few days.

pepijnve:
I've reverted a couple of things to get the tests passing again. I'll switch it back to 1.0.0 and AsciiMath once I get the gem published.

mojavelinux:
:+1:

pepijnve:
@mojavelinux There you go. Nice and tidy! :)

mojavelinux:
Super!

Just one change remaining. Could you disable the test for Ruby 1.8 since the asciimath gem doesn't support with Ruby 1.8. Change the `end` that closes the test to:

```ruby
end if ::RUBY_MIN_VERSION_1_9
```

Then we should be good to go.

pepijnve:
Lack of 1.8 support is not intentional; I simply never tested with it. I don't think I have a 1.8 VM installed even. :)
I'll have a look at getting the asciimath code to run under 1.8 instead. Shouldn't be hard to do. I don't think it has any advanced requirements.

pepijnve:
Ok, just tried to install MRI 1.8.7 but rbenv is complaining about gcc and my homebrew install is still broken since upgrading to El Capitan. I'll do the 1.8 support tomorrow and add the `end if` in the meantime.

mojavelinux:
I'd say that support for Ruby 1.8 in asciimath isn't necessary. We are dropping 1.8 support from core anyway, just not in this release, hence the need for the interim conditional. Ruby 1.8 is basically out of usage at this point, and no other parts of Asciidoctor support it (just core). We're dropping Ruby 1.8 formally in 1.6.0.

mojavelinux:
I wish we had already dropped support for Ruby 1.8, but I just didn't want to get distracted with that in light of all the other work I needed to do in core for 1.5.3.

pepijnve:
Just did some tests and the only thing I had to change was replacing `require_relative` with `require File.join(File.dirname(__FILE__), ...)`. I've made that change, built asciimath 1.0.1 and removed the Ruby 1.9 conditional from the test. Should the dev dependency in the gem spec be changed to `~> 1.0.1` as well?

mojavelinux:
Cool! Yes please.

pepijnve:
All done. With that I think we're all set.

mojavelinux:
Fantastic! Thanks for all your work on this. I know it was a long time coming, but we got there!

mojavelinux:
:beers:

PR #965, opened 2014-05-25
Allow icons/images to be used for list item bullets [wip]

### LINKED ISSUE #957: Allow icons or glyphs to be used for list item bullets
If font icons are enabled, allow the icons to be used for list item bullets. The same mechanism should also allow a unicode char to be specified.

The syntax for an icon might be:

```
[marker=icon-check']
* one
* two
* three
```

The `icon-` (or perhaps `icon:`) prefix is how we know to use the icon font.

The syntax for a glyph might be:

```
[marker=✓]
* one
* two
* three
```

If you think it's more clear, we could name the attribute `bullet` instead of marker.

ggrossetie:
This is a work in progress.
Resolves #957

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -817,6 +817,15 @@ Your browser does not support the audio tag.
       div_classes = ['ulist', node.style, node.role].compact
       marker_checked = nil
       marker_unchecked = nil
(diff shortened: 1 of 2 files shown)


Mogztter:
@mojavelinux Do we still want to include this feature in core ? Maybe we could implement it as an extension ?

Issue #1040, opened 2014-07-28
Integrate Rouge for source highlighting


mojavelinux:
Integrate the Rouge library as an alternate or replacement syntax highlighting library to Pygments.rb. Rouge is a port of Pygments to Ruby. The current Pygments integration calls Pygments by way of an interprocess communication bridge. This bridge doesn't work in AsciidoctorJ.

If we integrate Rouge and verify that it provides the same experience as Pygments, we can make it a drop-in replacement. If there are big enough differences, we can simply recognize it as an additional syntax-highlighter option named "rouge".

Rouge: https://github.com/jneen/rouge

AlexanderZobkov:
What is status of integration of Rouge?
Looking on the following line: https://github.com/asciidoctor/asciidoctor-maven-examples/blob/master/asciidoctor-pdf-example/pom.xml#L61-L63
it has been already integrated and this particular issue needs to be closed.

AlexanderZobkov:
Probably asciidoc documentation needs to be updated as well. Currently the documentation mention support for pygments and coderay only.

abelsromero:
Documentation is partially accurate and probably needs a review.
The thing is that not all source highlighters work on all backends and integration layers.
For instance rouge only works for PDF with asciidoctor-pdf, and pygments does not work in Java.

I always check the [table here](https://github.com/asciidoctor/asciidoctor.org/blob/c9d1b0c9dadafe04def8f8fe21d06541b177f9da/docs/_includes/src.adoc) for the details, but maybe we should add a new column with the supported backends.

mojavelinux:
That's exactly correct @abelsromero. Rouge was integrated into Asciidoctor PDF first as a trial. The alpha status of Asciidoctor PDF presented us with an opportunity to experiment with this library before committing to it. The results were good. Rouge works great! That's the point when we decided to integrate it into core. That part is not yet done.

Pygments is going to continue to be a problem because it just doesn't play well with JRuby (and other Ruby implementations). CodeRay is mediocre at best. That's why I think Rouge is the future.

mojavelinux:
I think when we integrate Rouge into core, we're going to drop CodeRay (and alias coderay to rouge). We have to keep Pygments because we have a handful of faithful users of it (and because it really is the most comprehensive), but we won't recommend it.

AlexanderZobkov:
[...]

PR #1063, opened 2014-08-13
added Prism.js syntax highlighting option

robfletcher:
I'd like to add some test coverage for this but I couldn't find where the existing highlighting is tested other than a check of the JS / CSS includes for highlight.js in blocks_test.rb

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -547,7 +547,7 @@ Your browser does not support the audio tag.
           pre_class = %( class="CodeRay highlight#{nowrap ? ' nowrap' : nil}")
         when 'pygments'
           pre_class = %( class="pygments highlight#{nowrap ? ' nowrap' : nil}")
-        when 'highlightjs', 'highlight.js'
+        when 'highlight.js', 'prism.js'
           pre_class = %( class="highlightjs highlight#{nowrap ? ' nowrap' : nil}")
           code_attrs = %( class="language-#{language}"#{code_attrs}) if language
         when 'prettify'


Mogztter:
Prism.js looks great and can support line numbers: http://prismjs.com/plugins/line-numbers/
Since Asciidoctor provides callouts this may not be important but I do think line numbers is popular demand.

Another cool "feature" is that by default Prism.js set a max-height. When you include long source code (not really a best practice though) you definitely want to be able to read the callout for a particular piece of code without scrolling back and forth.
@mojavelinux Do you think we should support a max-height attribute on source block or this is a "style preference" and should be defined in css ?

In my opinion:

 * `highlightjs` is the most common option when converting to HTML5 (and `prismjs` could be a nice alternative)
 * `rouge` will be a great successor/substitute to `pygments` and will integrate nicely with asciidoctor-pdf ?

What are the pros and cons of `coderay` and `prettify` ?

Sorry to start a discussion on a pull request but I didn't find an open issue for Prism.js :wink: 
Anyway I'd love to use Prism.js as a syntax highlighter.

mojavelinux:
> Prism.js looks great and can support line numbers

All the syntax highlighters we currently support can support line numbers too, so I don't see this as a benefit of Prism.js but rather a minimum requirement.

> Another cool "feature" is that by default Prism.js set a max-height.
> Do you think we should support a max-height attribute on source block?

We could easily support that as a general feature on source blocks. I don't see this as something that should be specific to Prism.js. To answer your question, yes, we should consider it! Let's track that in a separate issue.

> rouge will be a great successor/substitute to pygments and will integrate nicely with asciidoctor-pdf ?

I see Rouge more as a replacement for (the stalled) CodeRay and, furthermore, close enough to Pygments that it should allow easy migration for those users as well.

#1040 

> What are the pros and cons of coderay?

Now that Rouge exists, there are no pros for CodeRay. Rouge wins hands down.

> What are the pros and cons of prettify ?

Prettify is probably the most accurate source highlighter that exists (maybe second to Pygments), which is why we keep it. We have a lot of JBoss projects that rely on prettify for its accuracy.

> Anyway I'd love to use Prism.js as a syntax highlighter.

I'm really concerned about adding another source highlighter because the CSS permutations we have to support is already astronomic. Unless there's a clear reason why Prism.js is better than highlight.js (or prettify), I'd like to stick with the ones we have. If it does a better job, then that gives us good reason for adding it.

PR #1152, opened 2014-10-19
resolves #651 Manpage wip

### LINKED ISSUE #651: Add built-in man page backend
Add a built-in backend to generate a man page.

Use the erb templates from the asciidoctor-backends repository as a reference point:

https://github.com/asciidoctor/asciidoctor-backends/tree/master/erb/manpage

DavidGamba:
This is just a copy of the html backend with only enough edits to build
the asciidoctor manpage as closely as needed to the output of the a2x
command.

This commit doesn't clear any unused sections in the hopes that if you
need them, you do what I did and modify them in a per needed basis.

Should I clean anything I didn't update from the HTML backend?


DavidGamba:
What is the timeline for 1.5.3?

- I will create a proper pull request for the converter.
- I will create another issue for adding doctest, which is what I used to test, but I don't know if that will be merged right away because some dependencies needed to be updated and doctest doesn't support ruby 1.8.7 so I'll let you decide when to merge that part.

About the spacing, I will submit what I did but I don't think it is perfect. Unfortunately I work on this on my spare time only. 
The one thing I couldn't totally figure out is how to use the `manify` method. The problem that I have is that, for example, when I _manify_ the _content_ in the `block_paragraph`, that _manifies_ the content of the other nested asciidoc blocks. That causes some problems and have to watch for exceptions to avoid issues.

For example, _manifying_ the dot `.` char: Dot is used for troff commands and it is difficult to distinguish between user content and troff nested commands.

I didn't do too much on that area but tried to at least provide a good scaffolding to start working on it (and a working one). I also finished with the manpage tables which is something I wanted to do. I also finished most other blocks as well because I want the converted not to skip any of the original content, for example, showing the video block as a link with the metadata of start and finish.

mojavelinux:
> What is the timeline for 1.5.3?

I'm trying to make it happen this month. Let's just say the end of the month to be conservative.

mojavelinux:
If you can get the pull request to the state where you're happy with it in a few weeks, then I'd say this is probably good enough to make part of core and we can iterate on it once it's merged in with smaller improvements and fixes.

DavidGamba:
I consolidated the pull request into https://github.com/asciidoctor/asciidoctor/pull/1291. I forgot to close this one.

mojavelinux:
Nice! :+1: I'm hoping to look at it soon, perhaps at one of the conference hack sessions. \o/

PR #1189, opened 2014-11-10
Allow docinfo to be in separate directory from content

### LINKED ISSUE #511: allow docinfo to be in separate directory from content
Would be nice to specify an independent directory to draw `docinfo.[ext]` files from in the same vein as `template_dir`. co-mingling `*.adoc` with `docinfo.[ext]` seems counter to the core philosophies.

ggrossetie:
Resolves #511


stefanct:
I was about to request this because I could not find it in the documentation (please add if it is really not there) but it was already there :100:

mojavelinux:
I filed https://github.com/asciidoctor/asciidoctor.org/issues/604 to address the documentation gap.

PR #1213, opened 2014-11-23
upgrade Nokogiri to 1.6.x (except on Ruby 1.8)

mojavelinux:


Mogztter:
Also nokogiri 1.5 doesn't work in Ruby 2.x on Windows :disappointed:

mojavelinux:
We'll need to do some more testing to move this forward to figure out how
to get all the CI builds happy.

Mogztter:
Ruby 1.8 is dead no ? :grinning:

mojavelinux:
Correct. At this point, dropping 1.8 is just a formality. I just don't want to do it on a minor release. The plan is to drop it for 1.6.0...though I'll reconsider after we get beyond 1.5.3. I not only want to drop 1.8, but also the crufty 1.8 syntax that we're using.

Mogztter:
Some time ago, I've subscribed to Gemnasium and today I receive a security alert about Nokogiri: https://gemnasium.com/asciidoctor/asciidoctor/alerts

> MULTIPLE VULNERABILITIES IN LIBXML2, LIBXSLT

Time to upgrade ? :smile:

mojavelinux:
It's definitely time to upgrade, though since we only use Nokogiri for tests (at least in core) I don't think these vulnerabilities are a huge concern for us. We do use Nokogiri in Asciidoctor EPUB3, so we have to keep our eye on that component.

I'm going to send a message to the list now about switching master to 1.6.x and dropping Ruby 1.8 support.

mojavelinux:
I've finally figured out how to do this upgrade in a way that works with Ruby 1.8. It also seems like Nokogiri has addressed the bugs that were causing problems in JRuby.

PR #1223, opened 2014-11-27
resolves #1195 implementation of epigraphs using roles

### LINKED ISSUE #1195: Add support for epigraphs
DocBook provides an epigraph element, that it describes as "a short inscription, often a quotation or poem, set at the beginning of a document or component." I like to use epigraphs, and notable works like _The Practice of Programming_ use them as well.

It would be nice if Asciidoctor provided support for epigraphs. We should support `literallayout` elements (verse-style blocks) as well as `para`/`simpara` elements (prose blocks), and support attributions as well.

bk2204:
This adds support for epigraphs for quote and verse blocks by using the "epigraph" role.  I couldn't find any tests for the DocBook output of the existing quote and verse blocks, so I added some to make sure I didn't regress anything with my changes.

The quote and verse functions were almost identical, so they were refactored into a single function that did most of the work.  The tests pass after each commit.

Fixes #1195.

### DIFF
--- a/lib/asciidoctor/converter/docbook5.rb
+++ b/lib/asciidoctor/converter/docbook5.rb
@@ -766,8 +766,9 @@ module Asciidoctor
     end
 
     def quote_like node, content
+      tag_name = (node.has_role? 'epigraph') ? 'epigraph' : 'blockquote'
       result = []
-      result << %(<blockquote#{common_attributes node.id, node.role, node.reftext}>)
+      result << %(<#{tag_name}#{common_attributes node.id, node.role, node.reftext}>)
       result << %(<title>#{node.title}</title>) if node.title?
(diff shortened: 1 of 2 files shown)


bk2204:
I've rebased the changes so that they apply cleanly to `master`.

mojavelinux:
Thanks for updating. I'll have a look at getting this in during the 1.5.7 cycle.

mojavelinux:
I incorporated your changes into #2664. I added two commits in front of the change you made, then added your change and tests as the third commit.

I renamed the method `quote_like` to `blockquote_tag` to be consistent with the other tag helper methods in the converter. I also moved the content to a block argument since it's important that the node content is converted after the node title.

Let me know if this looks acceptable to you.

Issue #1140, opened 2014-10-11
Support options="breakable" and options="unbreakable" for tables in Docbook backends.


rockyallen:
These options tell FOP to allow tables to split across a page break or not: see

http://discuss.asciidoctor.org/Preventing-tables-from-breaking-across-pages-td2182.html#a2337

Changes required: add 

if (node.option? 'breakable')
  result << %(<?dbfo keep-together="auto"?>)
end 

and similar for unbreakable to docbook5.rb and docbook45.rb. Possibly update user guide and tests.

mojavelinux:
Keep in mind you can write this using the following syntax:

```
[%unbreakable, format=csv]
|===
customers.csv[]
|===
```

That's how it should be documented so we are recommending the most concise way :)

PR #1279, opened 2015-02-17
Fix for issue 1140, 2 files changed

rockyallen:
#1140 Support breakable and unbreakable for tables in docbook 5 and 4.5 backends.
Translated into dbfo processing instructions. Unbreakable is included for completeness, even though it is the default. Only needed to change the docbook5.rb converter - I think the 4.5 backend delagates to it (at least its tests pass without making any changes to docbook45.rb).
If this change is accepted, I will update the docs (I assume I can't do it in one pull request because the docs are in a different repository).

### DIFF
--- a/lib/asciidoctor/converter/docbook5.rb
+++ b/lib/asciidoctor/converter/docbook5.rb
@@ -376,10 +376,9 @@ module Asciidoctor
       result << %(<#{tag_name = node.title? ? 'table' : 'informaltable'}#{common_attributes node.id, node.role, node.reftext}#{pgwide_attribute} frame="#{node.attr 'frame', 'all'}" rowsep="#{['none', 'cols'].include?(node.attr 'grid') ? 0 : 1}" colsep="#{['none', 'rows'].include?(node.attr 'grid') ? 0 : 1}">)
 
       if (node.option? 'breakable')
-        result << %(<?dbfo keep-together="auto"?>)
-      end 
-      else if (node.option? 'unbreakable')
-        result << %(<?dbfo keep-together="always"?>)
+        result << '<?dbfo keep-together="auto"?>'
+      elsif (node.option? 'unbreakable')
+        result << '<?dbfo keep-together="always"?>'
       end 
 
       result << %(<title>#{node.title}</title>) if tag_name == 'table'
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/converter/docbook5.rb
```diff
@@ -374,6 +374,14 @@ def table node
       result = []
       pgwide_attribute = (node.option? 'pgwide') ? ' pgwide="1"' : nil
       result << %(<#{tag_name = node.title? ? 'table' : 'informaltable'}#{common_attributes node.id, node.role, node.reftext}#{pgwide_attribute} frame="#{node.attr 'frame', 'all'}" rowsep="#{['none', 'cols'].include?(node.attr 'grid') ? 0 : 1}" colsep="#{['none', 'rows'].include?(node.attr 'grid') ? 0 : 1}">)
+
+      if (node.option? 'breakable')
+        result << %(<?dbfo keep-together="auto"?>)
+      end
```
mojavelinux: Could you remove this `end` and use `elseif` on the next line instead of `else if`? You can see similar examples elsewhere in this file.

lib/asciidoctor/converter/docbook5.rb
```diff
@@ -374,6 +374,14 @@ def table node
       result = []
       pgwide_attribute = (node.option? 'pgwide') ? ' pgwide="1"' : nil
       result << %(<#{tag_name = node.title? ? 'table' : 'informaltable'}#{common_attributes node.id, node.role, node.reftext}#{pgwide_attribute} frame="#{node.attr 'frame', 'all'}" rowsep="#{['none', 'cols'].include?(node.attr 'grid') ? 0 : 1}" colsep="#{['none', 'rows'].include?(node.attr 'grid') ? 0 : 1}">)
+
+      if (node.option? 'breakable')
+        result << %(<?dbfo keep-together="auto"?>)
```
mojavelinux: Can you switch to using single quotes around this string? The `%()` is only needed when using interpolation. Same for the other dbfo string.

mojavelinux:
Overall, it looks good! I added two code review comments. If you can address those and force push and update to your branch, I'll merge it in!

>  Only needed to change the docbook5.rb converter - I think the 4.5 backend delagates to it 

Correct.

> I assume I can't do it in one pull request because the docs are in a different repository

Correct.

mojavelinux:
> Only needed to change the docbook5.rb converter - I think the 4.5 backend delagates to it 

Correct.

mojavelinux:
Merged as a1fdb78cc319f695c0a5cd1029c5188e12d0ec0a.

Thanks for this pull request @rockyallen!

PR #1304, opened 2015-03-29
Parse version alone in header as version, not date.

### LINKED ISSUE #790: parse revision line with revision only
If the revision line begins with a "v" and has no comma, treat that as the revision number, not the date.

bk2204:
If a version alone (without a date) is given in the document header, treat it as a version, not a date. The criterion used to determine this is whether the component begins with a "v"; dates hopefully will not begin that way.

Note that the tests use an attribute expansion to verify that the revnumber attribute is set, as the DocBook backend doesn't generate revision entries unless there's a revdate, and there are tests for that, so I didn't want to change it. I can, however, if you think that's a better idea.

This patch fixes #790.

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -1827,8 +1827,13 @@ class Parser
       if reader.has_more_lines? && !reader.next_line_empty?
         rev_line = reader.read_line 
         if (match = RevisionInfoLineRx.match(rev_line))
-          rev_metadata['revdate'] = match[2].strip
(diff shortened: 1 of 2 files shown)


mojavelinux:
Thanks @bk2204! Nice work. Merged as f0a31df265627a2cb86dc46249e9b698a7ce6da3.

I made some slight revisions, but used the same logic. I've updated the documentation to explain that if the revision line contains only the revision number, the revision number must begin with a "v". I think it's safe to assume that dates aren't going to begin with a lowercase "v". (Of course, it's always possible to use explicit attributes if someone has a strange date format).

PR #1310, opened 2015-04-06
Avoid direct access to config instance variable

robertpanzer:
There is currently a false positive in the AsciidoctorJ tests.
The tests for the InlineMacroProcessor should fail but they don't.
Depending on the regexp the ManPageProcessor should be applied or not but it always applies.
https://github.com/asciidoctor/asciidoctorj/blob/master/asciidoctorj-core/src/test/java/org/asciidoctor/extension/WhenJavaExtensionIsRegistered.java#L554

The reason is that Asciidoctor does not see the regexp in the options because it directly accesses the config instance variable.
Thereby it bypasses the possibility of the extension proxy to make the config of the instance visible.
The PR changes the access to the config by not accessing the config ivar directly but by calling the accessor that invokes the extension proxy and makes the correct configuration visible.

### DIFF
--- a/lib/asciidoctor/extensions.rb
+++ b/lib/asciidoctor/extensions.rb
@@ -425,7 +425,7 @@ module Extensions
   class InlineMacroProcessor < MacroProcessor
     def initialize name, config = {}
       super
-      @config[:regexp] ||= (resolve_regexp @name, @config[:format])
+      self.config[:regexp] ||= (resolve_regexp @name, @config[:format])
     end
 
     def resolve_regexp name, format


robertpanzer:
Maybe it would make sense to replace all direct accesses to @config by calling the accessor.
The same problem could occur in other constellations as well with Java based extensions.
This is not only a problem due to the new extension proxies in the asciidoctorj-1.6.0 branch.
In the previous extension Java classes @lordofthejars also remarked that there are two config maps, one in Ruby, one in Java.

Another option to resolve this could be to make a change to AsciidoctorJ by making the Java extension delegate to the extension proxy which could provide a map that is backed by the Ruby config instance variable.
But I am unsure if this works out.

First priority of this PR was to make my test case pass :smiley:

robertpanzer:
Sorry, stupid idea. 
I think the better solution will be to remove the redundancy completely and make AsciidoctorJ access the Ruby config hash instead of its own HashMap.

mojavelinux:
Are you sure you don't need this change. We could make a change to core to help with the situation?

robertpanzer:
Hi Dan,

I think I found a better way by not intercepting the accessors but using the same config instance in Java. 

I opened https://github.com/asciidoctor/asciidoctorj/pull/287
Would be great if you could review it. 

Best regards 
Robert

Issue #1320, opened 2015-04-30
Embedded Vimeo/YouTube videos don't display


crydalch:
It appears that Asciidoctor is omitting *https* in the link is the cause; when I add that, the videos display properly.

https://gist.github.com/crydalch/6a240344125ece78d73a

Loving Asciidoctor! Thank you for the hard work!

mojavelinux:
I wonder if the situation has changed from what I describe here:

https://github.com/asciidoctor/asciidoctor.org/issues/197

If so, that's encouraging.

crydalch:
Cool, it looks like it works now! I just served the page with a Ruby server (like you showed), and the videos all show up and seem to play just fine:

![screenshot-area-2015-04-30-115156](https://cloud.githubusercontent.com/assets/458856/7420199/653b82aa-ef2f-11e4-9448-91a972c98666.png)

mojavelinux:
Cool! It looks like YouTube now works properly for local files without having to use a web server! This change seemed to happen when they dropped Flash in favor of HTML5 (as reported in this new post http://www.theverge.com/2015/1/27/7926001/youtube-drops-flash-for-html5-video-default).

I was going to use the `asset-uri-scheme` as the prefix instead of hardcoding `https` so it's still possible to control the scheme in some way.

mojavelinux:
Merged!

crydalch:
Awesome, that's great, thanks so much!

PR #1321, opened 2015-04-30
Add https to video

crydalch:
I'm not a Ruby programmer by any means, but this looked like a simple fix for issue #1320  so I decided to give it a go!

I've updated the Html5 converter to include *https* in the iframe links for Vimeo and YouTube videos, and updated blocks_test.rb to make sure those links are correct.

Thanks!

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -918,7 +918,7 @@ Your browser does not support the audio tag.
         loop_param = (node.option? 'loop') ? %(#{delimiter}loop=1) : nil
         %(<div#{id_attribute}#{class_attribute}>#{title_element}
 <div class="content">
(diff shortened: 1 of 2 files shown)


mojavelinux:
Thanks for sending this patch. Unfortunately, I don't think this actually solves the problem as we identified in this issue:

https://github.com/asciidoctor/asciidoctor.org/issues/197

Even if you add the `https` prefix, the YouTube script still embeds the video without the scheme...at least, when I last tested it. Perhaps it's different for HTML5 video model. I'd be glad to test it again because not being able to view videos in local documents is very annoying.

crydalch:
I just saw this comment, sorry I missed it a few minutes ago; I ran a little Ruby server, and the YouTube and Vimeo links load and play just fine, in addition to it working on static docs too. I guess YouTube's changes allow this to work now? 

I agree, it's a pain not seeing videos on static pages!

Thanks for taking a look a this! Sorry if it's a bit confusing erratic :)

mojavelinux:
>  I guess YouTube's changes allow this to work now?

I hope so! I'll do some testing today and if this is now fixed in YouTube, then we can move forward with your change. (I've got one other idea I want to test too that would allow this to be controlled by the document settings).

crydalch:
Great, thanks!

mojavelinux:
Merged!

mojavelinux:
Thanks!

PR #1350, opened 2015-05-21
supporting to configure separator for title through [separator=...]

rmannibucau:


mojavelinux:
Nice start! I sent a review as a pull request. I preserved the design but made some slight modifications to the implementation. See if those changes look good to you.

Note that I am now automatically appending a space to the separator. I think this is the right way to go as it simplifies configuration, deals with a potential trailing whitespace issue and, it enforces a style guideline for writing.

mojavelinux:
Merged as 31b62fbcbeb5122b8d830d3fd49509eb32717821. Thanks!

Note that we'll need to document this in the user manual.

mojavelinux:
For the documentation, here's how this works.

If no separator is specified, the document title is separated into a main title and subtitle using the industry standard, a colon followed by a space.

```
= Main Title: Subtitle
```

You can modify the separator by specifying a separator explicitly using a block attribute on the document title. Note that a space will automatically be appended to the separator value.

```
[separator=::]
= Main Title:: Subtitle
```

You can also set the separator using a document attribute, either in the document or from the API / cli.

```
= Main Title:: Subtitle
:title-separator: ::
```

or

```
asciidoctor -a title-separator=:: document.adoc
```

You can also partition the title from the API in an arbitrary way by calling the `doctitle` method on Document and passing the separator as a value to the `partition` option.

```
title_parts = document.doctitle partition: '::'
puts title_parts.title
puts title_parts.subtitle
```

mojavelinux:
It's also important to note that the separator is searched from the end of the text. Therefore, only the last occurrence of the separator is used for partitioning the title.

mojavelinux:
This feature is now documented at http://asciidoctor.org/docs/user-manual/#subtitle-partitioning

PR #1353, opened 2015-05-23
 - Add a new JS platform (javafx)

rahmanusta:


lib/asciidoctor/opal_ext/file.rb
```diff
@@ -135,6 +135,46 @@ def self.read(path)
         }
         return data;
       )
+    # afx means asciidocfx
+    when 'javafx'
+      %x(
+        var data = '';
+
+        if(afx) {
```
Mogztter: `afx.readAsciidoctorResource` is specific implementation defined in the Asciidoctor FX editor code base ?

lib/asciidoctor/opal_ext/file.rb
```diff
@@ -135,6 +135,46 @@ def self.read(path)
...
+          }
+          catch (e) {
+            throw #{IOError.new `'No such file or directory: ' + path`};
+          }
+          return data;
+        }
+
+        var status = -1;
+        try {
+          var xhr = new XMLHttpRequest();
```
Mogztter: Maybe we should extract this chunk of code to avoid duplication between `browser` and `javafx` ?

lib/asciidoctor/opal_ext/file.rb
rahmanusta: Yes this is Asciidoc FX spesific implementation.

lib/asciidoctor/opal_ext/file.rb
rahmanusta: I tried to implement one method for duplicated implementation for XHR request. But, I couldnt do it because of my lower experience in ruby.

lib/asciidoctor/opal_ext.rb
```diff
@@ -7,6 +7,10 @@
   // or we can check for document
   //else if (typeof document !== 'undefined' && document.nodeType) {
     value = 'browser';
+
+    if(typeof navigator != 'undefined' && navigator.appVersion)
+       if(navigator.appVersion.indexOf('JavaFX') != -1)
+          value = 'javafx'
```
jirutka: As you can see, other code in this file has one space after `if` and block is always wrapped in braces. Please reformat it to:

```javascript
if (typeof navigator != 'undefined' && navigator.appVersion) {
  if (navigator.appVersion.indexOf('JavaFX') != -1) {
    value = 'javafx'
  }
}
```

lib/asciidoctor/opal_ext/file.rb
jirutka: The same as [here](https://github.com/asciidoctor/asciidoctor/pull/1353/files#r30944891), `if` should be followed with one space.

rahmanusta: Created a new PR

lib/asciidoctor/opal_ext.rb
rahmanusta: Created a new PR

lib/asciidoctor/opal_ext/file.rb
jirutka: Why another PR? Just fix it with next commit, or rebase your previous commits.

rahmanusta: It seems one PR with 3 commits. Should I rebase it?

mojavelinux:
I'm okay with adding detection of the JavaFX runtime, but I'm not okay with putting AsciiDocFx code directly into Asciidoctor core. You should be able to just load a replacement implementation of the `::File.read` method in the AsciiDocFX environment so it is used instead of the one we provide in core.

Ruby works like CSS in that code loaded later overrides code loaded earlier. You can simply create a new Ruby file with the `::File.read` method in it, implement it how you want it, cross compile it with Opal, then load it after the other JavaScript files in the browser view. Then, voila, you get your own implementation.

First, define the file:

```ruby
class File
  def self.read path
    %x(var data = '';
      try {
        data = afx.readAsciidoctorResource(path, null);
        if (data == 404) {
          throw new Error();
        }
      }
      catch (e) {
        throw #{IOError.new `'No such file or directory: ' + path`};
      }
      return data;)
  end
end
```

Then install opal:

```
$ gem install opal --version 0.6.3
```

Then compile the file to JavaScript:

```
$ opal -c file-read.rb > file-read.js
```

if that doesn't work, try:

```
$ ruby -r tilt `which opal` file-read.rb > file-read.js
```

Finally, you can load that file in the browser. It should work. Let me know if this resolves your issue. Also, let me know if you still want to pursue the "javafx" value for JAVASCRIPT_PLATFORM.

rahmanusta:
I think after that it is not necessary to put new JavaFX platform, also AsciidocFX. I was wondering how to override default File#read method in asciidoctor. Your last message "Ruby works like CSS in that code loaded later overrides code loaded earlier" clarified me. Thanks.

PR #1354, opened 2015-05-23
Allow to read assets when running in Opal

Mogztter:
See https://github.com/asciidoctor/asciidoctor-extensions-lab/pull/32#issuecomment-98092816

Absolute paths starting with `file://` were considered as relative path (in safe mode) and an exception was thrown:
```
SecurityError : asciidoctor: FAILED: : Failed to parse source, Jail is not an absolute path: file:///workspace/asciidoctor-extensions-lab/lib/chart-block-macro
```

In unsafe mode, the method `read_asset` was using `IO.read` (with no equivalent when running in Opal). AFAIK `::IO.read` and `::File.read` are equivalent so I replaced the first one with the latter.

Finally the method `readable` was undefined when running in Opal so I added the method in `opal_ext/file.rb`. I didn't know how to implement the method in a JavaScript environment so for now the method always returns `true` (until we decide how to implement `file?` and `readable?`)

With these changes I can load CSV files with the chart macro (in the Chrome extension!): `chart::sample.csv[line,800,300]`

### DIFF
--- a/lib/asciidoctor/opal_ext/file.rb
+++ b/lib/asciidoctor/opal_ext/file.rb
@@ -95,6 +95,10 @@ class File
     true
   end
   
+  def self.readable?(path)
+    true
+  end
+
   def self.read(path)
     case JAVASCRIPT_PLATFORM
     when 'node'
@@ -142,4 +146,13 @@ class File
       ''
     end
   end
+
+end
+
+class IO
+
+  def self.read(path)
+    File.read(path)
+  end
+
 end


Mogztter:
AppVeyor build was cancelled.

mojavelinux:
I absolutely want to support this, but I want to look closer at how we're doing it.

When we did this with includes, we actually used a guard around a block specifically for Opal. See https://github.com/asciidoctor/asciidoctor/blob/2893d1666060fbb360e717fd503e06f8cc0f6f32/lib/asciidoctor/reader.rb#L834-L843 rather than changing the PathResolver. We should be consistent, whether it means changing that code or changing the code in read_asset. The latter would be my short-term preference.

Also, we should implement `::IO.read` in Opal so we don't have to change these calls (as `::IO.read` is more flexible). Just map it to `::File.read` since it's basically the same method anyway in all the cases that concern JavaScript.

Mogztter:
>Also, we should implement ::IO.read in Opal so we don't have to change these calls (as ::IO.read is more flexible). Just map it to ::File.read since it's basically the same method anyway in all the cases that concern JavaScript.

OK.

> When we did this with includes, we actually used a guard around a block specifically for Opal

This "issue" is not related to Opal. Absolute paths starting with file:// are treated as relative paths (even in Asciidoctor core). Is this the intended behavior ?

mojavelinux:
> Absolute paths starting with file:// are treated as relative paths (even in Asciidoctor core). Is this the intended behavior ?

At the moment, yes. `file://` is only meaningful in the browser, which means it's only meaningful in Asciidoctor.js. I didn't want to change the behavior of core globally (at least, not yet) to accommodate a need that is specific to Asciidoctor.js. That's why I'd like to support Asciidoctor.js with a special condition so that the behavior of core doesn't otherwise change.

...the reason has to do with it being a security issue. I know the security team is going to zero in on a change to that file, so it means we have to really think hard about any attack vectors it could unlock. All that points to just doing the same hack we did for includes in read_asset and we can get out without any fundamental changes to core.

mojavelinux:
> I didn't know how to implement the method in a JavaScript environment so for now the method always returns true

That's the correct behavior for now because we don't have a good way to pre-check if a file exists in the browser environment.

lib/asciidoctor/opal_ext/io.rb
```diff
@@ -0,0 +1,7 @@
+class IO
+
+  def read(path)
```
Mogztter: @mojavelinux Not sure this is the right way to do it ?

Mogztter:
> ...the reason has to do with it being a security issue. I know the security team is going to zero in on a change to that file, so it means we have to really think hard about any attack vectors it could unlock. All that points to just doing the same hack we did for includes in read_asset and we can get out without any fundamental changes to core.

Removing the `file://` protocol in [path_resolver.is_root()](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/path_resolver.rb#L141) was the only solution I found to make it work.
Is that OK if I'm adding a conditional block for Opal in here ? (I think I know the answer...) 

I tried to move that logic in [reader.preprocess_include()](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L863) method but I don't really know how to transform `target` without breaking everything...

Mogztter:
@mojavelinux If you want I can split my contribution in two ? One to implement the missing methods in Opal (File.readable and IO.read) and another one to resolve the issue with absolute paths starting with `file://` ?

Mogztter:
Rebased on master and I've just kept the modification on `opal_ext`.

@mojavelinux Does this look good to you ?

mojavelinux:
Yep, that's good. Though don't we need to require it in lib/asciidoctor/opal_ext.rb? Perhaps it's just better to move it into file.rb since that's kind of already a metafile for all file related things (after all, it already has Kernel.open). Could you move the IO class into lib/asciidoctor/opal_ext/file.rb?

mojavelinux:
>  One to implement the missing methods in Opal (File.readable and IO.read) and another one to resolve the issue with absolute paths starting with file:// ?

:+1:

mojavelinux:
> another one to resolve the issue with absolute paths starting with file:// ?

It would be best if you opened a new issue so that we can focus specifically on that problem. I think that if we add an Opal guard in PathResolve#is_root, that should be sufficient. Also, if you can cite a use case, that should help speed things up.

Mogztter:
> It would be best if you opened a new issue so that we can focus specifically on that problem. I think that if we add an Opal guard in PathResolve#is_root, that should be sufficient. Also, if you can cite a use case, that should help speed things up.

OK I will create a new issue.

> Could you move the IO class into lib/asciidoctor/opal_ext/file.rb?

Sure.

Mogztter:
> Could you move the IO class into lib/asciidoctor/opal_ext/file.rb?

Done.

mojavelinux:
Super! I'll merge this, then we'll start working on getting the opal_ext moved over to Asciidoctor.js.

PR #1369, opened 2015-06-16
Fallback to require with a non relative path

Mogztter:
Allow to require asciidoctor when then `lib/` directory is not relative to the binary.
For instance, Debian package copies binary in `/usr/bin/asciidoctor` and lib/ in `/usr/lib/ruby/vendor_ruby/`


mojavelinux:
Thanks!

PR #1379, opened 2015-06-28
Load script at the end of document

### LINKED ISSUE #1238: Load highlight.js asynchronously
Hey,

Google Speed test thingie recommends to load highlight.js asynchronously to improve the above-the-fold rendering performance of my Asciidoctor site (http://docs.stardog.com/)...I suspect it would say the same for most Ad sites.

Would you consider switching to loading https://cdnjs.cloudflare.com/…x/libs/highlight.js/8.4/highlight.min.js asynchronously?

Mogztter:
See http://www.html5rocks.com/en/tutorials/speed/script-loading/
Resolves #1238

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -79,7 +79,7 @@ module Asciidoctor
         end
       end
 
-      case node.attr 'source-highlighter'
+      case (highlighter = node.attr 'source-highlighter')
       when 'coderay'
         if (node.attr 'coderay-css', 'class') == 'class'
           if linkcss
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/converter/html5.rb
```diff
@@ -237,6 +211,41 @@ def document node
...
         result << '</div>'
       end
+      
+      # Load Javascript at the end of body for performance
+      # See http://www.html5rocks.com/en/tutorials/speed/script-loading/
+      case node.attr 'source-highlighter'
+      when 'highlightjs', 'highlight.js'
+        highlightjs_path = node.attr 'highlightjsdir', %(#{cdn_base}/highlight.js/8.4)
+        result << %(<script src="#{highlightjs_path}/highlight.min.js"></script>
+<script>hljs.initHighlightingOnLoad()</script>)
```
mojavelinux: Now that the script is at the bottom of the page, we can change this method call to `hljs.initHighlighting()`. No need for the "OnLoad" part as we are already loaded :)

mojavelinux:
Thanks @Mogztter!

Should we move the stylesheets for highlight.js and prettify to the bottom of the document as well? I can't seem to figure out what the recommended practice is for that. Technically, the stylesheets are there to style the HTML that is added to the page by the highlighter, so it's not really needed until that point. I think this could also speed up rendering (though probably not by much). WDYT?

mojavelinux:
One more very minor optimization (more of a style change). Could you cache the result of the `source-highlighter` lookup as follows:

```
case (highlighter = node.attr 'source-highlighter')
```

Then use it later as:

```
case highlighter
```

Thanks!

lib/asciidoctor/converter/html5.rb
```diff
@@ -237,6 +211,41 @@ def document node
...
+      when 'prettify'
+        prettify_path = node.attr 'prettifydir', %(#{cdn_base}/prettify/r298)
+        result << %(<script src="#{prettify_path}/prettify.min.js"></script>
+<script>document.addEventListener('DOMContentLoaded', prettyPrint)</script>)
```
Mogztter: We can do the same here by calling `prettyPrint()` directly since the  DOM content is already loaded ?

Mogztter:
> Should we move the stylesheets for highlight.js and prettify to the bottom of the document as well? I can't seem to figure out what the recommended practice is for that. Technically, the stylesheets are there to style the HTML that is added to the page by the highlighter, so it's not really needed until that point. I think this could also speed up rendering (though probably not by much). WDYT?

Critical/essential stylesheets must be in the `<head>` to avoid "blank screen" but in our case I think you are right we could move highlight.js and prettify to the bottom of the document.

mojavelinux:
Very good. Two more very minor things then we'll be ready to go.

1. Can you drop the trailing `;` from `prettyPrint()`? I agree we can write it as a simple function call.
2. Can you move the `<link>` elements above the `<script>` element so that the CSS is ready when the JavaScript starts running?

mojavelinux:
@Mogztter Do you want me to make these last two changes when merging or do you plan to update the PR?

Mogztter:
@mojavelinux No I will make these two changes today. My bad I don't know why I inverted the script and the stylesheet...

mojavelinux:
:+1: Standing by.

Mogztter:
@mojavelinux should be OK, I rebased my commits

mojavelinux:
Fantastique ! Merci @Mogztter !

mojavelinux:
One small request for future pull requests. Could you put the `resolves #XYZ` in the commit subject? That makes it easier to find the fixed issues in the git log when doing a release. We should probably add this point to the CONTRIBUTING-CODE.adoc doc.

Mogztter:
> One small request for future pull requests. Could you put the resolves #XYZ in the commit subject?

Oui pas de problème :wink:

mojavelinux:
:+1:

Issue #841, opened 2013-12-05
[source,indent=0] does not replace tabs with spaces unless the first character is whitespace


philwebb:
It would be useful if `[source,indent=0]` could be used to replace tabs with spaces in source blocks. Unfortunately this only works if the first character is whitespace:

https://github.com/asciidoctor/asciidoctor/blob/3245bcacd49a812a900925f0816e9601a02fe2d3/lib/asciidoctor/lexer.rb#L2560

In other words:
```
[source,java,indent=0]
[subs="verbatim,quotes"]
---
	// this will work
	// line 2
---
```

```
[source,java,indent=0]
[subs="verbatim,quotes"]
---
// this will not work
	// line 2
---
```

mojavelinux:
We're going to implement this right in Asciidoctor 1.5.3 (in fact, I've already done it in Asciidoctor PDF).

As a side note, it is possible to control the tab size in HTML using CSS for the tabs that do slip through in the interim. Set the tab-size property on the pre element. See https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size

mojavelinux:
This is also related to #1170.

mojavelinux:
With the fix for this issue, tabs are always replaced with spaces in verbatim blocks (but not literal paragraphs) if the `tabsize` attribute is set on the document and > 0. It's no longer required to use the `indent` property to enable tab expansion.

If you want to disable tab expansion, don't set the `tabsize` attribute on the document or set the `tabsize` attribute on the block to 0 (or less).

mojavelinux:
Btw, you can also set `source-indent` to 0 on the document to apply `indent=0` to all source blocks. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/parser.rb#L841-L843

PR #1395, opened 2015-07-05
resolves #1170 and #841 expand tabs properly

### LINKED ISSUE #1170: Set an option to configure number of whitespaces in indent (support tabsize attribute)
When using indent in code block, by default Asciidoctor transforms tabs to 4 spaces. Sometimes you may want to change tabs to 2 spaces for example.

It may be interesting to have a way (document attribute, code block parameter, ...) to set the number of spaces to be used.

mojavelinux:
- introduce tabsize attribute to control size of tab character
- expand tabs in verbatim blocks if tab present and value of tabsize attribute is positive
- rename reset_block_indent! to adjust_indentation!
- only use adjust_indentation! on include if indent is specified
- only expand tabs in included content if indent attribute is set
- use adjust_indentation! to remove left margin in paragraphs marked normal
- don't add indentation to empty lines
- set tab-size property to 4 in default stylesheet
- increase letter spacing in blockquote cite in default stylesheet
- add and fix tests


mojavelinux:
Tests are still needed to check that tabs are properly expanded.

mojavelinux:
I'd also like to set the `tab-size` property in the default stylesheet to 4 for any tabs that slip through.

mojavelinux:
I think this is solid now.

PR #1403, opened 2015-07-08
Adding hl_lines parameter support for pygments source-highlighter.

jalbertcruz:
The HTMLFormater parameter "hl_lines" of pygments is really important in some situations, this pull request includes that.
Now you can write:

```
[source, python, numbered, hl_lines="[2, 3]"]
----
def function():
    for i in range(55):
        print(3)
----
```

and the lines number 2 and 3 will be highlighted.

### DIFF
--- a/lib/asciidoctor/block.rb
+++ b/lib/asciidoctor/block.rb
@@ -50,6 +50,7 @@ class Block < AbstractBlock
   # QUESTION should we store source_data as lines for blocks that have compound content models?
   def initialize parent, context, opts = {}
     super
+	@extraopts =  if opts.key? :attributes then opts[:attributes] else {} end
     @content_model = opts[:content_model] || DEFAULT_CONTENT_MODEL[context]
     if opts.key? :subs
       # FIXME feels funky; we have to be defensive to get lock_in_subs to honor override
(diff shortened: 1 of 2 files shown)


mojavelinux:
I think this is a great idea. It can be done in a much simpler way though as Asciidoctor already provides access to all the extra attributes on a block. I'll send a code review pull request to yours to clarify.

jalbertcruz:
Oh thank you, yes, I use that info (Block attributes) to pass it to highlight_source function in order to invoke the pygments wrapper. Surely it can be done more idiomatically.

mojavelinux:
@jalbertcruz, I've sent a pull request that implements this idea, but in a simpler way. Let me know if it looks acceptable to you.

https://github.com/asciidoctor/asciidoctor/pull/1430

I decided to name the attribute `highlight` instead of `hl_lines` to make it more clear.

mojavelinux:
Under the new proposal, your example would be written as follows:

```
[source, python, numbered, highlight="2,3"]
----
def function():
    for i in range(55):
        print(3)
----
```

mojavelinux:
Closing in favor of https://github.com/asciidoctor/asciidoctor/pull/1403.

jalbertcruz:
Yes, that's fine.

mojavelinux:
:+1:

mojavelinux:
Nice that I also implemented a lot of options for specifying the lines, which I describe in the related issue, #1429.

bweir:
I'd like to take advantage of this feature, but I'm not using syntax highlighting. Is that possible?

mojavelinux:
Not at the moment. One way you could take advantage of it is to use the syntax highlighter (in this case pygments), but set the source-language attribute on the document to text and use source blocks without indicating a language.

```
:source-language: text

[source,highlight="2"]
----
line 1
line 2
line 3
----
```

Issue #1224, opened 2014-11-30
Support inlining SVG images in HTML5 output


pepijnve:
When generating HTML5 output, SVG images can be inlined directly into the main output document. This could be an optional feature of the HTML5 backend that is enabled when data-uri is set.

mojavelinux:
Do you think we should do this when

a. the `data-uri` is set on the document and
b. the interactive option is set on the image

instead of using a base64 encoded data URI?

mojavelinux:
In other words, I'm trying to decide under what conditions this happens.

mojavelinux:
We are now discussing this as part of #1301.

mojavelinux:
I'm going to say that this change is covered by #1301. I don't think that `data-uri` alone should enact this behavior because there are certain implications brought on by using the `<svg>` element. In other words, these are two separate things.

PR #1404, opened 2015-07-09
resolves #1301 and #1224 add support for interactive & inline SVGs

### LINKED ISSUE #1301: Embed SVG images using <object> instead of <img>
The HTML5 backend always generates `` tags for image nodes (see https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L523). For SVG images it would be nice to have the option of generating `<object>` tags instead to allow hyperlinks inside the SVG images to work. This would resolve https://github.com/asciidoctor/asciidoctor-diagram/issues/64.

mojavelinux:
- use object element to embed SVG in HTML if interactive option is set 
- embed SVG data directly if inline option is set (including remote URI)
- always use img element for SVG if safe mode is secure
- add CSS to make object element with image/svg+xml data behave like img element
- add tests for SVG images
- consolidate/cleanup code in converter

### DIFF
--- a/data/stylesheets/asciidoctor-default.css
+++ b/data/stylesheets/asciidoctor-default.css
@@ -1,5 +1,5 @@
 /* Asciidoctor default stylesheet | MIT License | http://asciidoctor.org */
-/* Remove the comments around the @import statement below when using this as a custom stylesheet */
+/* Remove comment around @import statement below when using as a custom stylesheet */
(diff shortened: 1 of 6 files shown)


mojavelinux:
I think we're good to go here. We didn't change any assumptions about the alt text that we didn't already have, so if we address that I'd like to do it in a separate issue.

mojavelinux:
Instead of checking the env-github, we should use the `` element when running in secure mode.

mojavelinux:
Updated to include the latest proposal from #1301.

mojavelinux:
I have reviewed and optimized this pull request many, many times. I'm confident we are ready to move forward!

PR #1417, opened 2015-07-15
resolves #1400 don't copy css if stylesheet or stylesdir is a URI

### LINKED ISSUE #1400: stylesdir problem if set to a URL
If I set the stylesdir global attribute to a URL at which CSS style sheets can be found, the generated HTML doc works properly (i.e. correct <link> line is generated inside <head>) but Asciidoctor also creates a local directory mimicking the URL path. Because of the 'http://' in the URL, the result is an illegal directory which causes tools to crash (e.g. git). This directory is created because Asciidoctor wants to copy a code highlighting CSS somewhere (in my case, pygments.css).

I think this behaviour should be changed. One approach would be for stylesdir to be allowed to be a URL; if it is, all required CSS need to be found there, and no CSS copying is done, since that is the obvious implication of a remote CSS location.

I think the current situation would have to be regarded as a bug since the silent creation of an illegal directory certainly isn't a good thing.

Platform: Asciidoctor 1.5.2 / Cygwin 64 / Windows 8. Asciidoctor run under cygwin.

mojavelinux:


mojavelinux:
I also added logic to not attempt to copy the stylesheet if the stylesheet value itself is a URI (before the stylesdir is prepended).

PR #1450, opened 2015-08-01
Add new slide template converters

rahmanusta:
Those changes first proposal for template slide converters. Can we discuss the changes?

### DIFF
--- a/lib/asciidoctor/converter/deckjs.rb
+++ b/lib/asciidoctor/converter/deckjs.rb
@@ -1,0 +1,8 @@
+module Asciidoctor
+
+  class Converter::DeckjsConverter < Converter::Base
+    def convert node, template_name = nil, opts = {}
+      Slide.render(node)
+    end
+  end
+end
--- a/lib/asciidoctor/converter/factory.rb
+++ b/lib/asciidoctor/converter/factory.rb
@@ -212,6 +212,16 @@ module Asciidoctor
             require 'asciidoctor/converter/manpage'.to_s
           end
           ManPageConverter.new backend, opts
+        when 'revealjs'
+          unless defined? ::Asciidoctor::Converter::RevealjsConverter
+            require 'asciidoctor/converter/revealjs'.to_s
+          end
+          RevealjsConverter.new backend, opts
+        when 'deckjs'
+          unless defined? :: Asciidoctor::Converter::DeckjsConverter
+            require 'asciidoctor/converter/deckjs'.to_s
+            DeckjsConverter.new backend, opts
+          end
         end
 
         return base_converter unless opts.key? :template_dirs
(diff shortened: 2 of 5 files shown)


mojavelinux:
Thanks for the PR @rahmanusta! However, before we continue, I think this should be in the asciidoctor.js repository instead of the Asciidoctor core repository as it's specifically for the asciidoctor.js environment. Would you agree?

(If you're concerned we won't be able to make the change from asciidoctor.js alone, you can trust me that I'll find a way to integrate it from there).

rahmanusta:
Thank you @mojavelinux ! Yes, those converters can only work with JavaScript platforms (browser,node,nashorn ..). I see that Asciidoctor.js uses same core repository to generate JavaScript dist (https://github.com/asciidoctor/asciidoctor.js/blob/master/Gemfile#L9).

We can just put JS template converters to /dist directory but, we need to declare (https://github.com/asciidoctor/asciidoctor/pull/1450/files#diff-660cd2bae2d18e2bb4b054f51c7b1c22R215) that converters to core (asciidoctor-core.js and asciidoctor-all.js).How can we integrate those changes without touching to asciidoctor-core? If there is a way, that wil be best.

Mogztter:
@rahmanusta 
> We can just put JS template converters to /dist directory

You can generate the JavaScript from Ruby code in Asciidoctor.js, see https://github.com/asciidoctor/asciidoctor.js/pull/119

> We need to declare that converters to core

The factory is a tricky part, I don't know how we can "overload" the `create` method in Asciidoctor.js

rahmanusta:
Thank you @Mogztter for integration. 

  The factory is a tricky part, I don't know how we can "overload" the create method in Asciidoctor.js

I'dont know the way how to override a method in class with Opal but, If we load the Factory class again, it will override whole Factory class together with methods. @mojavelinux said me before it acts as CSS.

Mogztter:
So I guess we can duplicate the class and load the generated JavaScript file after asciidoctor-core.js to override the factory. This is the quick and dirty solution but we can start with that and then find a better solution.

mojavelinux:
>  I guess we can duplicate the class and load the generated JavaScript file after asciidoctor-core.js to override the factory.

Actually, the solution would just be to monkeypatch it, which is to add a Ruby file that redefines just that method. You can put any code you want into that method definition and it will be used in place of the original.

The goal should then be to change core so that we don't need to do that. Once the monkeypatch is in place, the requirement should be clear.

mojavelinux:
Can we agree that this has been reopened in asciidoctor.js as https://github.com/asciidoctor/asciidoctor.js/pull/119?

rahmanusta:
Yes, I agreed with @Mogztter 's PR (https://github.com/asciidoctor/asciidoctor.js/pull/119)

mojavelinux:
Let's close this one then and continue the discussion there.

PR #1492, opened 2015-10-01
Fixed Travis Status button in README file

jasnow:
Fixed Travis Status button in README file.

Example: https://github.com/jasnow/asciidoctor

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -73,7 +73,7 @@ Asciidoctor.js is used to power the AsciiDoc preview extensions for Chrome, Atom
 
 ifdef::badges[]
 .*Project health*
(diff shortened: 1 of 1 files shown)


jasnow:
New pull request

mojavelinux:
What's the problem with using shields.io?

jasnow:
Not familiar with shields.io but went back to it.

mojavelinux:
That's good, but the double quotes aren't necessary. So I'm not sure a change is needed.

If we think that shields.io is too slow, then I'm willing to consider using Travis CI directly. Travis CI does offer an SVG and that's really the only requirement.

jasnow:
Appears to not be needed since I see that travis is green now.

mojavelinux:
:+1:

PR #1496, opened 2015-10-03
unnecessary ruby version rechecking

nplusp:


mojavelinux:
Actually, the current syntax is intentional. Opal's preprocessor looks for that exact string (`RUBY_ENGINE == 'opal'`) to decide whether to include the source code in a conditional branch. In this case, Opal skips the else branch. That's what the comment above this line is trying to communicate.

mojavelinux:
I'd be open to a clarification to the comment line to explain this.

nplusp:
@mojavelinux thanks for the explanation.

Mogztter:
(For my curiosity) we can use `if RUBY_ENGINE_OPAL` in [asciidoctor.rb#L1609](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L1609) because there's no else block to skip ?

mojavelinux:
@Mogztter It's more that it is harmless to ignore it at runtime. The reason for `RUBY_ENGINE == 'opal'` filtering is to hide something from Opal so it doesn't choke on it. So it's like extra hidden.

PR #1500, opened 2015-10-05
remove deprecated rubyforge_project option

y-yagi:
Ref: http://www.rubydoc.info/github/rubygems/rubygems/Gem%2FSpecification%3Arubyforge_project


mojavelinux:
Thanks! I always wondered about that. Good to know!

y-yagi:
Thank you for confirmation!

PR #1508, opened 2015-10-11
resolves #1240 read docinfo footer files in AsciiDoc format

### LINKED ISSUE #1240: Allow docinfo footer files to be in AsciiDoc syntax
I want my foo-docinfo-footer.ad instead of foo-docinfo-footer.html because I want to re-use variables in my doc in the footer and, yeah, because HTML is the devil.

Sniffing the file extension should be sufficient for most cases.

mojavelinux:


mojavelinux:
This patch needs to be fixed so that docinfo files in AsciiDoc format are considered for the footer.

mojavelinux:
This will also need to be reworked in light of #1510.

mojavelinux:
I'm withdrawing this proposal as the preferred way to do this is to use the docinfo extension. See https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/master/lib/asciidoc-docinfoprocessor.rb for an example.

Issue #1464, opened 2015-08-14
number is ignored/replaced by 1 if it is followed by a dot 


madmas:
The result of a document like 

```
= Index page
:toclevels: 1
:numbered:

=== Sprint 2015/16

14 - 27.8.2015

=== Sprint 2015/16

14. - 27.8.2015
```

is 
![sample](https://cloud.githubusercontent.com/assets/37251/9271765/d513df00-427c-11e5-9ba7-c40b7361e5d8.png)

Note the missing 4 in the second part where the 14 should be followed by a dot.  

My assumption is that the parser interprets this as the beginning of a new list and ignores the characters before the dot while the intention is completely different ;-)

I reproduced this with a plain asciidoc based on the current lazybones template, which then uses these dependencies:

```
         'org.asciidoctor:asciidoctor-gradle-plugin:1.5.2'
         'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.6'
         'com.github.ben-manes:gradle-versions-plugin:0.8'
         'com.bluepapa32:gradle-watch-plugin:0.1.5'
```

AlanDaniels101:
This seems to be because of the ordered list regex. If you have one letter or any number of digits followed by a dot and another character, it treats it as an ordered list. 

There's another issue in that the list always starts at the first number/letter, regardless of what you begin the list at. This contradicts the documentation: http://asciidoctor.org/docs/user-manual/#ordered-lists. Using `[start]` works though. More on this, the roman numerals are a subset of [a-z,A-Z], and they get detected as alpha, not roman. Finally, multiple roman numerals aren't detected, such as `ii.`. 

I would be happy to look at this, but I don't know what the intended behavior is. 

Option 1 - Only accept `.`, `a.`, or `1.` (etc) as the start of an ordered list. 

Option 2 - Correctly keep track of the starting digit/letter.

mojavelinux:
This happens to the expected behavior assuming the goal is alignment with AsciiDoc Python. (Unexpected, perhaps, but still compliant). @AlanDaniels101 correctly points out that an ordered list is detected if the line begins with a letter or number followed by a dot followed by a space.

One workaround you can use today is to put a blank placeholder in front of the number to hide it from the parser:

```
{empty}14. this is just a number
```

or

```
:dot: .

14{dot} this is just a number
```

It works, but it's not pretty.

[...]

PR #1509, opened 2015-10-12
track ordered list starting marker

AlanDaniels101:
See https://github.com/asciidoctor/asciidoctor/issues/1464. This change checks the marker of an ordered list and adjusts the start attribute to match the marker. This lets
```
5. five
6. six
```
do the same thing as
```
[start=5]
. five
. six
```
This is still under discussion.


mojavelinux:
While working with a documentation set, I came across this situation again. With a fresh look, it's clear we should absolutely do this. I'm going to review this change and try to get it into 1.5.6. Sorry for the delay!

mojavelinux:
Thanks again for submitting this pull request. I used it to help me think about the best way to implement this feature. I submitted that change as #4794, which has been merged into main and will be included in the 2.1.0 release.

PR #1512, opened 2015-10-13
Fixes #1366: Added method AbstractNode#add_role

### LINKED ISSUE #1366: Methods to add roles on a node are missing
AFAICS there is no simple way at the moment to add a role to a node for an extension.
The roles are stored as a blank separated string in the attribute `role`.
Without knowing about this implementation detail it is hard to add a role.
It would be nice to have dedicated methods add_role and remove_role.

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/abstract_node.rb#L197-L227

robertpanzer:
This PR adds a method `AbstractNode#add_role` to add a role to a certain node.
Document level roles are not considered when adding a role, it will be added in any case, if the node itself does not carry that role yet.

### DIFF
--- a/lib/asciidoctor/abstract_node.rb
+++ b/lib/asciidoctor/abstract_node.rb
@@ -221,6 +221,21 @@ class AbstractNode
     end
   end
 
+  # Public: A convenience method that adds the given role directly to this node
+  def add_role(name)
+    unless (roles = (@attributes['role'] || '').split(' ')).include? name
+      @attributes['role'] = roles.push(name) * ' '
+    end
+  end
+
+  # Public: A convenience method that removes the given role directly from this node
+  def remove_role(name)
+    if (roles = (@attributes['role'] || '').split(' ')).include? name
+      roles.delete name
+      @attributes['role'] = roles * ' '
+    end
+  end
+
   # Public: A convenience method that checks if the reftext attribute is specified
   def reftext?
     @attributes.has_key?('reftext') || @document.attributes.has_key?('reftext')
--- a/test/attributes_test.rb
+++ b/test/attributes_test.rb
@@ -1067,6 +1067,79 @@ Text
       assert para.attributes.has_key?('option2-option')
     end
 
+    test 'a role can be added using add_role when the node has no roles' do
+        input = <<-EOS
+A normal paragraph        
+        EOS
+      doc = document_from_string(input)
+      para = doc.blocks.first
+      para.add_role 'role1'
+      assert_equal 'role1', para.attributes['role']
+      assert para.has_role? 'role1'
+    end
+
+    test 'a role can be added using add_role when the node already has a role' do
+        input = <<-EOS
+[.role1]
+A normal paragraph        
+        EOS
+      doc = document_from_string(input)
+      para = doc.blocks.first
+      para.add_role 'role2'
+      assert_equal 'role1 role2', para.attributes['role']
+      assert para.has_role? 'role1'
+      assert para.has_role? 'role2'
+    end
+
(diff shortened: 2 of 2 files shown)


lib/asciidoctor/abstract_node.rb
```diff
@@ -221,6 +221,13 @@ def roles
     end
   end
 
+  # Public: A convenience method that adds the given role directly to this node
+  def add_role(name)
+    role_attribute = @attributes['role']
+    roles = role_attribute ? role_attribute.split(' ') : []
+    @attributes['role'] = (roles.dup << name).join(' ')
```
mojavelinux: I don't think you need to dup the roles array since it is a local variable.

Here's how I recommend writing this:

```ruby
@attributes['role'] = (@attributes['role'] || '').split(' ').push(name).uniq * ' '
```

mojavelinux: Actually, one small change to make it even more optimal:

```ruby
unless (roles = (@attributes['role'] || '').split(' ')).include? name
  @attributes['role'] = roles.push(name) * ' '
end
```

mojavelinux:
> Document level roles are not considered when adding a role

That's the correct behavior.

lib/asciidoctor/abstract_node.rb
mojavelinux: Ping me after you update.

robertpanzer:
Ping :)

lib/asciidoctor/abstract_node.rb
robertpanzer: I added the `dup` because an extension could create multiple nodes, everytime passing in the *same* value for the attribute *role*.
This is maybe a pathological case and could be handled as a simple "user error" I initially wanted to stay on the safe side.

mojavelinux: In this case, the value was already local so the reference was already detached.

mojavelinux:
A few more things:

1. Could you rebase into a single commit?
2. Could you modify the commit message so it starts with "resolved #1366" (that way it can be found using git log)
3. What about the `remove_role` method?

Here's a possible impl for remove_role:

```ruby
def remove_role(name)
  if (roles = (@attributes['role'] || '').split(' ')).include? name
    roles.delete name
    @attributes['role'] = roles * ' '
  end
end
```

And yes, it's becoming painfully clear we should be storing roles as a collection instead of a string :)

lib/asciidoctor/abstract_node.rb
mojavelinux: To be clear, it's the `split` that creates the local detached reference in the current implementation. So there's no risk we are modifying the original reference.

mojavelinux:
@robertpanzer I sent you a PR that fixes the failing test, adds one additional test and fixes some of the wording. Once you merge that, can you rebase again into a single commit? Then I think we're good to merge.

robertpanzer:
Thank you!
Sorry for making you so much work. :-(
I somehow missed the failing test between two other failures I see locally.

mojavelinux:
You're doing great! And the build gods are happy!

mojavelinux:
I finally remembered how to delete an item from an array without mutating the array so that we can cut out one of the lines:

```ruby
@attributes['role'] = (roles - [name]) * ' '
```

It's very strange that Ruby doesn't provide a method to remove an item from an array without modifying the array in place.

I can make this change after merging. I just wanted to mention it so I didn't forget.

mojavelinux:
Thanks again for submitting this patch!

PR #1526, opened 2015-10-18
Add option to control timestamp generation.

### LINKED ISSUE #1453: Add support for reproducible builds
Asciidoctor should support building documents reproducibly. Debian is using this to allow people to prove that their builds actually were built from the specified source, and numerous organizations use rsync or other content-based transfer systems where data is only copied or downloaded if it is modified ($DAYJOB does this).

From what I can see of the code, the only real change that's required for this is to change the code that creates and emits date stamps. I'm thinking this could be controlled by a "reproducible" attribute, but I'm open to other ideas.

bk2204:
This commit adds a "timestamps" option, defaulting to true, that controls whether timestamps (docdate, doctime, docdatetime, localdate, localtime, and localdatetime) are generated. If unset, this option ensures that output is reproducible.

Fixes #1453.

Viewing with git diff -w is recommended.

### DIFF
--- a/lib/asciidoctor/converter/docbook5.rb
+++ b/lib/asciidoctor/converter/docbook5.rb
@@ -661,7 +661,9 @@ module Asciidoctor
       result = []
       result << %(<#{info_tag_prefix}info>)
       result << document_title_tags(doc.doctitle :partition => true, :use_fallback => true) unless doc.notitle
-      result << %(<date>#{(doc.attr? 'revdate') ? (doc.attr 'revdate') : (doc.attr 'docdate')}</date>)
+      if (doc.attr? 'revdate') || (doc.attr? 'timestamps')
+        result << %(<date>#{(doc.attr? 'revdate') ? (doc.attr 'revdate') : (doc.attr 'docdate')}</date>)
+      end
       if doc.has_header?
         if doc.attr? 'author'
           if (authorcount = (doc.attr 'authorcount').to_i) < 2
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -199,7 +199,7 @@ module Asciidoctor
         result << '<div id="footer">'
         result << '<div id="footer-text">'
         result << %(#{node.attr 'version-label'} #{node.attr 'revnumber'}#{br}) if node.attr? 'revnumber'
-        result << %(#{node.attr 'last-update-label'} #{node.attr 'docdatetime'}) if node.attr? 'last-update-label'
+        result << %(#{node.attr 'last-update-label'} #{node.attr 'docdatetime'}) if (node.attr? 'last-update-label') && (node.attr? 'timestamps')
         result << '</div>'
         result << '</div>'
       end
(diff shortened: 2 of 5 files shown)


mojavelinux:
Do you think it's necessary for the attributes not to be set, or do you think it's sufficient to just not use them in the converters? Obviously, your pull request suggests that you don't think they should be set, but are you open to the approach of simply not using them?

bk2204:
I'm fine with simply not using them if you think that's a better idea.  It's as simple as removing that one block.

mojavelinux:
I think the right approach is that the flag indicate solely whether or not timestamps should be used. The information still may be important to have to the client of the API (for one reason or another). What's important is that it not make its way into the output.

We also want the behavior to be consistent regardless of whether the attribute is set from the API/CLI or the document header. We'd have to do more work to prevent the timestamp attributes from being set if this attribute is set in the document header (because of how Asciidoctor is implemented)...but rather than go down that road we can simply recommend the converters honor the flag (and certainly for the built-in converters).

Also, the value of the attribute should be empty string. In AsciiDoc/Asciidoctor, that's the equivalent of true (and consistent with other boolean attributes).

bk2204:
The code now uses the empty string. It also leaves the docdate variable present, just controlling the generation of timestamps.

mojavelinux:
There are two changes I'd like to make. I'm happy to make these when merging to avoid the back and forth.

First, I think we should reverse the logic so that you add an attribute to disable the output of timestamps. I'm thinking of one of the following names:

* reproducible
* static
* invariable
* consistent

The reason I'm leaning in this direction is because I think there is more to making the output reproducible than just timestamps. For instance, I think the version of Asciidoctor should also be dropped, which is the second change I'd like to make.

wdyt? (I'm leaning back towards "reproducible" since it seems to be the most clear).

mojavelinux:
Actually, my first choice may be "consistent" since "reproducible" isn't technically a word (according to official language sources). I think the intent is equally clear. wdyt?

mojavelinux:
After a brief drive and some more thinking, I came up with a few other ideas. First, if we simply consider renaming "timestamps", I discovered a fitting replacement: "transitory". Transitory means only lasting a short time.

If we went for the opposite meaning, here's one that stands out:

* stable-output

That's very specific and thus hard to be misinterpreted. I suppose it says the same thing as reproducible.

I recognize I'm flip-flopping a bit here, but since we're introducing a new attribute, I want to make sure we get it right.

bk2204:
I think "reproducible" would be best.  It's listed as a word by both Wiktionary and Merriam-Webster. It's also what Debian's reproducible builds project is called.

As for the version number, we can drop it if you like, but it's generally assumed that output of document processing systems will not always be consistent among versions; otherwise, no new features could be added.

robertpanzer:
I don't think that dropping the version number is required for reproducible
builds.
A different asciidoctor version could produce a different HTML besides the
asciidoctor version number.


Am Donnerstag, 31. Dezember 2015 schrieb brian m. carlson :

mojavelinux:
Thanks for the insight! I think we'll go with "reproducible" then (off by default) and leave the Asciidoctor version number.

I tried to rebase and send a PR to your PR, but somehow I botched it. So I just merged to master. Let me know if the commit looks acceptable to you. We can still revise if necessary.

mojavelinux:
Merged as cbb85753bbd754ac4f2ba093c77938dcbf373e03

bk2204:
That change looks good.

mojavelinux:
:+1: Thanks for looking it over.

PR #1534, opened 2015-10-22
resolves #1533 convert footer before body content

### LINKED ISSUE #1533: Convert body content after footer content
The main body content should be converted after the footer content. The scope data available to the footer content should be the same as the data available to the header.

Making this change ensures that the footer only references attributes that are defined in the document header and not those defined partway through the document.

mojavelinux:


mojavelinux:
We're going to pull back on this idea until we have a more concrete definition of what is expected.

PR #1553, opened 2015-10-27
Add link for MIT license

glamouracademy:


mojavelinux:
I'd rather link to the LICENSE file in the project. So if we do make a change here, it would be:

```
{uri-license}[the MIT license]
```

glamouracademy:
totally makes sense. I'll update and squash.

glamouracademy:
Updated! Thanks!

mojavelinux:
Thanks!

PR #1576, opened 2015-11-17
Initialize slim-related attributes regardless of when slim was loaded

terceiro:
Setting instance attributes of the Template class cannot depend on
whether slim has been loaded before or not.


mojavelinux:
Thanks for catching this!

PR #1592, opened 2015-11-29
Fix for #1590

### LINKED ISSUE #1590: manpage: translating links to .URL introduces extraneous newline/space
When translating `http://asciidoctor.org[Asciidoctor],` as
```
.URL "http://asciidoctor.org" "Asciidoctor" ","

```
the introduction of the newline plays havoc with following text, unless the following text is another link. For instance,

- If the comma is followed by a space and then some regular text, that space becomes a line-leading space which forces a linebreak in troff.

- If the comma is followed by a newline, that newline is added to the newline already introduced by the `.URL` and causes vertical space that was not intended (so it looks like a paragraph break).

As an example:
```
The relevant sites are
http://asciidoctor.org[Asciidoctor], http://racket-lang.org[Racket],
http://neovim.io[NeoVim], and
http://gnu.org/software/groff[Groff].
```
gets translated to
```
The relevant sites are
.URL "http://asciidoctor.org" "Asciidoctor" ","
.URL "http://racket\-lang.org" "Racket" ","

.URL "http://neovim.io" "NeoVim" ","
 and
.URL "http://gnu.org/software/groff" "Groff" "."

```
so the two newlines after `Racket,` and the space before `and` cause problems.

While the newline is certainly needed in order to terminate the arguments to `.URL`, in order to stop it from being treated as actionable space by troff, it needs to be immediately followed by `\c`.

I.e., the translation above should be
```
The relevant sites are
.URL "http://asciidoctor.org" "Asciidoctor" ","
\c
.URL "http://racket\-lang.org" "Racket" ","
\c

.URL "http://neovim.io" "NeoVim" ","
\c
 and
.URL "http://gnu.org/software/groff" "Groff" "."
\c

```

ds26gte:
Resolves #1590: Generated .URL call should be followed by a \c so no extraneous space is introduced.

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -23,7 +23,7 @@ module Asciidoctor
       str = str.
         gsub(/^\.$/, '\\\&.').    # a lone . is also used in troff to indicate paragraph continuation with visual separator
         gsub(/\\$/, '\\(rs').     # a literal backslash at the end of a line
-        gsub(/^\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*)(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\3"). # quote last URL argument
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks!

Issue #1589, opened 2015-11-28
manpage: leading \. is left as is, should be \&.


ds26gte:
When the manpage backend is used, `\.` (backslash, period) at the head of a line in the source document is left as is. This causes the line to be treated as a macro call by `man` (and if the characters following the `\.` don't constitute a macro, the entire line is ignored).

It should be translated to `\&.` (i.e., backslash, ampersand, period).

mojavelinux:
Perhaps add a replacement under the one for the lone `.` at https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/manpage.rb#L24

```
.gsub(/^\\\./, '\\\&.')
```

ds26gte:
Yes, your suggested snippet will resolve it. You may either add it yourself, or take my recent PR.

PR #1595, opened 2015-11-30
Fix for #1589

ds26gte:
Fix for #1589. Line beginning with `\.` should be translated as `\&.`, so as not to be mistaken for a troff macro call.


mojavelinux:
Thanks @ds26gte!

PR #1607, opened 2015-12-09
Resolves #1604: backslash in `...`, listing/literal blocks should typeset as itself not escape

### LINKED ISSUE #1604: manpage: \ inside `...` and listing/literal block should be verbatim in manpage too
Occurrences of `\` (backslash) inside \`...\` and listing/literal blocks are currently being sent as-is to the manpage output. Man however treats backslash as an escape, so what the user sees depends on whatever character happens to follow the backslash. (Typically, the backslash gets swallowed, so `\input` becomes `input`.)

This is in contrast to the HTML backend of asciidoctor, where the backslash in such contexts is treated as a verbatim backslash (i.e., it typesets as itself).

E.g., the following document fragment gets treated differently between the HTML and manpage backends:

````
TeX uses `\input` to include files.  For example,

   \input otherfile

----
\input otherfile
----
````

Let me know if you agree this is a bug. I have a fix for this that makes the manpage output align with the HTML output (which I consider to be correct).

### PR DESCRIPTION
**Pull request by ds26gte**
A backslash inside `...` and listing/literal blocks (and only within those contexts) is converted to `\e`, which typesets as a verbatim backslash.  This matches what happens in the HTML backend.

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -51,6 +51,12 @@ module Asciidoctor
       append_newline ? %(#{str}#{LF}) : str
     end
 
+    def preserve_backslash str
+      str = str.
+        gsub(/\\/, '\\e')
+      str
+    end
+
     def skip_with_warning node, name = nil
       warn %(asciidoctor: WARNING: converter missing for #{name || node.node_name} node in manpage backend)
       nil
@@ -254,7 +260,7 @@ T})
 .RS 4
 .\\}
 .nf
-#{manify node.content}
+#{preserve_backslash (manify node.content)}
 .fi
 .if n \\{\\
 .RE
@@ -272,7 +278,7 @@ T})
 .RS 4
 .\\}
 .nf
-#{manify node.content}
+#{preserve_backslash (manify node.content)}
 .fi
 .if n \\{\\
 .RE
@@ -655,7 +661,7 @@ allbox tab(:);'
       when :strong
         %[\\fB<BOUNDARY>#{node.text}</BOUNDARY>\\fP]
       when :monospaced
-        %[\\f[CR]<BOUNDARY>#{node.text}</BOUNDARY>\\fP]
+        %[\\f[CR]<BOUNDARY>#{preserve_backslash node.text}</BOUNDARY>\\fP]
       when :single
         %[\\(oq<BOUNDARY>#{node.text}</BOUNDARY>\\(cq]
       when :double
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -19,12 +19,10 @@ module Asciidoctor
     def manify str, opts = {}
       append_newline = opts[:append_newline]
       preserve_space = opts.fetch :preserve_space, true
-      preserve_backslash = opts[:preserve_backslash]
       str = preserve_space ? str.gsub(TAB, ETAB) : str.tr_s(%(#{LF}#{TAB} ), ' ')
       str = str.
-        gsub(/^\.$/, '\\\&.').    # a lone . is also used in troff to indicate paragraph continuation with visual separator
-        gsub(/^\\\./, '\\\&.').   # line beginning with \. shouldn't be mistaken for troff macro
-        gsub(/\\$/, '\\(rs').     # a literal backslash at the end of a line
+        gsub(/\\(?!f(?:B|I|P|\[CR\])|0|\()/, '\\(rs'). # literal backslash (ignores \fB, \fI, \fP, \f[CR], \0 & \( sequences)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/asciidoctor/converter/manpage.rb [hunk 1]
```diff
@@ -51,6 +51,12 @@ def manify str, opts = {}
       append_newline ? %(#{str}#{LF}) : str
     end
 
+    def preserve_backslash str
+      str = str.
+        gsub(/\\/, '\\e')
```
mojavelinux: Could you reduce this to one line and change the first argument to a single quoted backslash string?

Perhaps we can add this as an option named `:preserve_backslash` to the manify method. That way, we don't need an extra method in the first two cases. Obviously, we still need it for the inline case.

ds26gte: Submitted new PR (97ff93a).

--d

ds26gte: Submitted a new PR 97ff93a.

ds26gte:
a7fe749.  (ignore previous.)

mojavelinux:
I hate to do this to you, but I gave it some more thought and realized we were missing the general case. When a backslash is present in `node.content`, and it's not one of the inline formatting marks, then we know it is a literal backslash and needs to be escaped. So I switched to escaping them by default except for known formatting patterns.

See https://github.com/ds26gte/asciidoctor/pull/1

Once you merge, please squash your commits so we get down to just one (you can squash my changes into yours).

This change is still a little risky because we cannot know for absolute certain when a backslash is literal backslash. We can make a pretty good guess. The reason stems from how we convert. We convert inline before we convert the block content. As a result, some troff formatting marks have already been inserted in the content (like bold and italic).

The only truly safe way to do it would be to convert the inline elements to an intermediary format (like XML) then convert to troff in manify. But let's stick with this approach for now.

mojavelinux:
> `\<space>` and `\&` are generated for .TH's fourth and fifth arguments (the manpage title), and the backslashes in these become verbatim, when they should not.

Oops, I forgot about those. I'm trying to avoid using the `\<space>` sequence (since we can't tell that from a standalone backslash), so I'll switch that to `\0` and I'll add the exception for `\&`.

> \fB and the like (i.e., your exceptions on line 24) entered inside inline-code and listings should be verbatim, but they are not.

I just don't know how to distinguish between literal content and formatting marks in this case (without using the intermediary markup I previously proposed). Keep in mind this has nothing to do with inline code or listings as these formatting marks are technically valid anywhere. (formatting can be used in literal/verbatim sections by modifying the subs attribute).

In Asciidoctor PDF, we used the intermediary format approach and I'm afraid we're going to have to bite the bullet and do the same thing here if we really want to cover all the cases. Until then, you just can't use these sequences as content.

mojavelinux:
I thought of a workaround that I think will work for 99% of the cases. When we lay down the inline formatting, we can use the null character instead of a backslash before each command. For example, `\0fB` instead of `\fB`. Then, in manify, we can switch all null characters to backslash. We could really choose any non-printable character, even a sequence. That way, we can tell a formatting backslash from one in the content. I gave it a test and this works out pretty nicely.

mojavelinux:
I'll send a pull request to your branch with this proposal.

mojavelinux:
Btw, we should probably add some tests for these changes (and the manpage converter in general) so that we don't break these cases in the future.

ds26gte:
Does this essentially rely on the user not using a literal null character in their listings (is that the 1%)? I think that's a very reasonable expectation and requirement of the user. 

The earlier proposed prohibition on using certain backslash sequences in listings would have been too draconian -- especially in programming-manual-type texts (and not just texts about troff).

mojavelinux:
> Does this essentially rely on the user not using a literal null character in their listings

Yes. This would be extremely rare. In fact, we already rely on something similar in core (except we use a different escape character because a) we need a pair and b) syntax highlighting libraries actually detect the null character...but not a concern here).

mojavelinux:
> The earlier proposed prohibition on using certain backslash sequences in listings would have been too draconian

I totally agree. Sometimes, you have to say it to hear how reasonable it sounds :) In this case, we had to dig deeper. And this totally works in all my tests so far. I'll send a pull request w/ some tests just to be sure.

mojavelinux:
"this totally works" refers to our null character for escape sequence hack.

We do have to make sure that all text that can be formatted passes through manify, but I'm pretty sure it does already...or we can deal with the exception if and when we find it. Another approach is to restore the null character as the escape sequence backslash in the document node, thus across the whole document. I'll leave a note of this.

mojavelinux:
PR is ready with tests https://github.com/ds26gte/asciidoctor/pull/2

mojavelinux:
Merged as 31feac906f023ca413dc96040c3464815fa3bee1.

PR #1632, opened 2015-12-25
Resolves #1631. Inside listings, escape leading dots...

### LINKED ISSUE #1631: manpage: listings with leading `.` that happens to be valid troff are mistranslated
This is almost related to #1589 but not quite.

If a listing contains a line that's valid troff, the translated manpage treats it as valid troff (instead of verbatim text) and tries to execute it!  E.g.,

```
AAA should show.

----
BBB should show.
.if 1 .nx
CCC should show but doesn't.
----

DDD should show but doesn’t.
```

Same thing happens for the indent-everything-approach to getting a listing:

```
AAA should show.

   BBB should show.
   .if 1 .nx
   CCC should show but doesn't.

DDD should show but doesn’t.
```

A user remedy is to make sure that the `.`-line is indented (in the second case, indented relative to the other lines), but this does put the onus on the author.

The conceptual fix is to prefix  line-leading `.` with `\&` inside listings, so `manify` may need an option just for this.  (BTW, `manify` currently has a `preserve_space` option which I don't think is used.)

ds26gte:
...so they don't cause troff calls.

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -22,8 +22,8 @@ module Asciidoctor
     def manify str, opts = {}
       str = ((opts.fetch :preserve_space, true) ? (str.gsub TAB, ETAB) : (str.tr_s %(#{LF}#{TAB} ), ' ')).
         gsub('\\', '\\(rs').      # literal backslash
-        gsub(/^\.$/, '\\\&.').    # lone . is used in troff to indicate paragraph continuation with visual separator
-        gsub(/^\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*)(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\c#{LF}\\3"). # quote last URL argument
+        gsub(/^\./, '\\\&.').     # leading . is used in troff for macro call or other formatting
+        gsub(/^\!MAN!.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*)(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\c#{LF}\\3"). # quote last URL argument
         gsub(/(?:\A\n|(?:\n *| +)(\n))^\.(URL|MTO) /, "\\1\.\\2 "). # strip blank lines in source that precede a URL
         gsub('-', '\\-').
         gsub('&lt;', '<').
@@ -48,7 +48,6 @@ module Asciidoctor
         gsub(/<\/?BOUNDARY>/, '').# artificial boundary
         gsub(ESC, '\\').          # restore backslash used for escape sequences (NOTE could be applied on document content)
         rstrip                    # strip trailing space
-      str = opts[:inside_listing] ? str.gsub(/^\./, '\\\&.') : str # inside listings, escape leading dots so they don't cause troff calls
       opts[:append_newline] ? %(#{str}#{LF}) : str
     end
 
@@ -255,7 +254,7 @@ T})
 .RS 4
 .\\}
 .nf
-#{manify node.content, :inside_listing => true}
+#{manify node.content}
 .fi
 .if n \\{\\
 .RE
@@ -273,7 +272,7 @@ T})
 .RS 4
 .\\}
 .nf
-#{manify node.content, :inside_listing => true}
+#{manify node.content}
 .fi
 .if n \\{\\
 .RE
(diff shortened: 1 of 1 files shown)


mojavelinux:
This should be restricted to listings. All node content is going to need this treatment.

We could distinguish between an inline (quoted) string and paragraph content. For instance, technically there is no reason to check for a period in a section title since it's a quoted string.

However, it also doesn't hurt to escape the period, so either way will work.

What we want to do is just change this line:

```ruby
gsub(/^\.$/, '\\\&.'). # lone . is used in troff to indicate paragraph continuation with visual separator
```

to


```ruby
gsub(/^\./, '\\\&.'). # a leading . is used in troff to indicate a macro or other formatting call
```

Also, could you add a test to cover this case? I liked the example you gave:

```
AAA this line should be visible.
.if 1 .nx
BBB this line and the one above it should be visible.
```

See https://github.com/asciidoctor/asciidoctor/blob/master/test/manpage_test.rb#L42-L48

ds26gte:
But there is a line immediately following

```
gsub(/^\.$/, '\\&.').
```

that looks for lines beginning with `.URL` or `.MTO`.  Escaping all line-beginning dots would prevent this line from ever taking effect, no?

Something isn't right: If `manify` only ever receives unprocessed lines from the source document, how is it coming across `.URL` and `.MTO` lines, which are troff? Are you making an exception for `.URL` and `.MTO` because they need the additional processing (for adding the `\c` and for removing any preceding blank line)?

mojavelinux:
Shoot, you are right. We add our own macros (in this case by inline_anchor), so we need to make sure those pass though. I'll give it some more thought.

ds26gte:
I'm inserting `!MAN!.URL` and `!MAN!.MTO` rather than the raw troff calls. `manify` then takes care to lop off the `!MAN!`.  This allows leading-dot lines to be treated uniformly by `manify` without worrying about whether the context is a listing or not.

An objection: What if the user themselves started a line with `!MAN!.URL`. This must be addressed: Can you insert unforgeable strings as prefixes? (I remembered you used `\0` before.)

The objection is already valid for existing code: A user can't start a line with `.URL` without confusing `manpage.rb`.

mojavelinux:
I have an idea for a solution. I'll send a PR to your PR.

mojavelinux:
You had the right idea, except we should use a control character instead of a visible character so that it cannot be easily typed (and thus extremely unlikely).

I repurposed the code to use the escape character as a prefix instead of a replacement for a troff backslash. That allows us to also use it to escape the troff macro. Then, we just remove the escape character when we are done with it.

PR sent. It looks like a large change, but a small change just affected a lot of lines.

ds26gte:
Excellent! Works on all my examples. Thanks for including the test too. (I didn't know how.)

One very minor suggestion that you may ignore: I would personally replace the line
```
gsub(ESC, '').
```
with the two lines
```
gsub(ESC_BS, '\\').
gsub(ESC_FS, '.').
```
as this allows a user-inserted `ESC` character to pass through at least whenever it doesn't precede `\` or `.`. Not that this is tremendously useful, but it doesn't hurt, and it follows the don't-touch-user-input-unnecessarily principle.

(_Full stop_ rather than _period_ or even the less-sentence-oriented _dot_?  :-)  `ESC_FS` does go nicely with `ESC_BS`.)

mojavelinux:
That's definitely a fair change. I agree that we shouldn't be removing marks we didn't add.

It's important to keep in mind that the nature of the ASCII-style escaping technique we are using is that we cannot perfectly differentiate between our own marks and marks in the original input. It's reasonable given an understanding we have with the user. In no way should the user ever be entering a command sequence into content (perhaps we should be explicit about this in the docs). (Most people don't even know how). If that constraint is followed, everything will work perfectly.

Control characters (aka escape sequences) are reserved for text processors like Asciidoctor. That's been true ever since forever ago.

mojavelinux:
I squashed and merged.

mojavelinux:
Note that we didn't have to unescape the full stop because it is already done earlier in the method. I added a note to make sure that this line is reviewed if we ever do add addition escapes.

mojavelinux:
Go ahead and give it a final test (on master) and make sure everything looks good to you.

ds26gte:
Looks good! However, while testing, I found an unrelated bug that had to do with `.URL`s for which I've opened a PR.

mojavelinux:
I noticed this problem as well with my sample document. I think our endline fix changed an assumption, but also made it easier to address. Thanks for digging into it!

PR #1637, opened 2015-12-28
resolves #1633 disable toc if document has no sections

### LINKED ISSUE #1633: Hide table of contents if empty
Hide the table of contents if it would otherwise be empty. There's no point in showing an empty toc.

If, for some reason, we want to allow an empty toc, then it should require an attribute to enable it (e.g., `toc-show-if-empty`).

mojavelinux:


mojavelinux:
@nawroth Does this implement the desired behavior?

nawroth:
According to the test it does!

mojavelinux:
:+1:

mojavelinux:
Actually, I was asserting on the input instead of the output. Fixed!

PR #1638, opened 2015-12-28
Remove horizontal space after .URL's third argument…

ds26gte:
…since linefeed after \c is already a space.

This is a subtle bug that occurs at only some line lengths. E.g.,
```
The script troff2page runs on the Unix command-line on Cygwin, Linux,
Mac OS X, and Solaris.  It uses the Common Lisp implementation mentioned
in the shell environment variable LISP, which can currently be set to
either clisp, clozure, cmu, ecl or sbcl: the corresponding Lisp
implementations being
http://clisp.sf.net[CLISP],
http://ccl.clozure.com[Clozure CL],
http://cmucl.org[CMUCL],
http://ecls.sf.net[ECL], and
http://sbcl.sf.net[SBCL].
```
The line `http://ecls.sf.net[ECL], and` currently gets translated to
```
.URL "http://ecls.sf.net" "ECL" ","
\c
 and
```
Note the space before `and`. If the man pager happens to line break just at the `,`, that space shows up as a glaring leading space in the immediately following line.  (My example above will show up this error if you have a `MANWIDTH` of 80.)

The translation is placing after the generated `\c` whatever's after `.URL`'s 3rd argument, and this typically has some leading space.  But this isn't needed (indeed, _shouldn't_ be included), because the linefeed introduced after the generated `\c` is already a space. (The extraneous space shows up even when there is no coincidental linebreak, but is less visible to human eyes then because it seems like it could be an artifact of the rather courageous monospace-based both-sides line justification done by the man pager.)

PR submitted.

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -26,6 +26,7 @@ module Asciidoctor
         gsub(/(?:\A|[^#{ESC}])\\/, '\&(rs'). # literal backslash (not a troff escape sequence)
         gsub(/^\./, '\\\&.').     # leading . is used in troff for macro call or other formatting
         # unescape troff macro and quote last URL argument
+        gsub(/^#{ESC}\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*) *$/,          ".\\1\"\\2\"#{LF}\\c").
         gsub(/^#{ESC}\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*) *(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\c#{LF}\\3").
         # strip blank lines in source that precede a URL
         gsub(/(?:\A\n|(?:\n *| +)(\n))^\.(URL|MTO) /, '\1.\2 ').
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -26,7 +26,7 @@ module Asciidoctor
         gsub(/(?:\A|[^#{ESC}])\\/, '\&(rs'). # literal backslash (not a troff escape sequence)
         gsub(/^\./, '\\\&.').     # leading . is used in troff for macro call or other formatting
         # unescape troff macro and quote last URL argument
-        gsub(/^#{ESC}\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*)(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\c#{LF}\\3").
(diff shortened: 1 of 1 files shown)


ds26gte:
More space issues: If the `.URL` has nothing following its third argument (which itself might be missing), there shouldn't be a `#{LF}` after the `\c`, because there's already a linefeed in the source.  (The extra linefeed will show up as an additional space.) 

I currently see no way other than using two `gsub`s (see below). Maybe there is a way to head off the whole something-following-the-third-argument when you generate the `.URL` in the first place? (Troff itself doesn't have to deal with this, because there a `.URL` is given at most 3 arguments. In your intermediate troff generation, you are adding more text on the `.URL` line, which you then have to explicitly shove to the next line later when doing `manify`.)

```
# unescape troff macro and quote last URL argument
gsub(/^#{ESC}\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*) *$/,          ".\\1\"\\2\"#{LF}\\c").
gsub(/^#{ESC}\.((?:URL|MTO) ".*?" ".*?" )( |[^\s]*) *(.*?)( *)$/, ".\\1\"\\2\"#{LF}\\c#{LF}\\3").
```

ds26gte:
PR updated.

mojavelinux:
The best way to handle this scenario is to use a block on the gsub. That allows us to use logic to determine how to handle the two different situations.

(Technically, it's possible to do conditional logic in a replacement in Ruby 1.9+, but it gets really hard to read...and we haven't officially Ruby 1.8 yet, so we'll use what we've got).

I was also able to consolidate the two substitutions so that blank lines are handled at the same time we handle the quoting. This increases accuracy of the substitutions.

Finally, I added tests to make sure we don't break this :)

mojavelinux:
See my PR to your PR.

mojavelinux:
I just realized that we could skip this whole blank line removal business if we used `\c` *before* the `.URL` and `.MTO` macros. I didn't know about that escape sequence at the time, but now I realize it's a perfect fit.

The `.URL` and `.MTO` macros address the case of a character that immediately trails, but they don't address the case of a character that immediately precedes. `\c` gives us that control.

mojavelinux:
Using `\c` as a proposed also addresses the situation we still have, that a space is getting inserted between a link and the non-blank character that immediately proceeds it.

ds26gte:
Unfortunately, in a snow-induced brain-freeze (and also pushing the limits of my git skills), I created a new `ds26gte:manpage-url-fix` rebasing it off your new `master` (instead of just rebasing `ds26gte:manpage-default-boilerplate` and leaving this one alone).

I've resubmitted the PR, but there's a slight possibility that something got lost. Could you re-send your PR-to-my-PR without modifying anything so I can make sure I didn't lose anything?

mojavelinux:
Yep, I'll resend.

Issue #1639, opened 2015-12-29
manpage backend shouldn't give up on source that doesn't have standard format


ds26gte:
New issue peeled off from the #1636 discussion:

Currently, Asciidoctor's `manpage` backend abandons creating a man page if the input source doesn't conform to a standard man pattern, in particular, if the backend can't determine the `manname` and `manpurpose`.  Asciidoctor does cook up meaningful fallbacks for `mantitle` and `manvolnum`. 

It is useful to generate a manpage even for nonconforming input for two reasons:
1. It is confusing for a document to fail on one backend while being OK on others.
2. Non-conforming man pages are still viewable, even by the `man` command.
3. Such man pages are useful as plain-text documents, even if they're ineligible for entry into an indexable man-page library.

PR submitted.

mojavelinux:
I like this idea. I'm not sure yet about the use of the attributes `man-first-section-name` and `man-first-section-line`. I'm going to think a bit more about an alternative approach for automatic recovery.

ds26gte:
Here's the rationale in case the output for some of your test cases is confusing. I wanted to be able to take _any_ valid Asciidoctor file, say `jobname.adoc`, and do
```
asciidoctor -b manpage jobname.adoc
man -l jobname.n                                      # where n = deduced volume number
```
and view something whose content is faithful to `jobname.adoc`.

Right now, either Asciidoctor fails to produce a `jobname.n`, or if the input is malformed but still processable, it will mess with the content in the following ways:

- change the level of the first section if it isn't 1
- change the name of the first section if it isn't `NAME`
- toss the second and following lines in the first section, if there any

I think these are intrusive changes, so I corrected for them with those strange-looking `document.attributes`.

mojavelinux:
Thanks for clarifying the requirements. That's extremely useful when looking for the right solution.

mojavelinux:
I've decided to take a slightly different approach. I still think the error messages are useful since the purpose is to generate a document from a specific document type. But I agree that if it doesn't conform, the converter should still produce an output document. So I've just filled in some more sensible defaults to ensure it succeeds. See #2937.

mojavelinux:
I also changed the parser so it doesn't drop the first section if it's non-conforming.

PR #1640, opened 2015-12-29
Manpage default boilerplate

ds26gte:
Have the `manpage` backend generate a man page even for input documents that don't strictly conform to man standard.  Addresses issue #1639.


mojavelinux:
I'm going to remove trailing spaces as a separate commit. Can you rebase?

PR #1646, opened 2015-12-31
resolves #1561 recognize file:/// as a file root in Opal browser env

### LINKED ISSUE #1561: Local path with file:// protocol are resolved as relative paths
Local path with file:// protocol are resolved as relative paths.

When running Asciidoctor.js in a browser environment, local paths starting with `file://` should be resolved as local path.

One use case is to be able to read csv file for the chart extension:

```
chart::sample.csv[line]
```

Exception when running in Asciidoctor.js and safe mode = `UNSAFE`:
```
Exception : asciidoctor: FAILED: : Failed to parse source, (intermediate value)(intermediate value)(intermediate value).$readable? is not a function
```

Exception when running in Asciidoctor.js and safe mode = `SAFE` or `SERVER`:

```
SecurityError : asciidoctor: FAILED: : Failed to parse source, Jail is not an absolute path: ...
```

mojavelinux:


mojavelinux:
@Mogztter I think that will do the trick. I'm sure there are cases we aren't yet considering, but it at least gives us a start. We can reevaluate as we discover corner cases.

Technically, we could add this as an override in the Asciidoctor.js code base (monkeypatch), but for now it was a small enough change that I could make it to core. If you are more comfortable with the patch in Asciidoctor.js, I'm open to putting it there instead. Just let me know your opinion.

I tested this with the chart block macro extension and it successfully reads the file. I can see the data in the generated HTML. (It doesn't show the chart, but that's because the sample code isn't evaluating the embedded script tag).

mojavelinux:
@Mogztter wdyt, look good to you?

lib/asciidoctor/path_resolver.rb
```diff
@@ -255,6 +258,11 @@ def partition_path path, web_path = false
     # shift twice for a UNC path
     if root == DOUBLE_SLASH
       path_segments = path_segments[2..-1]
+    # shift twice for a file:/// path and adjust root
```
Mogztter: Why do we keep this commented lines ?

lib/asciidoctor/path_resolver.rb
```diff
@@ -145,6 +145,9 @@ def is_root? path
     # Windows roots can begin with drive letter
     elsif @file_separator == BACKSLASH && WindowsRootRx =~ path
       true
+    # Absolute paths in the browser start with file:///
+    elsif ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_PLATFORM == 'browser' && (path.start_with? 'file:///')
+      true
```
Mogztter: :+1:

Mogztter:
> Technically, we could add this as an override in the Asciidoctor.js code base (monkeypatch), but for now it was a small enough change that I could make it to core. If you are more comfortable with the patch in Asciidoctor.js, I'm open to putting it there instead. Just let me know your opinion.

Why not, single line monkeypatch are easy to setup in the Asciidoctor.js build but since you already made this change in core, we can keep it here for now.

> I tested this with the chart block macro extension and it successfully reads the file. I can see the data in the generated HTML

Cool :smile: 

> It doesn't show the chart, but that's because the sample code isn't evaluating the embedded script tag

Hmmm :confused: 

> wdyt, look good to you?

Yes this looks good!

lib/asciidoctor/path_resolver.rb
mojavelinux: To remind myself I may want to implement it in the future. I haven't decided yet whether the code is correct how it is or whether this is needed. Need more time to think about it.

mojavelinux:
> Why not, single line monkeypatch are easy to setup in the Asciidoctor.js build

In this case, it would involve appending logic to an existing method, so it's a bit trickier. You'd implement it by using the `append` feature to inject a layer into the class hierarchy. Then you can call super. If super returns false, you do an additional check for the file:/// prefix and return true when in the browser environment.

But I want to experiment with having the logic in core so that I can remember to think about the impact of changes on paths.

atbest:
@mojavelinux sorry to bring this up, but I think `is_root?` should also return `true` for a `file://` path in other situation. There are two reasons for this, 1) `:` is not a valid path character, 2) `file://` is always followed by absolute path. 

Based on the two reasons, people use a directory starts with `file://`, they do this on purpose, but `system_path` will append the path to `work_dir`, which is unexpected, although I can check the prefix before calling `system_path` method.

I came across this problem when I was previewing the rendered html file with firefox, which cannot load css file from `<link rel="stylesheet" href="D:/document/styles/asciidoctor.css">`. The stylesheet was applied when I use `<link rel="stylesheet" href="file:///D:/document/styles/asciidoctor.css">`.

Issue #1732, opened 2016-04-21
Debian package is out of date, updating with gem installs parallel version in secondary path


oddhack:
The Debian package for Jessie is still at 0.1.4. It's possible to use gem update to get 1.5.4 installed, but doing so results in /usr/bin/asciidoctor still being the 0.1.4 Debian version, while /usr/local/bin/asciidoctor is the new version.

Presumably there's some way to configure gem to overwrite the Debian versions (which might or might not be a desirable default), but as it isn't obvious this is what's happening, newbies such as myself get very confused trying to find the new version.

I suggest enhancing the README.adoc to note this issue.

mojavelinux:
The rule of thumb is that `gem update` should never be used for a gem that is managed with a package manager. That gets you into undefined territory in the Linux world. That's true of all gems, but also for all module managers in other languages (Perl, Python, etc).

mojavelinux:
> Presumably there's some way to configure gem to overwrite the Debian versions (which might or might not be a desirable default)

The answer is rvm. System gems should be managed only with the system package manager. User-space gems should be managed by rvm (or something equivalent). That's the position of the Asciidoctor project. We simply can't afford to support other scenarios because it's just too difficult to provide assistance.

mojavelinux:
Asciidoctor 1.5.x is only in stretch and sid. We should link to the following page to show the full picture:

https://packages.debian.org/search?keywords=asciidoctor&searchon=names&suite=all&section=all

oddhack:
@mojavelinux unfortunately, the advice given on the front page of the project for installation results in an out-of-date install, and trying to follow any other advice on that page apparently gets me into "officially unsupported" territory (I have no clue what "rvm" is, but it sure isn't mentioned there).

I'm not arguing with your policy, but the install advice given for Debian is simply not helpful or accurate, and should be fixed.

mojavelinux:
I agree we need to provide better instructions and a more accurate description of the current state. There's just a certain way I'd like to explain it so it gives people the best path to success. Specifically, it's important to keep a clear separation between the system and user space, and to understand why there's a difference. If we aren't careful with those instructions, people can find themselves in hot water very quickly.

I'll submit a proposed revision to your PR.

oddhack:
Cool. At the end of the day I just want to be able to run the current asciidoctor, without knowing anything about Ruby (which I, basically, don't), as I expect others do.

Mogztter:
[...]

PR #1733, opened 2016-04-21
Add a tip noting that the Debian package is out of date

oddhack:
And that using `gem update asciidoctor` results in both the older and newer versions being installed on
different paths (issue #1732).

I hope changing README.adoc in master will propagate through to the asciidoctor.org home page?


mojavelinux:
Merged as 22310d93ccdb11078fccdbc3feba5243dc904c26.

I do want to note that we have a direct guide for using RVM planned. See https://github.com/asciidoctor/asciidoctor.org/issues/399. We'll link to that when ready.

PR #1744, opened 2016-05-07
Translate Attributes to German

fwilhe:
I must admit I'm not sure how one would translate "Listing", but I guess using the English loanword is common today.

### DIFF
--- a/data/locale/attributes.adoc
+++ b/data/locale/attributes.adoc
@@ -2,8 +2,22 @@
 // Please add your language!
 
 ifeval::["{lang}" == "de"]
+:appendix-caption: Anhang
 :caution-caption: Achtung
-//...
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks @fwilhe!

Two small change requests.

Could you put a line comment in front of the preface-title and listing-caption entries? These are off by default in Asciidoctor. Turning them on changes behavior, so we should leave the update to the user.

Could you also rebase since I caused a merge conflict.

fwilhe:
@mojavelinux Done.

I've wondered why the lines are commented, maybe this should be explained at the beginning of this file.

mojavelinux:
Thanks @fwilhe!

Good point. I'll add a note at the top of the file about why those attributes are behind comments.

PR #1745, opened 2016-05-07
French translations for built-in labels

ncomet:
Added french translations for built-in labels

### DIFF
--- a/data/locale/attributes.adoc
+++ b/data/locale/attributes.adoc
@@ -34,10 +34,10 @@ ifeval::["{lang}" == "fr"]
 :figure-caption: Figure
 :important-caption: Important
 :last-update-label: Dernière mise à jour
-//:listing-caption: Liste
+:listing-caption: Liste
 :manname-title: Nom
 :note-caption: Note
-//:preface-title: Préface
+:preface-title: Préface
 :table-caption: Table
 :tip-caption: Suggestion
 :toc-title: Table des matières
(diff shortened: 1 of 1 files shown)


data/locale/attributes.adoc
```diff
@@ -25,3 +25,23 @@ ifeval::["{lang}" == "it"]
...
+:example-caption: Exemple
+:figure-caption: Figure
+:important-caption: Important
+:last-update-label: Dernière mise à jour
+:listing-caption: Liste
+:manname-title: Nom
+:note-caption: Note
+:preface-title: Préface
+:table-caption: Table
+:tip-caption: Suggestion
```
Mogztter: Je pense que suggestion est un peu trop générique . Peut être "Astuce" ?

mojavelinux:
Merci beaucoup !

data/locale/attributes.adoc
mojavelinux: I missed this comment before I merged. I agree, though, I like a translation like "trick" instead of the more subtle "suggestion".

mojavelinux: Updated.

Shouldn't Table be Tableau? Or do you not translate the word for data table?

ncomet: Yes you are right. I also like more "Astuce" meaning "trick", I was hesitating. And it definitely should be "Tableau".

ncomet: I'll open a new pull request

mojavelinux: :+1:

PR #1746, opened 2016-05-07
French translations update after remarks

ncomet:
I took into account both remarks about "Tableau" and "Astuce"


ncomet:
I have a conflict, but Tableau is definitely better than Table, I'll go on when I have access to my laptop.

PR #1748, opened 2016-05-07
changed according to remark on Tableau

ncomet:
I changed Table to Tableau, since it is more accurate.


mojavelinux:
Thanks @ncomet!

PR #1754, opened 2016-05-08
Adding Russian translation for build-in attributes

AlexanderZobkov:
Adding Russian translation for build-in attributes.
Fixing broken `ifeval `and `endif `pairs.


mojavelinux:
Thanks for the translations and for fixing the broken pairs! Merged as 635a19e912c2f3b3d1463ce1742635800fef2378.

PR #1758, opened 2016-05-09
add Simplified Chinese translation for build-in attributes

dongwq:


mojavelinux:
Thanks for lending your bilingual skills, @dongwq!

PR #1770, opened 2016-05-14
Updated Italian translation

ciampix:
I've changed my mind.

If the "Chapter" string is present "by default" it is just right to translate it. If I do not want it, it is easy to un-assign the string for all languages altogether.


mojavelinux:
I agree. If/when we decide to switch it off by default in Asciidoctor PDF, we can change it back.

Thanks!

PR #1780, opened 2016-05-24
Fix for git ls-files filter

rahmanusta:


mojavelinux:
Aha! I can totally see why the previous command was not working on Windows. It was relying on Bash expansion. Thanks for figuring this out @rahmanusta!

rahmanusta:
You are welcome @mojavelinux :) :)

mojavelinux:
Thanks @rahmanusta!

I decided to take it a step further and use Ruby to filter the results. The result of the command is already split into an array, so this is just a matter of adding a grep call to the tail. The logic is also more accurate when git is not installed.

See https://github.com/asciidoctor/asciidoctor/blob/567830203d9052f226813f263a427fd89f404f19/asciidoctor.gemspec#L14-L18.

PR #1787, opened 2016-06-03
translate README into Japanese

Mizuho32:


mojavelinux:
Thanks @Mizuho32!

Mizuho32:
Not at all.

PR #1801, opened 2016-06-23
Support alternate font icon sets

### LINKED ISSUE #539: Support alternate font icon sets
With the release of Bootstrap 3, Glyphicons become an external project (http://glyphicons.getbootstrap.com/) similar as Font Awesome. It is true that is not as powerful as Font Awesome but it could a good addition to Asciidoctor icons system.

Mogztter:
Inspired by [Asciidoctor PDF](https://github.com/asciidoctor/asciidoctor-pdf/blob/876b4e480b1d7cfe3c24f46a5f1174fec1df1382/lib/asciidoctor-pdf/prawn_ext/extensions.rb#L11).

#### Naming

Asciidoctor PDF is using `icon-set` but Asciidoctor "core" is using `iconfont-` suffix:
- `iconfont-remote`
- `iconfont-cdn`
- `iconfont-name`

I've decided to stay consistent with Asciidoctor PDF but I don't know if it's the right choice ?

#### Using multiples font icon sets ?

We still need to load Font Awesome even if we are using an alternative icon set because [checked lists are using Font Awesome icons](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L879-L880).

``` ruby
marker_checked = '<i class="fa fa-check-square-o"></i> '
marker_unchecked = '<i class="fa fa-square-o"></i> '
```

Having said that, I think it would be nice to be able to load multiple icon sets.
For instance, Payment font is really specific, and I think users will want to load more than one icon sets.

Another cool feature in Asciidoctor PDF, which is missing from Asciidoctor "core", is the `set` attribute:

```
:icon-set: octicon

icon:pf-american-express[set=pf]
```

In the example above, I can use Payment font for a specific inline icon. 
The main issue is that we will need to "register" every font icon sets needed and then "update" the HTML document header to add the corresponding `<link>`.
@mojavelinux Any thoughts on how to implement this feature ?

Resolves #539

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -22,6 +22,24 @@ module Asciidoctor
     SvgPreambleRx = /\A.*?(?=<svg\b)/m
     SvgStartTagRx = /\A<svg[^>]*>/
     DimensionAttributeRx = /\s(?:width|height|style)=(["']).*?\1/
+    IconSets = {
+      'fa' => {
+        :cdn_path => 'font-awesome/4.6.3/css/font-awesome.min.css',
+        :default_name => 'font-awesome'
+      },
+      'fi' => {
+        :cdn_path => 'foundicons/3.0.0/foundation-icons.min.css',
+        :default_name => 'foundation-icons'
+      },
+      'octicon' => {
+        :cdn_path =>'octicons/3.5.0/octicons.min.css',
+        :default_name => 'octicons'
+      },
+      'pf' => {
+        :cdn_path => 'paymentfont/1.1.2/css/paymentfont.min.css',
+        :default_name => 'paymentfont'
+      }
+    }
 
     def initialize backend, opts = {}
       @xml_mode = opts[:htmlsyntax] == 'xml'
@@ -73,11 +91,19 @@ module Asciidoctor
       end
 
       if node.attr? 'icons', 'font'
+        valid_icon_sets = (node.attr 'icon-sets', 'fa').split(',').select { |name| IconSets.include? name }
+        icon_sets = (valid_icon_sets.empty? ? ['fa'] : valid_icon_sets)
         if node.attr? 'iconfont-remote'
-          result << %(<link rel="stylesheet" href="#{node.attr 'iconfont-cdn', %[#{cdn_base}/font-awesome/4.6.3/css/font-awesome.min.css]}"#{slash}>)
+          icon_sets.each do |name|
+            icon_set = IconSets[name]
+            result << %(<link rel="stylesheet" href="#{node.attr "iconfont-#{name}-cdn", %[#{cdn_base}/#{icon_set[:cdn_path]}]}"#{slash}>)
+          end
         else
-          iconfont_stylesheet = %(#{node.attr 'iconfont-name', 'font-awesome'}.css)
-          result << %(<link rel="stylesheet" href="#{node.normalize_web_path iconfont_stylesheet, (node.attr 'stylesdir', ''), false}"#{slash}>)
+          icon_sets.each do |name|
+            icon_set = IconSets[name]
+            iconfont_stylesheet = %(#{node.attr "iconfont-#{name}-name", icon_set[:default_name]}.css)
+            result << %(<link rel="stylesheet" href="#{node.normalize_web_path iconfont_stylesheet, (node.attr 'stylesdir', ''), false}"#{slash}>)
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/converter/html5.rb
```diff
@@ -75,6 +76,15 @@ def document node
       if node.attr? 'icons', 'font'
         if node.attr? 'iconfont-remote'
           result << %(<link rel="stylesheet" href="#{node.attr 'iconfont-cdn', %[#{cdn_base}/font-awesome/4.6.3/css/font-awesome.min.css]}"#{slash}>)
+          icon_set = node.document.attr 'icon-set', 'fa'
+          icon_set = 'fa' unless IconSets.include? icon_set
```
Mogztter: This code is duplicated but I think we don't have the choice ?

mojavelinux:
I definitely want to be able to support this. Obviously, it's a little trickier since, unlike Asciidoctor PDF, we have to load the related font/stylesheet. But as you pointed out, we have the chicken-egg problem that we don't know which fonts we're going to use in the document. What we could require is that all fonts that you plan to use are declared in the `icon-sets` attribute. wdyt?

Mogztter:
@mojavelinux I think this is the way to go :+1: 
What is the syntax for an attribute with multiple values ? comma separated values ? `:icon-sets: octicon,pf` ?

Mogztter:
## Important changes

The following attributes have been renamed (for consistency, though we can support backward compatibility):
 * `iconfont-name` ->`iconfont-fa-name` 
 * `iconfont-cdn` ->  `iconfont-fa-cdn` 

If no `set` is defined on an icon directive and multiple font sets are available, we are using the "first" font defined.

In this example, `octicon` is the first font defined in `icon-sets`:
```adoc
:icons: font
:icon-sets: octicon,pf

icon:bitcoin[set=pf] // <1>
icon:octoface[size=mega] // <2>

// <1> we are using payment font because we explicitly choose to use payment font (pf) with the set attribute
// <2> we are using octicon font because not set is defined and octicon is the first font defined
```

Mogztter:
I came across [Feather](https://github.com/colebemis/feather) and I really like the approach and the design.
Since Feather is using SVG icons, there's no need to load CSS or web fonts into the final HTML document, we could replace `icon:heart[]` with [heart.svg](https://raw.githubusercontent.com/colebemis/feather/master/icons/heart.svg) directly in the HTML5 converter.

One way to support SVG icon could be to define `icons` attribute to `svg`:

```
:icons: svg
:icon-sets: feather

icon:heart[]
```

If we don't want to replace icon with SVG in the converter, we can use a JavaScript file to replace HTML tag with the corresponding SVG: https://github.com/colebemis/feather#quick-start

I'm bringing this idea to this issue, so that we can make sure the chosen solution will be generic enough to integrate no font-based icon sets.

elextr:
Does this need to be documented somewhere?

Mogztter:
@elextr Sure but first I think we should decide what we want to support:

* Web Fonts with CSS (client side rendering)
* SVG icons with JS (client side rendering)
* SVG icons (during the conversion phase) 

We can choose to support only one or all three. We also need to decide which icons libraries will be integrated in Asciidoctor by default and how to add a new icons library (extension point ?)
It could be nice to be able to resolve the icon via an extension, especially in JavaScript environment... but I digress! 

If we don't want to refactor this feature too much, we could limit the scope of this pull request to just "allow the user to use multiple Web Fonts for icons".

Note, I just saw that since version 5 Octicons is also using SVG :wink:

elextr:
@Mogztter sorry I thought you were more advanced than that.  Any way just prompting :)

As for which one to support, as a wise man once said, when offered a choice, take all.

Mogztter:
@mojavelinux I would love your feedback on this pull request :smile:

mojavelinux:
I definitely want to revisit how we're handling icons, but I don't want to rush a solution trying to get 2.0 out. This is exactly the type of issue that I hope the switch to semver will give us room to work on.

Mogztter:
@mojavelinux I'm revisiting this feature again. Maybe we should introduce a pluggable adapter for integrating icons libraries?

Pseudo-code:

```rb
module Asciidoctor
module IconLibrary
  def docinfo? location; end
  def docinfo location, doc, opts; end
  # Indicates whether rendering is handled by this library or by the client.
  def render?; end
  # Indicates whether this icon exists in this library.
  def exists? name; end
  def convert node, name, opts; end
  def register_for *names; end
end
end

module Asciidoctor
class IconLibrary::FontAwesome4Adapter < IconLibrary
  register_for 'fontawesome', 'fontawesome4'

  FONT_AWESOME_VERSION = '4.7.0'

  def docinfo? location
      location == :head
  end

  def docinfo location, doc, opts
    unless (asset_uri_scheme = (doc.attr 'asset-uri-scheme', 'https')).empty?
      asset_uri_scheme = %(#{asset_uri_scheme}:)
    end
    cdn_base_url = %(#{asset_uri_scheme}//cdnjs.cloudflare.com/ajax/libs)
    if node.attr? 'iconfont-remote'
        %(<link rel="stylesheet" href="#{doc.attr 'iconfont-cdn', %[#{cdn_base_url}/font-awesome/#{FONT_AWESOME_VERSION}/css/font-awesome.min.css]}"/>)
      else
        iconfont_stylesheet = %(#{doc.attr 'iconfont-name', 'font-awesome'}.css)
        result << %(<link rel="stylesheet" href="#{doc.normalize_web_path iconfont_stylesheet, (doc.attr 'stylesdir', ''), false}"/>)
      end
  end

  def render?
    false # client-side rendering using CSS and a custom font
  end

  def exists? name
    true
  end

  def convert node, name, opts
    title_attr = (title = opts[:title]) ? %( title="#{opts[:title]}") : ''
    %(<i class="fa #{name}"{title_attr}></i>)
  end
end
```

What do you think?

PR #1817, opened 2016-07-15
Translate built-in attributes to Serbian

BojanStipic:
For both Cyrillic and Latin alphabet


BojanStipic:
Should I also translate `chapter-label`? I see it listed in [user manual](http://asciidoctor.org/docs/user-manual/#customizing-labels), but it is not translated in any language under `data/locale/attributes.adoc`

mojavelinux:
Thank you! And yes, please translate chapter-label as well. But leave the line disabled using a line comment.

mojavelinux:
Thanks @BojanStipic!

PR #1818, opened 2016-07-18
skip-front-matter: also support TOML front matter

cyrilRomain:


cyrilRomain:
I just noticed several JSON front matter syntax exist :
 - ‘{’ and ‘}’ in hugo :  https://gohugo.io/content/front-matter/
 - '{{{' and '}}}' in json-front-matter : https://github.com/jsantell/node-json-front-matter

cyrilRomain:
One could even make asciidoctor use gray-matter, a fast front-matter parser that supports multiple front matter formats : https://github.com/jonschlinkert/gray-matter
But it may be not worth the additionnal dependency.

mojavelinux:
Thanks for submitting this patch!

I think we need to draw a line somewhere because it's not the job of Asciidoctor to parse syntax that isn't AsciiDoc. I'm okay with supporting some of the additional front matter forms, but we need some agreement in this industry about how front matter is defined. The single bracket JSON is entirely too risky in my opinion. I think Hugo should change to the triple bracket syntax.

I'm still deciding whether I want to support the triple bracket syntax. Regardless, anything besides the forms we pick can handled with a Preprocessor extension. We're also going to require them to be in block form (delimiters on separate lines).

I'm going to propose a PR that includes a slightly more efficient way of writing this...which also extracts these delimiters into a constant (or constants). I need to fiddle a bit.

cyrilRomain:
Thanks for the feedback. I’m completely in line with it.

sanmai-NL:
What's the current status? Note that TOML frontmatter is in common use with the popular Hugo static website generator.

mojavelinux:
Perhaps we should consider "standardizing" on the gray matter syntax, which is used in Metalsmith and Phenomic. https://www.npmjs.com/package/gray-matter

mojavelinux:
Superseded by the discussion in #4300, which focuses on Hugo front matter.

PR #1829, opened 2016-07-31
use triple equals operator instead of is_a?

mojavelinux:


Mogztter:
I'm curious, is this a best practice documented somewhere ?

mojavelinux:
This is the style I prefer to use for Ruby. It's the style I've been using throughout the Asciidoctor ecosystem, but some older code still wasn't using it.

I've been documenting the Asciidoctor Ruby styles here in the jekyll-asciidoc project.

https://github.com/asciidoctor/jekyll-asciidoc#coding-style

I plan to migrate these somewhere more appropriate once I record most of the rules I've been following.

mojavelinux:
As I've mentioned before, the motivation for my coding style is a) speed and b) readability. A simple benchmark reveals that triple equals is the most optimized type check in various Ruby implementations.

jirutka:
> A simple benchmark reveals that triple equals is the most optimized type check in various Ruby implementations.

Are you sure about this? I’d expect exactly opposite; `is_a?` is well defined single purpose method, `===` is much more complicated.

mojavelinux:
I'm quite certain, but might check again anyway. MRI in particular tends to
rewrite special operators like these directly in C code, whereas is_a?
remains interpreted at a higher level. Notice I put the type first, which
is the optimized scenario.

mojavelinux:
I did some more tests and I'm unable to discern a performance difference between === and is_a? in the latest version of Ruby MRI. So this PR becomes exclusively about consistency. In some places, we were using === and others is_a?. Now we are only using ===.

I tend to prefer that syntax because it never requires parenthesis. If we do decide to switch to is_a? in the future, I would just want to make sure we go full sail.

Mogztter:
Thanks Dan for the explanation. I will do some benchmarks with Opal to see
if there's a notable difference.

PR #1831, opened 2016-08-02
resolves #1830 set reversed attribute on ol if reversed option is enabled

### LINKED ISSUE #1830: Add support for reversed option on ordered list
Add an option flag to reverse the order of the numbers in an ordered list (descending instead of ascending). In the HTML converter, this option would set the `reversed`  HTML5 attribute on the `<ol>` element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attr-reversed

mojavelinux:


Mogztter:
I think we should create an issue in the asciidoctor.org repository to document this feature or maybe wait for the release note ?

mojavelinux:
I agree. Let's make sure this gets documented. Do you want to file the issue or would you like me to do it?

Mogztter:
Ok done: https://github.com/asciidoctor/asciidoctor.org/issues/602

mojavelinux:
Thanks @Mogztter!

Issue #1729, opened 2016-04-17
Enable tagging include regions in circumfix comments, beyond line comments


sanmai-NL:
At least one important programming language, OCaml, [only supports circumfix comment delimitation](https://ocaml.org/learn/tutorials/basics.html#Comments). Asciidoctor currently only allows tagged regions in included content where tags immediately precede the end-of-line marker.

``` ocaml
(* tag::mytag[] *)
let problem = "This code sample will not be included. ";;
(* end::mytag[] *)
```

``` ocaml
(* tag::mytag[]
*)
let problem = "This will include a syntactically incorrect code sample. ";;
(* end::mytag[]
*)
```

I propose that Asciidoctor gets support for including tagged regions even in these cases. One solution is perhaps to reimplement tag detection so that lines with four or less non-whitespace, non-tag characters are accepted as tag markers.

mojavelinux:
True, this is a limitation. Right now, we only support XML circumfix comments using a hack. (See https://github.com/asciidoctor/asciidoctor/blob/v1.5.4/lib/asciidoctor/reader.rb#L950-L951).

> One solution is perhaps to reimplement tag detection so that lines with four or less non-whitespace, non-tag characters are accepted as tag markers.

This wouldn't work for XML since the closing delimiter is `-->`. We'd need to match a few extra characters.

The reason we match in the way we do today is for speed, which for Asciidoctor is paramount. `end_with?` is much, much faster than a regular expression match.

The right solution may be to make the hack more generic. Use `chomp` appropriate for the file extension of the include. For XML, it's `-->`. For OCaml, it's `*)`. That way, we can still use the logic but get the flexibility of smart truncation. It would also speed things up for non-XML languages since we would only `chomp` for certain file types.

mojavelinux:
> At least one important programming language, OCaml

While OCaml is certainly important, I'd argue that XML and CSS provide even more compelling justification for supporting this feature.

mojavelinux:
Note that there's a similar problem with callout numbers in source code listings.

mojavelinux:
I was able to abstract out the logic so that the suffix part of the circumfix comment is ignored if the extension of the include file matches one of the entries in the circumfix comments map. As far as the parser is concerned, the tag looks like:

```
(* tag::snippet[]
```

even though it's written like:

```
(* tag::snippet[] *)
```

The requirement is that there must be exactly one space between the end of the tag directive and the start of the comment suffix.

[...]

PR #1856, opened 2016-08-30
fix tags for languages like Mathematica, OCaml or F#

bobzhang:
in such languages, comments are like this

``` ocaml
let x = 3 (* this is a line comment *)
```

note this is a proposed fix, I don't really know how asciidoctor works, maybe you can have better fix

### DIFF
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -949,6 +949,7 @@ class PreprocessorReader < Reader
                 l = l.rstrip
                 # tagged lines in XML may end with '-->'
                 tl = l.chomp('-->').rstrip
+                tl = l.chomp('*)').rstrip
                 if active_tag
                   if tl.end_with?(%(end::#{active_tag}[]))
                     active_tag = nil


mojavelinux:
Are you trying to define an include tag in a comment?

```
(* tag::example-1[] *)
let x = 3
(* end::example-1[] *)
```

That's where this comes into play.

bobzhang:
Yes, the example you used will not work for current ascidoctor, it will work with my fix though

mojavelinux:
That's correct. I just wanted to confirm we were on the same page.

It's not a good idea to truncate `*)` from every file as it's error prone and impacts performance. In fact, we shouldn't even be doing that for `-->`. Instead, we should have a map that looks for tags using a special matcher based on file extension...falling back to the default matcher in the general / unknown case.

mojavelinux:
Can you provide a list of file extensions that should be matched in this way?

bobzhang:
you are right, so I guess you probably have a better fix.
currently I am using ocaml, but F# should also apply, so `.ml`, `.mli`, `.fs`, `.fsi`, `.fsx` should apply

mojavelinux:
We can keep this PR open as it serves as the issue (and a suggested impl).

bobzhang:
any progress? currently I have a local patched asciidoctor running, it would be nice we can remove this patch in the future, thanks

mojavelinux:
Not yet. I'm still working to stabilize core for the major upgrade to Asciidoctor.js. I'm going to do my best to tackle this issue for 1.5.7.

mojavelinux:
The related issue is #1729.

mojavelinux:
F# seems to support line comments, so I'm not sure how that applies here (otherwise, we're talking about supporting block comments across all languages, which is a much larger list and task).

mojavelinux:
This will make it into 1.5.6.

mojavelinux:
A fix is now available. I kindly request your review. See https://github.com/asciidoctor/asciidoctor/pull/2199/files.

PR #1868, opened 2016-09-11
resolves #1691 support nested emphasis in HTML

### LINKED ISSUE #1691: Support nested <em>
In html we can nest `<em>` elements:

```html
<p>Some normal text <em>Emphatic text <em>with an emphatic word inside</em> and other emphatic text</em></p>
```

This way we preserve the semantic info that `<em>with an emphatic word inside</em>` is emphatic, that instead we lose when doing:

```
Some normal text _Emphatic text_ with an emphatic word inside _and other emphatic text_
```

bk2204:
We do emit `i` and `b` in some cases, but not for the purpose of adding emphasis or bold, so I didn't bother styling those, in an attempt to keep the stylesheet small.


mojavelinux:
Merged as 455a0e86d698027d952e5f8eff4578869445567a. Thanks!

Issue #1895, opened 2016-10-20
Image callout icons disappear for long callouts [html5]


cristoper:
When the `icons` attribute is set to "image" with the html5 backend, then the callout icons in callout table shrink and then disappear as any of the callout text wraps to more than one line.

For example, this asciidoc file:

``` [source,python]
----
def normalize_tokenize(text):
    norm = text.lower() #<1>
    norm = re.sub(r'(?u)[^\w\s]', '', norm) #<2>
    return tokens
----
<1> First we coerce the entire string to lower case using the `lower()` method of Python's standard https://docs.python.org/3/library/stdtypes.html#str[string] type.

<2> This https://docs.python.org/3/library/re.html[regular expression] matches everything that is not a word (`\w`) or whitespace (`\s`) and replaces it with the empty string. The `(?u)` flag at the start enables Unicode matching (this is only needed in Python 2; Python 3 matches Unicode by default).
```

Looks like this when rendered (Chromium version 48.0.2564.82 on Linux):

![bug](https://cloud.githubusercontent.com/assets/331548/19545852/0ed5e1ec-9647-11e6-867f-c247e86e92a8.png)

I think the fix is to set the 'max-width' CSS property of icon img's to 'none' as is done for admonitionblock icons. Following the example of admonitionblock, the patch in the subsequent pull request adds an 'icon' class to the <td> element containing the callout icon and a corresponding line in the default stylesheet.

With that fix, the above snippet renders as expected:

![expected](https://cloud.githubusercontent.com/assets/331548/19545867/3fcdde58-9647-11e6-8077-fa8af4b8f338.png)

mojavelinux:
Let me know if the fix works for you.

cristoper:
Yep, your CSS change in aa3e33e fixes it for me. Thanks!

mojavelinux:
Great! Thanks for testing!

PR #1896, opened 2016-10-20
Pull request to fix Issue 1895

cristoper:
I don't think I remembered how to use github pull requests very well. This is the issue I was hoping this pull request would be associated with: https://github.com/asciidoctor/asciidoctor/issues/1895


mojavelinux:
I decided to fix this slightly differently. I didn't want to add the CSS class to the td. As it turns out, we already had a CSS rule for styling that cell. See https://github.com/asciidoctor/asciidoctor/pull/1919. Thanks for getting it started!

PR #1898, opened 2016-10-22
Use a special logic for include processing only on a browser environment

Mogztter:
This special case is unnecessary (or at least not necessary anymore on Node).

Should fix this issue: https://github.com/asciidoctor/atom-asciidoc-preview/issues/189

### DIFF
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -844,10 +844,9 @@ class PreprocessorReader < Reader
         warn %(asciidoctor: ERROR: #{line_info}: maximum include depth of #{@maxdepth[:rel]} exceeded)
         return false
       end
-      if ::RUBY_ENGINE_OPAL
+      if ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
         # NOTE resolves uri relative to currently loaded document
         # NOTE we defer checking if file exists and catch the 404 error if it does not
-        # TODO only use this logic if env-browser is set
         target_type = :file
         include_file = path = if @include_stack.empty?
           ::Dir.pwd == @document.base_dir ? target : (::File.join @dir, target)
--- a/lib/asciidoctor/path_resolver.rb
+++ b/lib/asciidoctor/path_resolver.rb
@@ -132,24 +132,24 @@ class PathResolver
     @_partition_path_web = {}
   end
 
-  # Public: Check if the specified path is an absolute root path
-  # This operation correctly handles both posix and windows paths.
+  # Public: Check if the specified path is an absolute root path.
+  # This operation considers posix paths, windows paths, and file URLs.
+  #
+  # Unix absolute paths and UNC paths start with slash. Windows roots can start
+  # with a drive letter. Absolute paths in the browser start with file://.
   #
   # path - the String path to check
   #
   # returns a Boolean indicating whether the path is an absolute root path
-  def is_root? path
-    # Unix absolute paths and UNC paths start with slash
-    if path.start_with? SLASH
-      true
-    # Windows roots can begin with drive letter
-    elsif @file_separator == BACKSLASH && (WindowsRootRx.match? path)
-      true
-    # Absolute paths in the browser start with file://
-    elsif ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_PLATFORM == 'browser' && (path.start_with? 'file://')
-      true
-    else
-      false
+  if RUBY_ENGINE == 'opal'
+    def is_root? path
+      (path.start_with? SLASH) ||
+          (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (path.start_with? 'file://')) ||
+          (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
+    end
+  else
+    def is_root? path
+      (path.start_with? SLASH) || (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
     end
   end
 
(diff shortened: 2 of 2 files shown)


lib/asciidoctor/reader.rb
```diff
@@ -835,10 +835,9 @@ def preprocess_include raw_target, raw_attributes
         warn %(asciidoctor: ERROR: #{line_info}: maximum include depth of #{@maxdepth[:rel]} exceeded)
         return false
       end
-      if ::RUBY_ENGINE_OPAL
+      if ::RUBY_ENGINE_OPAL && @document.attributes.has_key? 'env-browser'
```
Mogztter: Could we also check if the global variable `JAVASCRIPT_PLATFORM` matches `browser` ? https://github.com/asciidoctor/asciidoctor.js/pull/248/files#diff-c3680ed68dbcf2cc0b36fa1af7000debR62

By default, I think Asciidoctor should choose the "right way" to process include files (and if it fails, users can always set the "env-browser" attribute). Wdyt ?

Mogztter: @mojavelinux outdated diff because I forgot to add parenthesis but the comment is still relevant and I will amend my commits once we have agreed

mattadamson:
@Mogztter how can I actually try and build this locally and use in my atom app?

Mogztter:
@mattadamson See my comment https://github.com/asciidoctor/atom-asciidoc-preview/issues/189#issuecomment-256622230

mojavelinux:
Can we / should we check for the window object instead?

```
if ::RUBY_ENGINE_OPAL && `typeof window !== 'undefined'`
```

Or perhaps the next configuration object you're proposing?

Mogztter:
`window` object is also defined in an Electron environment :unamused: 

But it made me think... is it a good idea to use an Asciidoctor attribute to control how the code should read a file at runtime ? I mean, I think Asciidoctor attribute should be limited to control how the content is rendered no ?

I opened a pull request on Asciidoctor.js to improve the auto-detection but also to make it possible to explicitly define the "runtime environment": https://github.com/asciidoctor/asciidoctor.js/pull/263

In Node.js we will be able to do:

```javascript
var config = {
  runtime: {
    platform: 'node',
    engine: 'v12',
    framework: 'lollipop',
    ioModule: 'xmlhttprequest'
  }
};
var asciidoctor = require('asciidoctor.js')(config);
```
In a browser environment I have not (yet) decided how users will be able to define the "runtime environment". The main issue is that we need to know the runtime environment when instantiating Asciidoctor.js. Unless we introduce a global variable like `AsciidoctorConfig` I do not know how to do it. We could of course use `Opal.ENV` but I want to hide as much as possible Opal because I think this is confusing for users and this is just an implementation detail.

Anyway, back to the point, I think you are right we should use `JAVASCRIPT_IO_MODULE`:

```ruby
if ::RUBY_ENGINE_OPAL && (JAVASCRIPT_IO_MODULE == 'xmlhttprequest')
```
I'm using `JAVASCRIPT_IO_MODULE` because this logic only make sense if we are using XMLHttpRequest right ?

Mogztter:
I think we should use `JAVASCRIPT_IO_MODULE == 'xmlhttprequest'` instead of `@document.attributes.has_key? 'env-browser'` because we can use [XMLHttpRequest](https://developer.mozilla.org/fr/docs/Web/API/XMLHttpRequest) in a non browser environment. 

Or maybe check both ?
```javascript
(@document.attributes.has_key? 'env-browser' || JAVASCRIPT_IO_MODULE == 'xmlhttprequest')`
```

As a browser environment will (always?) uses XMLHttpRequest.

mojavelinux:
I think we can just check the constant. You're correct, this is not something that should be tied to a document attribute.

The only change I might suggest is to combine the constants into a single Hash. Technically, constants can't be changed. But it is possible to change the entries in a Hash that is stored in a constant. So it perhaps should be something like:

```ruby
JAVASCRIPT_RUNTIME = {
  ioModule: %x(ioModule),
  platform: %x(platform),
  engine: %x(engine),
  framework: %x(framework)
}
```

That also makes it a little easier to report on.

Another idea is to mimic Ruby's [RbConfig](http://idiosyncratic-ruby.com/42-ruby-config.html#rbconfigconfig). Something like:

```ruby
module JsConfig
  CONFIG = {
    ioModule: %x(ioModule),
    platform: %x(platform),
    engine: %x(engine),
    framework: %x(framework)
  }
end
```

(btw, I kind of like ioLibrary or ioStrategy better than ioModule, but I might be over thinking it).

mojavelinux:
I reworked the patch to be based on the `JAVASCRIPT_IO_MODULE` constant. I also did some small optimizations.

I still strongly recommend that we consider defining something like `JsConfig::CONFIG`, perhaps as an alternative to these top-level constants. We can still support the old constants for backwards compatibility, but it seems cleaner to bundle this stuff together. But let's address that in a separate issue. I'll move forward with this path with your :+1: .

Mogztter:
> Another idea is to mimic Ruby's RbConfig. Something like...
> I still strongly recommend that we consider defining something like JsConfig::CONFIG, perhaps as an alternative to these top-level constants.

Yes polluting the global namespace is bad but I want to make sure that the `JsConfig` (`JsRuntime` ?) will be easily accessible in Asciidoctor.js:

```javascript
var Asciidoctor = require('asciidoctor.js');
console.log(Asciidoctor.JsConfig.platform); // prints nodejs
```

> I kind of like ioLibrary or ioStrategy better than ioModule

`ioLibrary` is nice :smile: 

> But let's address that in a separate issue. I'll move forward with this path with your +1

I agree, you can proceed :+1:

mojavelinux:
> Yes polluting the global namespace is bad but I want to make sure that the JsConfig (JsRuntime ?) will be easily accessible in Asciidoctor.js:

I agree that there's no real reason for the CONFIG map. Many people tend to extend RbConfig to do exactly what you're proposing. I think we should just start there. So JsConfig.platform and JsConfig.io_library (gotta be consistent w/ Ruby convention) (or perhaps just JsConfig.io) ftw!

mojavelinux:
Is there a way to make it top-level on the JavaScript side like Opal is? Or perhaps it should be Opal.JsConfig. On the Ruby side, it would definitely be a top-level class.

mojavelinux:
I'll proceed with this merge so that you can test. If we decide we want to make the switch before releasing Asciidoctor 1.5.6, I can always update the reference. What's important is that the logic is in there.

PR #1901, opened 2016-10-28
Update attributes.adoc - fragment identifier

stonio:
Update fragment identifier: http://asciidoctor.org/docs/user-manual/#customizing-labels


mojavelinux:
Thanks!

PR #1918, opened 2016-11-10
support anchors on labeled lists

buckelij:
This PR adds support for anchors on labelled lists with the `dtanchors` attribute, similar to the [sectanchors](http://asciidoctor.org/docs/user-manual/#anchors) attribute.

It depends on the the `sectid` attribute also being set, so it can form sensible `id`s within sections, but if that doesn't seem reasonable I'm happy to modify the behavior. It does not create anchors for 'nested' lists.

This is motivated by a desire to be able to easily refer people to a specific subsection of a Git man page within https://git-scm.com/docs (see https://github.com/git/git-scm.com/pull/849).


buckelij:
Is there any interest in this PR?

mojavelinux:
Yes, I've just been very busy lately. I filed a similar issue for discrete headings. See #1997. We may want to think about broadening this to all "auto-anchorable" locations.

mojavelinux:
I like this idea, but I don't think this is something that should be done in the converter. The converter shouldn't be generating or assigning IDs. If this was to be part of core, it would need to be done in the parser. I think there's also a distinction to be made between adding anchors to dt elements (dtanchors) and auto-generating IDs for them (dtids).

As a stepping stone, it's possible to do this today without modifying core by using an extension.

```rb
  tree_processor do
    process do |doc|
      next unless doc.attr? 'dtanchors'
      (doc.find_by context: :dlist).each do |dlist|
        dlist.items.each do |(terms, _)|
          Array(terms).each do |term|
            term_id = Asciidoctor::Section.generate_id term.text, doc
            term.text = %(pass:[<a id="#{term_id}" class="anchor" href="##{term_id}"></a>]#{term.instance_variable_get :@text})
          end
        end
      end
      nil
    end
  end
end
```

The only thing I don't like about this extension is that we have to use a pass macro to insert the anchor. I though about using the anchor macro (`anchor:<id>[role=anchor,link=self]`), but the anchor macro doesn't currently support attributes. Perhaps it should.

I'm going to promote this to an issue so we can discuss how this could become part of core.

mojavelinux:
I have opened #2717 to discuss. It turns out, there are a lot of decisions to make before we could make this a part of core. Please feel free to join the discussion. If you're happy with just using the extension, that's fine to. I'll reference this PR as a proof of concept, but I won't likely move forward with this implementation.

buckelij:
@mojavelinux thanks for picking this up ✨

I’ve subscribed to the issue. Feel free to close this PR.

mojavelinux:
Thank you for subscribing. We will continue this discussion there. Thank you again for proposing this feature and providing a prototype.

Issue #858, opened 2014-01-07
Add possibility to use figure number or text in link text


megathaum:
I would like to write
"The overall architecture of the system is shown in Figure 1."
instead of 
"The overall architecture of the system is shown in "System overview". 

In printed documents (PDF) it is much more convenient for the reader to look up the figure by number than by text. Moreover, as the caption describes the figure and that information is already provided in the text that refers to the figure it makes no sense to repeat it.

jmini:
From a LaTeX point of view, what asciidoctor supports now is `\refname{system_overview}` and the missing feature is `\ref{system_overview}`.

[Related discussion](http://discuss.asciidoctor.org/How-to-reference-a-figure-with-its-number-td1260.html).

jxxcarlson:
Could you clarify what you mean?  Thanks!

jxxcarlson:
I think I understand now — I hadn’t looked at the discussion.

jmini:
In the project I am working for, we would like to have a single way of writing documentation. For the moment we have plain HTML pages, MediaWiki pages and documents produced with LaTeX. This is a great oportunity for me to check missing (or nice-to-have) features in Asciidoctor.

In our current LaTeX document, we use most of the time `\ref` and sometimes `\refname`.

To obtain:
> The overall architecture of the system is shown in Figure 1.

The LaTeX code is:

    The overall architecture of the system is shown in Figure~\ref{system_overview}

This issue is exactly about adding this missing feature. With Acsiidoctor:

    The overall architecture of the system is show in <<system_overview>>.

Produces: 
> "The overall architecture of the system is shown in System overview

I miss something like `<<system_overview,Figure ?>>` where the `?` should be replaced at generation time by the correct figure number in this document. (This is only a syntax proposition. As you understood it, I am not an Asciidoctor expert).

jxxcarlson:
Very good points — I later understood what you meant from the discussion, and I agree with it.
I need it myself!

A couple of things that you might be interested in related to Asciidoctor and LaTeX.

(1) We are working on Asciidoctor-latex, an extension which (a) adds latex-lke
functonality to Asciidoctor, and (b) which will allow one to convert an Asciidoctor
document to latex.  Part (a) is already being used at http://noteshare.io <http://noteshare.io/> — more on
this below.  Part (b) has  working prototype, but this needs more work.  I am also
working on a latex to asciidoctor converter, but this is much less advanced.

[...]

PR #1921, opened 2016-11-11
Adds option to include the section number in a cross-reference

fap-:
This is my implementation for the proposal I made in [asciidoctor-pdf-464](https://github.com/asciidoctor/asciidoctor-pdf/issues/464#issuecomment-259584614).
I broke one test but I currently don't understand how and why.
Feedback would be welcome.

### DIFF
--- a/lib/asciidoctor/abstract_block.rb
+++ b/lib/asciidoctor/abstract_block.rb
@@ -404,7 +404,7 @@ class AbstractBlock < AbstractNode
   #
   # IMPORTANT You must invoke this method on a node after removing
   # child sections or else the internal counters will be off.
-  # 
+  #
   # Returns nothing
   def reindex_sections
     @next_section_index = 0
@@ -415,6 +415,7 @@ class AbstractBlock < AbstractNode
         block.reindex_sections
       end
     }
+    sections.each(&:set_references)
   end
 end
 end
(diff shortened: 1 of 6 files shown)


fap-:
Okay, so the test breaks because the references get set in a different order after my changes and Document#register doesn't overwrite :ids by default.
I'm not sure what's the best way to handle this. Registering the references in Section#set_reference with :force would fix the test but wouldn't guarantee that the first appearance of a reference id would be the one in the final document; which I assume is the desired behavior?

mojavelinux:
I still think we can simplify this a lot. I believe the right solution is to store the node instead of the text in the references table (at the time the ID is being stored). Then, we just need a helper method like reftext on the node that does something similar to what the title method does. It puts everything together during conversion to output the reference text in the specified style. (There's still a lot of rework we need in the model to handle sections properly, but it's doable).

We want to try to keep as much of this out of the view of the user as possible. That's why I don't think we should be messing with section-label (which should be static) and reftext (which is really for modifying the text part of the reference, not the whole reference string). Where this all belongs is in the converter, with helper methods on node. The parser shouldn't really even know about any of this, IMO.

mojavelinux:
I've added additional comments to #858 about how I think this should be implemented. I'm working on a prototype, which hopefully I'll push later this week.

fap-:
Closing this PR then. Thank you.

mojavelinux:
I'll still use this PR as a reference, and probably take some of the tests. I just need to integrate all these ideas together. Then I'll have you review the new proposal and we can collaborate on that to make sure we hit all your needs.

PR #1925, opened 2016-11-12
Paths with file URI scheme are inevitably absolute

Mogztter:
* `file://c:/foo/bar` is absolute (Windows path)
* `file:///foo/bar` is absolute (Linux path)
* `file://host/foo/bar` is absolute (URI path)

Currently `is_root?` method will return false when the path is `file://c:/foo/bar` (absolute path on Windows)


mojavelinux:
Seems reasonable to me!

PR #1952, opened 2016-11-28
Include entire test suite in gem

voxik:
Some of test files were missing (test_helper.rb and fixtures folder). This PR included them into the gem, although it is not as careful about included files as the other RegExps ...

Other option would be to drop the test suite from the gem entirely ...

### DIFF
--- a/asciidoctor.gemspec
+++ b/asciidoctor.gemspec
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
   end
   s.files = files.grep(/^(?:(?:data|lib|man)\/.+|Gemfile|Rakefile|(?:CHANGELOG|CONTRIBUTING|LICENSE|README(?:-\w+)?)\.adoc|#{s.name}\.gemspec)$/)
   s.executables = files.grep(/^bin\//).map {|f| File.basename f }
-  s.test_files = files.grep(/^(?:test\/.*_test\.rb|features\/.*\.(?:feature|rb))$/)
   s.require_paths = ['lib']
+  s.test_files = files.grep(/^(?:(?:features|test)\/.+)$/)
   s.has_rdoc = true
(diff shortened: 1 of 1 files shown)


mojavelinux:
Oops! We should be including everything under test.

The reason I include the test suite is specifically for the rpm packaging. Is there a way to run the test suite when packaging the gem if the files are not included in the gem?

mojavelinux:
I see absolutely no indication than anything in the Ruby ecosystem makes use of the test_files field in the specification. Therefore, I'm inclined to simply put all the files in the files array.

mojavelinux:
(they end up there anyway).

mojavelinux:
For now I'm just going to include all the files needed to run the tests in test_files. If you can confirm there is a way to run the tests when packaging the RPM without including the test_files in the gem, then I will just drop the line.

mojavelinux:
Useful discussion on the topic of the test_files field: http://stackoverflow.com/questions/18871541/what-is-the-purpose-of-test-files-configuration-in-a-gemspec

voxik:
I appreciate you attitude toward RPM. Yes, it is definitely better for that case to have the entire test suite in the package and that is actually where I am coming from ;)

On the other hand, there is plenty of gems which decided to remove the test suite and I can completely understand that decision. Luckily it is just a bit more work to execute the test suite, since you have to prepare tarball with the tests. Not a big deal.

Thx for the analysis of the situation and merging!

mojavelinux:
I appreciate RPM (and DEB) very much and want to make it easy as possible for packagers to do the critical job that they do. I especially appreciate your work in making Ruby so awesome in Fedora.

I'm happy to continue to include the test suite in the gem. Now, all the files should be there. I'll be publishing the 1.5.6 release before the end of the year.

PR #1953, opened 2016-11-28
Upgrade to Slim 3

voxik:
Why test against old Slim if there is newer version ...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,11 @@ source 'https://rubygems.org'
 # Look in asciidoctor.gemspec for runtime and development dependencies
 gemspec
 
+group :development do
+  gem 'racc', '~> 1.4.10' if RUBY_VERSION == '2.1.0' && RUBY_ENGINE == 'rbx'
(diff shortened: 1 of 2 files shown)


voxik:
Hmmm, Ruby 1.8.7 ....

mojavelinux:
I can switch to a newer Slim in the gemspec and override it in the Gemfile for the Ruby 1.8.7 tests. (Yes, I realize 1.8.7 needs to be dropped, but I want to do that at the transition to 1.6.0, which I'm trying hard to get to).

voxik:
I was not lobbying for removing Ruby 1.8.7 ... I just not expected the failure :)

mojavelinux:
Oh, don't worry, we've already decided to drop Ruby 1.8.7...it's just a formality at this point.

I still want to test against a recent Slim. That may mean loosing the dependency, upgrading it in Gemfile, or something else. I'll play around with it.

mojavelinux:
I found a way to do it. Thanks!

Issue #1131, opened 2014-10-08
Scrub strings to eliminate encoding errors


mojavelinux:
Scrub strings (or similar technique) when sanitizing them in the reader to eliminate encoding errors. If there are any characters which had to be dropped, report the line information to the user.

See http://globaldev.co.uk/2014/05/ruby-2-1-in-detail/#stringscrub

It's not enough just to force the encoding to UTF-8 since errors can still happen. We want to help the user understand where to fix the problem in the source file.

mojavelinux:
We might also consider adding the `'r:UTF-8'` flag to all IO read operations instead of using force_encoding after loading the file. The only downside of that approach is that we lose support for files with a UTF-16 BOM. I suppose we could detect that and use the `'r:UTF-16'` flag in that case.

mojavelinux:
I think we're going to need to allow the input encoding to be specified via the API or CLI. Here's how the logic works.

* By default, we read the file as UTF-8.
  * If a BOM is detected, we force the appropriate encoding, then encode as UTF-8.
* If an input encoding is specified, we read the file as that input encoding, then encode as UTF-8.

We always have to end up with UTF-8, but we can provide more support for hints to get there. Right now, we're only honoring a BOM in the file. We could extend this to honor a request to encode to UTF-8 from a specified encoding. We would do this on faith.

mojavelinux:
We should not honor the input encoding defined in Ruby because that makes Asciidoctor non-portable.

mojavelinux:
When reading a file, we either honor the BOM or assume UTF-8.

When accepting a string (or string array), we encode to UTF-8, allowing Ruby to handle the cross encoding. This is where string scrubbing comes into play. We might want to set the fallback character.

mojavelinux:
If we do decide to do this, we definitely need to warn. I think it's better to fail when there are invalid character in a file than to silently replace them because then the user has no idea there's an issue. A warning is a good compromise.

PR #1969, opened 2016-12-21
Encode filename by UTF-8 when convert_file

zhuermao:
Encode filename by UTF-8 when convert_file, so that source document of which path/file name contains CJK characters would be available.

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -1582,7 +1582,7 @@ module Asciidoctor
   # Returns the Document object if the converted String is written to a
   # file, otherwise the converted String
   def convert_file filename, options = {}
-    ::File.open(filename) {|file| self.convert file, options }
+    ::File.open(filename.encode("utf-8")) {|file| self.convert file, options }
   end
 
   # Alias render_file to convert_file to maintain backwards compatibility


mojavelinux:
We're planning to make this change when we move to 1.6.0, which the version in which we'll officially drop support for Ruby 1.8.7. Lack of support for this flag in Ruby 1.8.7 is the reason for the test failure.

Until then, you must set your system (or Ruby) to use UTF-8 for input and output. See https://github.com/asciidoctor/asciidoctor#requirements.

Can you show me the output of the command `asciidoctor -v`?

mojavelinux:
I plan to close this pull request, but rest assured I have added this to my list of things to do for the 1.6.0 release. See https://github.com/asciidoctor/asciidoctor/issues/1131#issuecomment-283910569.

mojavelinux:
Even with all the new encoding features in Ruby 2, I can't managed to demonstrate how this solves a problem. As of Asciidoctor 2.0.0, we now handle encoding at the interface to the filesystem by specifying the internal and external encoding flags when reading and writing a file. (And we may still make the source encoding configurable).

I don't want to blindly add encode calls in Asciidoctor without understanding why they are needed. Unless you can demonstrate in a reproducible / CI environment that this actually solves a problem, I'd rather not make this change.

Note that Asciidoctor (as with any Ruby application) relies on the Encoding.default_internal to be UTF-8 (which you can set using `-E :utf-8`). Otherwise, encodings get mixed up internally as Ruby concatenates strings. That's really out of the application's control.

PR #1978, opened 2016-12-30
preserve inline break in manpages

letheed:
* escape the roff line break command
* unescape escaped roff commands after the leading period substitution pass


mojavelinux:
Nice work! I know that code isn't easy to understand. I'm impressed!

PR #2015, opened 2017-02-04
Run pygments with startinline = true

ricpelo:
Works better for PHP language.

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -1469,7 +1469,7 @@ module Substitutors
           :bold_every => false}].highlight source
     when 'pygments'
       lexer = ::Pygments::Lexer[attr('language', nil, false)] || ::Pygments::Lexer['text']
-      opts = { :cssclass => 'pyhl', :classprefix => 'tok-', :nobackground => true }
+      opts = { :cssclass => 'pyhl', :classprefix => 'tok-', :nobackground => true, :startinline => true }
       unless (@document.attributes['pygments-css'] || 'class') == 'class'
         opts[:noclasses] = true
         opts[:style] = (@document.attributes['pygments-style'] || Stylesheets::DEFAULT_PYGMENTS_STYLE)


ramsey:
Does this always turn on `startinline`? If so, that may not be desirable in all cases.

mojavelinux:
What I would prefer is to either:

* break php into two langauges, php and phpinline
* add an option to the block for inline mode, opts=inline

Then we can do the right thing.

I'm leaning more towards the second. wdyt?

mojavelinux:
I've seen other discussions about this setting, such as in Jekyll, and most people agree it shouldn't be the default.

ramsey:
I agree that an option to the block for inline mode is the right approach.

ricpelo:
I agree: the best is to have an option for the block.

mojavelinux:
@ricpelo Would you like to update the PR? Check for the language and if it is php, then check for the option. Something like:

```ruby
opts[:startinline] = true if lexer.name == 'PHP' && (option? 'startinline')
```

mojavelinux:
Given that this behavior is unique to Pygments, and all other source highlighters just highlight PHP normally, I'm incline to just go ahead and set this. If I decide to make at an option before the release, I'll just update the code and document it here.

mojavelinux:
As it turns out, this will break "mixed mode" PHP. Therefore, I propose that if the option "mixed" is set, then we disable this option. Otherwise, we enable it by default since most of the time, a fragment of pure PHP is being highlighted.

See the related issue in Asciidoctor PDF, where I am exploring these ideas. https://github.com/asciidoctor/asciidoctor-pdf/issues/755

mojavelinux:
@ricpelo I'd like to get your thoughts on the related issue. I think we're close to formalizing the option.

ramsey:
Won't this break everyone's documents who are currently using PHP and having to use the PHP open tag to make highlighting work?

ramsey:
Since mixed mode is the default on Pygments, I would use it as the default here, since that's the expected behavior.

ricpelo:
@mojavelinux I think you're right: `startinline` option enabled by default works great with any PHP snippet, with or without the `<?php` tag. About mixing HTML and PHP code, remember that Pygments have the `html+php` lexer specifically for that. See http://pygments.org/demo/6343473/ for example.

mojavelinux:
@ramsey From my research, all the integrations are switching to enabling startinline by default, including code highlighting on GitHub. So I'd argue if we don't follow that, we will look broken. But what we offer that others don't is the ability to switch it back off by using `source%mixed`.

@ricpelo Yes, another option is to simply set the language to "html+php". That also works. So we don't break that option at all. And that convinces me even more that using startinline by default for "php" is the right choice. You can use "source,php" for pure PHP. You can use either "source%mixed,php" or "source,html+php" for mixed mode (or a full PHP document that's enclosed in PHP tags).

mojavelinux:
@ramsey To answer your question even more directly, the only thing that breaks if you have a full PHP document enclosed in PHP tags without setting the `mixed` option (or html+php language) is the highlighting of the PHP start tag itself. It does not actually break the highlighting of the code.

ramsey:
Sounds great, then. Looking forward to seeing this. Thanks!

PR #2026, opened 2017-02-20
Avoid boundary pseudo-tags in manpage table output

bk2204:
When using inline styling in a manpage table cell, the resulting manpage would result in `<BOUNDARY>` tags in the output. Call the manify function to fix up the text and remove these tags.


mojavelinux:
Thanks @bk2204!

PR #2038, opened 2017-02-23
promote subs to top-level constants

mojavelinux:
* promote subs to top-level constants
* reuse constants instead of using hard-coded subs arrays
* make lock_in_subs more efficient
* change lock_in_subs to return resolved subs
* minor style improvements


mojavelinux:
Merged as b6de807c4d5d7746b63121292a8f8d6e1a474b10.

PR #2063, opened 2017-03-09
pass symbol ref to map when calling method on iteration variable

mojavelinux:


mojavelinux:
I did some quick tests and the `.map(&:method)` syntax is twice as slow on Opal. I'm going to defer that part of the patch and just pull in the enclose + join optimizations.

mojavelinux:
Merged as 94bfcc41dda060026532eb25774cf8d0d7fd13dc (minus the method refs).

Issue #2068, opened 2017-03-11
Enhance lineno API


ldez:
Store the line number for the start of the description list block.

And if possible for list item also.

related to https://github.com/asciidoctor/atom-asciidoc-preview/pull/216

mojavelinux:
This is resolved in 1.5.7. As a bonus, you get table cells too.

#2257 was filed as duplicate of this issue, which is where the development notes can be found.

PR #2069, opened 2017-03-11
Adds sourcemap on list items

Mogztter:
Quick solution to add sourcemap on list items.

At some point, I think we should refactor this code:

```ruby
# QUESTION should we introduce a parsing context object?
```

If we had a parsing context, we could know that `sourcemap` is enabled and use this information to set the `source_location` when we instantiate a new block. 

### TODO
* [x] line number is invalid on multi-line term/definition or multi-line questions and answers

Ref: https://github.com/asciidoctor/asciidoctor/issues/2068

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -1277,7 +1277,9 @@ class Parser
   def self.next_list_item(reader, list_block, match, sibling_trait = nil)
     if (list_type = list_block.context) == :dlist
       list_term = ListItem.new(list_block, match[1])
+      list_term.source_location = reader.cursor if list_block.document.sourcemap
       list_item = ListItem.new(list_block, match[3])
+      list_item.source_location = reader.cursor if list_block.document.sourcemap
       has_text = !match[3].nil_or_empty?
     else
       # Create list item using first line as the text of the list item
@@ -1295,6 +1297,7 @@ class Parser
         end
       end
       list_item = ListItem.new(list_block, text)
+      list_item.source_location = reader.cursor if list_block.document.sourcemap
 
       if checkbox
         # FIXME checklist never makes it into the options attribute
@@ -1310,7 +1313,9 @@ class Parser
 
     # first skip the line with the marker / term
     reader.advance
-    list_item_reader = Reader.new read_lines_for_list_item(reader, list_type, sibling_trait, has_text), reader.cursor
+    cursor = reader.cursor
+
+    list_item_reader = Reader.new read_lines_for_list_item(reader, list_type, sibling_trait, has_text), cursor
     if list_item_reader.has_more_lines?
       # NOTE peek on the other side of any comment lines
       comment_lines = list_item_reader.skip_line_comments
@@ -1337,6 +1342,9 @@ class Parser
       # list
       while list_item_reader.has_more_lines?
         if (new_block = next_block(list_item_reader, list_item, {}, options))
+          if list_type == :dlist
+            list_item.source_location = cursor if list_block.document.sourcemap
+          end
           list_item << new_block
         end
       end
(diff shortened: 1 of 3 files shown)


mojavelinux:
> At some point, I think we should refactor this code:

Trust me, I wholeheartedly agree. The list parsing code is the roughest in all of core. In fact, it's actually parsing every list twice. But at the time, it was a choice over getting it working or getting it right. We definitely need to revisit.

mojavelinux:
Due to how list parsing is handled (the double pass), the line numbers for nested lists are way off. I'd like to address that issue before accepting this change to avoid support overhead. If the line numbers are wrong, people are going to report that via issues.

We need to rewrite the list parsing so that it works against the original reader using a recursive descent strategy. Fortunately, we have plenty of tests to verify that we got it right.

As a more short-term solution, we could figure out how to count the line offsets currently. The information will still be incorrect if the list spans several include files, but at least we'll have the basic case working.

lib/asciidoctor/parser.rb
```diff
@@ -1337,6 +1341,9 @@ def self.next_list_item(reader, list_block, match, sibling_trait = nil)
       # list
       while list_item_reader.has_more_lines?
         if (new_block = next_block(list_item_reader, list_item, {}, options))
+          if list_type == :dlist
+            list_item.source_location = list_item_reader.cursor if list_block.document.sourcemap
```
Mogztter: This is not working anymore.
Given:

```
1 * level 1
2 ** level 2
3 *** level 3
4 **** level 4
5 ***** level 5
6 * level 1
```

When we are processing the first item of the nested list (ie. level 2), the cursor is already line 6.
Do we need to increment a counter in the while loop to keep track of the line number ?

Mogztter:
> Due to how list parsing is handled (the double pass), the line numbers for nested lists are way off. I'd like to address that issue before accepting this change to avoid support overhead. If the line numbers are wrong, people are going to report that via issues.
> We need to rewrite the list parsing so that it works against the original reader using a recursive descent strategy. Fortunately, we have plenty of tests to verify that we got it right.

I'm starting to understand your comment :thinking: 
I need to spend some time on the parser code to understand what needs to be done :see_no_evil:

lib/asciidoctor/parser.rb
```diff
@@ -1310,7 +1313,9 @@ def self.next_list_item(reader, list_block, match, sibling_trait = nil)
 
     # first skip the line with the marker / term
     reader.advance
-    list_item_reader = Reader.new read_lines_for_list_item(reader, list_type, sibling_trait, has_text), reader.cursor
+    cursor = reader.cursor
+
+    list_item_reader = Reader.new read_lines_for_list_item(reader, list_type, sibling_trait, has_text), cursor
```
Mogztter: @mojavelinux Here's the trick, we are "saving" the cursor, otherwise the lineno is wrong on nested lists.

lib/asciidoctor/parser.rb
```diff
@@ -1337,6 +1342,9 @@ def self.next_list_item(reader, list_block, match, sibling_trait = nil)
       # list
       while list_item_reader.has_more_lines?
         if (new_block = next_block(list_item_reader, list_item, {}, options))
+          if list_type == :dlist
```
Mogztter: Arguably ugly :wink:

Mogztter:
I'm fully aware that this is a "naive" implementation but the parser code is really dense and I'm not ready (*yet*) to dig into it :stuck_out_tongue: 

The good news is that tests are green and I'm only adding 8 lines so I don't think it will make the refactoring harder when we are ready to improve the list parsing.

@mojavelinux Let me know what you think

mojavelinux:
The changes in this PR have been incorporated in #2701. You had the code exactly right. I just made a few minor style changes and optimizations. This is an excellent addition! Thanks!

PR #2070, opened 2017-03-13
move regular expression constants for Opal to Asciidoctor.js build

mojavelinux:


lib/asciidoctor.rb
```diff
@@ -7,10 +7,9 @@
 
 require 'set'
 
-# NOTE RUBY_ENGINE == 'opal' conditional blocks are filtered by the Opal preprocessor
+# NOTE RUBY_ENGINE == 'opal' conditional blocks like this are filtered by the Opal preprocessor
 if RUBY_ENGINE == 'opal'
-  # NOTE asciidoctor/opal_ext is supplied by the Asciidoctor.js build
-  require 'asciidoctor/opal_ext'
+  require 'asciidoctor/opal' # supplies compile and runtime overrides for the Opal environment; fulfilled by Asciidoctor.js
```
Mogztter: > It's more technically correct to put the require at the top of the file. I plan to add it right under the require for opal_ext. 

I've done some tests recently and with the require at the top, I couldn't override `path_resolver` methods.

> Then, you can decide on your end what you want that to resolve to. This is a clear API between the two compilers, I think. Would you agree?

`require 'asciidoctor/opal'` is fine :smile:

mojavelinux:
@Mogztter Let's try to coordinate this merge. I don't want to break any existing builds. Is there a need to compile an old branch of Asciidoctor.js against Asciidoctor master? Or can we assume that Asciidoctor.js will just update to conform to the latest state of Asciidoctor? I'm just trying to think of scenarios that this change might break.

mojavelinux:
Btw, I think we need to start compiling Asciidoctor.js against Asciidoctor Ruby master like we do with AsciidoctorJ. That way, we don't have surprises when we go to release a new version of Asciidoctor.js.

lib/asciidoctor.rb
mojavelinux: That's a good point. Then I think we'll need two includes. Perhaps at the beginning we have:

```
require 'asciidoctor/opal'
```

That's where you put all the stuff that is supplemental. Then, at the very end of the file we can have:

```
require 'asciidoctor/opal/coda'
```

That would be a "coda" where you can add overrides.

wdyt? Should we try it?

mojavelinux: We could also name the first one:

```
require 'asciidoctor/opal/init'
```

Let me know your thoughts.

mojavelinux: I see you are doing some overrides in various places, so I think we definitely need that "coda" require.

mojavelinux: I tested it out and it seems to work great.

mojavelinux: I'm open to other names as well. Ideas I had include coda, postscript, post, overrides, patch, annex, omega. I kind of like "postscript" and "omega".

Mogztter: Basically the goal of the first `require` is to provide/override Ruby method (provided by Opal) and the goal of the second `require` is to provide/override Asciidoctor method.

Maybe we could create a module named `js` ?

```ruby
require 'asciidoctor/js/opal'
```

```ruby
require 'asciidoctor/js/asciidoctor'
```
(I don't like the repetition of `asciidoctor`)

> I'm open to other names as well. Ideas I had include coda, postscript, post, overrides, patch, annex, omega. I kind of like "postscript" and "omega".

I like `omega` but at the same time I think it's the least understandable

mojavelinux: I'm happy with a module named "js". It would essentially replace "opal". So the includes would be:

```ruby
require 'asciidoctor/js' # at top of asciidoctor.rb
```

and

```ruby
require 'asciidoctor/js/postscript' # at bottom of asciidoctor.rb
```

Here's what `asciidoctor/js` would look like:

```ruby
require 'asciidoctor/js/opal_ext'
require 'asciidoctor/js/runtime'
require 'strscan'
require 'asciidoctor/js/rx'
```

Here's what `asciidoctor/js/postscript` would look like:

```ruby
require 'asciidoctor/converter/composite'
require 'asciidoctor/converter/html5'
require 'asciidoctor/extensions'
```

With this layout, the regular expression constants will be defined before anything else in the Asciidoctor module is defined. But I think this is completely reasonable.

mojavelinux: Btw, if you provide extra Asciidoctor stuff, you should use asciidoctor_ext as the folder name.

```
lib/asciidoctor/js/asciidoctor_ext/
```

The avoids the confusion of a duplication of the folder name asciidoctor. I do this, for example, in Asciidoctor PDF.

mojavelinux:
I'll update the PR so you can see my proposal in action.

Mogztter:
Ok, sounds good to me 👍

mojavelinux:
Updated.

Mogztter:
> Let's try to coordinate this merge. I don't want to break any existing builds. Is there a need to compile an old branch of Asciidoctor.js against Asciidoctor master? Or can we assume that Asciidoctor.js will just update to conform to the latest state of Asciidoctor? I'm just trying to think of scenarios that this change might break.

I think you can merge this pull request, I will conform to the latest version of Asciidoctor: https://github.com/asciidoctor/asciidoctor.js/pull/290

mojavelinux:
Let's do this!

Issue #2071, opened 2017-03-13
Add rel="noopener noreferrer" to target="_blank".


EdOverflow:
`rel="noopener noreferrer"` should be added to links containing `target="_blank"`  as a precaution against reverse tabnabbing. For more information, please refer to the following article:
https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

AsciiDoc: `http://example.com[Link^]`
HTML: `<a href="http://example.com" target="_blank">Link</a>`
Fix: `<a href="http://example.com" target="_blank" rel="noopener noreferrer">Link</a>`

~~~rb
attrs << %( target="#{node.attr 'window'}") if node.attr? 'window', nil, false
%(<a href="#{target}"#{attrs.join}>#{node.text}</a>)
~~~

Link to source code: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L1033

mojavelinux:
Thank you for bringing this up for discussion. I now understand the vulnerability.

@hsablonniere what are your thoughts about this? Do you think we should add this if target is _blank? FWIW, Jake seems to like it. See https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/.

It seems like Firefox has fixed this problem in Firefox 52, so perhaps we don't need the exception for Firefox anymore. We can just go with rel="noopener".

mojavelinux:
Btw, I found this site by mathias to be very helpful in understand the issue. https://mathiasbynens.github.io/rel-noopener/

hsablonniere:
Hey, I read mathias and Jake's article when they got out. This is very important. I would say 100% agree for `noopener`. I'm a bit less enthusiastic about `noreferrer` because of analytics but it seems to be the solution for old browsers right?

hsablonniere:
Then I would say, let's to this and put an attribute to opt-out.

DeepakSingla1:
Removing the rel=”noreferrer noopener” tag will expose you to the risks carried by [target="_blank"](https://dev.to/ben/the-targetblank-vulnerability-by-example) tag. Must add the rel=”noreferrer noopener” tag to avoid falling into trouble. Read here to get the insight into the issue: https://blog.templatetoaster.com/rel-noreferrer-noopener-seo/

mojavelinux:
[...]

PR #2072, opened 2017-03-13
Add rel="noopener noreferrer" to target="_blank".

EdOverflow:
Solves https://github.com/asciidoctor/asciidoctor/issues/2071.


mojavelinux:
This isn't necessary for all links. I've applied more specialized logic in a PR that supercedes this one. See #2168.

PR #2094, opened 2017-03-21
Fix a typo base64 data

Mogztter:


mojavelinux:
Thanks for catching this! It looks like this line is 3 years old, which means we are missing a test case. Could you insert one here https://github.com/asciidoctor/asciidoctor/blob/master/test/blocks_test.rb#L1811 ?

Mogztter:
> Thanks for catching this! It looks like this line is 3 years old, which means we are missing a test case.

An Asciidoctor.js test case caught it :wink: 

> Could you insert one here

Sure

Mogztter:
Done!

mojavelinux:
Thanks @Mogztter! Now we're covered in both places!

PR #2107, opened 2017-04-02
path_resolver.relative_path: use standard library that just works

ztmr:
`path_resolver.relative_path` does not work if the path of interest is not in the subtree of the reference directory.

Original function naively cut longest common path prefix.

NOTE: Can we afford requiring pathname library? Does it work properly with OPAL?

WARNING: unable to run tests due to dependency errors so cannot confirm the test is correct, please review/rerun.

### DIFF
--- a/lib/asciidoctor/path_resolver.rb
+++ b/lib/asciidoctor/path_resolver.rb
@@ -470,12 +470,10 @@ class PathResolver
   #
   # Return the relative path String of the filename calculated from the base directory
   def relative_path filename, base_directory
-    if (is_root? filename) && (is_root? base_directory)
-      offset = (base_directory.end_with? @file_separator) ? base_directory.length : base_directory.length + 1
-      filename[offset..-1]
-    else
-      filename
-    end
+    Helpers.require_library 'pathname'
+    f = Pathname.new filename
+    d = Pathname.new base_directory
+    (f.relative_path_from d).to_s
   end
 end
 end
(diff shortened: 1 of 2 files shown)


mojavelinux:
> path_resolver.relative_path does not work if the path of interest is not in the subtree of the reference directory.

If I remember correctly, that is actually by design. What I can't remember is whether that was a compliance issue with AsciiDoc Python or there was another reason for placing that restriction. I need to revisit my notes and document what's going on here.

> Original function naively cut longest common path prefix.

Again, I don't think it's naive. I think it's intentional...either for reasons of security or reasons of best practice. I need to refresh my memory.

I would prefer not to use the Pathname library. In my experience, it's very slow and not well written. There have been numerous blogs I've come across challenging it's quality. That should help explain its intentional absence here.

mojavelinux:
In other words, PathResolver is meant to be a more robust replacement for Pathname.

mojavelinux:
I did some research and I agree this is a bug.

mojavelinux:
I fixed this by simply checking to ensure the filename shares a common root with the base directory before truncating.

The main reason we don't use Pathname is because we don't need all that logic. Part of the contract of the method is that both paths are already expanded. So all we have to do is take away the base directory part if they share the same root. This is very much an internal method.

mojavelinux:
See https://github.com/asciidoctor/asciidoctor/pull/2335

mojavelinux:
I've improved this logic even further in 1.5.7.

PR #2111, opened 2017-04-04
resolves #2110 remove regexp hacks for Opal

### LINKED ISSUE #2110: Remove regexp hacks for Opal in substitutors
Since Firefox 34, Firefox no longer returns empty string for non-matching groups passed to the String#replace function. I think it's safe to remove these hacks from core as all modern JavaScript implementations do the right thing.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Gecko-specific_notes

mojavelinux:


mojavelinux:
Let's do it!

mojavelinux:
Except, it just :snowflake: in Denver last night :cry:

Mogztter:
WUT?!
It's 20 degrees Celsius in France :sun_with_face:

mojavelinux:
I know. They should call it Crazy Colorado, because that is what the weather is like here sometimes.

Don't worry, next day was 20C and all the snow was gone.

mojavelinux:
I created a polyfill that makes Firefox < 34 behave like modern browsers. I'd be open to adding this to Asciidoctor.js. This addresses the problem from the JavaScript side instead of from the Ruby side.

https://gist.github.com/mojavelinux/2cebc5fd0d2139715f7723f31ba5bfdd

PR #2118, opened 2017-04-13
add stripes attribute for tables

### LINKED ISSUE #1365: Allow table striping to be controlled by role or attribute
Allow the striping of table rows (alternating colors) to be controlled by either a role or attribute.

To be consistent with how the borders are handled, the attribute might make the most sense.

```
[stripes="none"]
|===
|first row
|second row
|third row
|===
```

Though, we could disable it by default and allow it to be enabled using a role:

```
[.striped]
|===
|first row
|second row
|third row
|===
```

This would be similar to how Bootstrap works. See http://getbootstrap.com/css/#tables-striped.

As a side note, hovered rows might be interesting too. This may tip the scale towards using roles.

```
[striped.spotlight]
|===
...
|===
```

miltador:
It is already present in `asciidoctor-pdf`. I repeated the logic to match that.
Changed default stripes background color to match `asciidoctor` default styles.
Also added possibility to define own color through `stripesbgcolor` document attribute.

See https://github.com/asciidoctor/asciidoctor-pdf/issues/724 for reference.

Fixes #1365

### DIFF
--- a/data/stylesheets/asciidoctor-default.css
+++ b/data/stylesheets/asciidoctor-default.css
@@ -105,7 +105,6 @@ table{background:#fff;margin-bottom:1.25em;border:solid 1px #dedede}
 table thead,table tfoot{background:#f7f8f7;font-weight:bold}
 table thead tr th,table thead tr td,table tfoot tr th,table tfoot tr td{padding:.5em .625em .625em;font-size:inherit;color:rgba(0,0,0,.8);text-align:left}
 table tr th,table tr td{padding:.5625em .625em;font-size:inherit;color:rgba(0,0,0,.8)}
-table tr.even,table tr.alt,table tr:nth-of-type(even){background:#f8f8f7}
 table thead tr th,table tfoot tr th,table tbody tr td,table tr td,table tfoot tr td{display:table-cell;line-height:1.6}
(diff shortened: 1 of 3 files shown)


miltador:
Added handling of rowspans too so rows will be colored with desired order starting from after header.

mojavelinux:
I like this idea, but I want to do it in such a way that does not introduce a new attribute into the AsciiDoc document. Defining colors in the AsciiDoc document doesn't maintain a proper separation of content and presentation. I do think it's reasonable to be able to declare from the AsciiDoc document whether or not stripes are used for a given table, since that often depends on what information the table contains. But the colors should be something that is configuring using a CSS class.

What I'm looking for is an implementation that toggles the use of a CSS class on the table based on the value of the stripes attribute (even, odd, all, none).

miltador:
Thanks for the feedback and sharing thoughts! I appreciate your aim.
Would you mind if I open an issue for table stripes?

mojavelinux:
I apologize for taking so long to reply to your PR. Work on Asciidoctor 1.5.6 got a bit overwhelming and I had to draw the line and wrap it up so we could move forward.

I don't think there's a need to open another issues as I believe #1365 covers it. We just need to start a new PR. Let me know if you have any questions, or feel free to comment on that issue. Thank you for your participation!

PR #2120, opened 2017-04-13
resolves #2119 add handles? method to DSL for IncludeProcessor

### LINKED ISSUE #2119: Add handles? method to DSL for IncludeProcessor
handles? should be a DSL method of the IncludeProcessor. In other words, it should be possible to specify the handles? method for an IncludeProcessor using the DSL syntax similar to how the process method can be specified.

Example:

```ruby
  include_processor do
    process do |doc, reader, target, attrs|
      reader.push_include %(content of #{target}), target, target, 1, attrs
    end

    handles? do |target|
      target.end_with? '.txt'
    end
  end
end
```

(Currently, handles? can only be defined as a method using the def keyword).

The other forms are also legal (though all normalized by Ruby itself)

```ruby
handles? {|target| target.end_with? '.txt' }
```

```ruby
handles? &-> (target) { target.end_with? '.txt' }
```

```ruby
handles? &proc {|target| target.end_with? '.txt' }
```

mojavelinux:


Mogztter:
Excellent ! 👍

mojavelinux:
I also plan to add a DSL option that lets you control whether attributes are parsed...and if so, how any positional attributes are mapped. I'll create a follow-up issue.

PR #2127, opened 2017-04-15
apply default_attrs to custom inline macro

mojavelinux:
- apply default_attrs to custom inline macro
- only apply default if key is missing from attributes
- update inline macro test to verify new behavior

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -578,9 +578,9 @@ class Parser
                 attributes['text'] = raw_attributes
               end
               if (default_attrs = extension.config[:default_attrs])
-                default_attrs.each {|k, v| attributes[k] ||= v }
+                attributes.update(default_attrs) {|k, old_v, _| old_v }
               end
-              if (block = extension.process_method[parent, target, attributes.dup])
+              if (block = extension.process_method[parent, target, attributes])
                 attributes.replace block.attributes
               else
                 attributes.clear
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -635,6 +635,9 @@ module Substitutors
               { 'text' => (unescape_bracketed_text m[2]) }
             end
           end
+          if (default_attrs = extension.config[:default_attrs])
+            attributes.update(default_attrs) {|k, old_v, _| old_v }
+          end
           extension.process_method[self, target, attributes]
         }
       end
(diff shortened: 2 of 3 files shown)


mojavelinux:
@Mogztter I'm thinking that it doesn't make sense to apply default attributes unless the content model is `:attributes` (in other words, when attributes are being parsed). Can you think of a case when default attributes are needed but the attributes aren't being parsed? The way I see it, they are just passed through anyway, so it's no different than defining the attributes in a constant or in a local variable. The whole purpose of a default attribute is to fall back to a value if one isn't specified by the user.

Mogztter:
> I'm thinking that it doesn't make sense to apply default attributes unless the content model is :attributes (in other words, when attributes are parsed). Can you think of a case when default attributes are needed but the attributes aren't being parsed? The way I see it, they are just passed through anyway, so it's no different than defining the attributes in a constant or in a local variable.

It's true, but it's easy to mess up.
If we do not apply default attributes when `content_model` is not `attributes`, I think we should print a warning message or throw an exception:

```ruby
# Good
class TemperatureMacro < Asciidoctor::Extensions::InlineMacroProcessor; use_dsl
  content_model :attributes
  seed_attributes_with 'precision' => '1'
  def process parent, target, attributes
    precision = attrs['precision'] # 1
  end
end

# Bad
class TemperatureMacro < Asciidoctor::Extensions::InlineMacroProcessor; use_dsl
  content_model :text
  seed_attributes_with 'precision' => '1'
  # Note, you should define a constant 
  # default_precision = '1'
  def process parent, target, attributes
    precision = attrs['precision'] # nil :(
  end
end
```

Is `:attributes` the default value of `content_model` ?
If we only define `seed_attributes_with 'precision' => '1'` will `content_model` be `:attributes` ?

mojavelinux:
Since there's a risk of confusion, and there's really no harm in setting default attributes even when attributes aren't parsed, I think we should just allow it. I can also think of the case where default attributes could be useful in an inheritance situation as a way to pass common values to subclasses.

> Is :attributes the default value of content_model ?

Yes. See https://github.com/asciidoctor/asciidoctor/blob/1e987096e6da24de912146c864be7bfa9a543af4/lib/asciidoctor/extensions.rb#L405

> If we only define seed_attributes_with 'precision' => '1' will content_model be :attributes ?

No. But that's how the new `resolve_attributes` will work. I'm also going to remove the `seed_attributes_with` method as part of that change.

mojavelinux:
Thanks for providing feedback!

Mogztter:
You're welcome.

> Since there's a risk of confusion, and there's really no harm in setting default attributes even when attributes aren't parsed, I think we should just allow it.

This sounds good to me.

PR #2129, opened 2017-04-17
resolves #1797 resolve attributes for short inline macros if requested

### LINKED ISSUE #1797: Getting positional attributes from a short inline macro
I'm processing this `stem:[x^2]` inline macro in AsciidocFX in short format. I'm getting the text with target parameter. However, I optionally want to provide users to declare image name and extension like `stem:[x^2,abc,png]`. But in this way, inside the [] becomes target, I can't get the parts as positional attributes. How can I do that?

Thanks.

mojavelinux:
- resolve attributes for short inline macros if content model is attributes
- don't attempt to parse an empty attribute string
- use text between square brackets as fallback value for target for backwards compat
- overlay parsed attributes on top of default attributes
- use abbreviated regexp delimiters
- add additional tests for various permutations of the inline macro

### DIFF
--- a/lib/asciidoctor/extensions.rb
+++ b/lib/asciidoctor/extensions.rb
@@ -468,9 +468,9 @@ module Extensions
     def resolve_regexp name, format
       # TODO memoize these regular expressions!
       if format == :short
-        %r(\\?#{name}:\[((?:\\\]|[^\]])*?)\])
+        /\\?#{name}:(){0}\[((?:\\\]|[^\]])*?)\]/
       else
-        %r(\\?#{name}:(\S+?)\[((?:\\\]|[^\]])*?)\])
+        /\\?#{name}:(\S+?)\[((?:\\\]|[^\]])*?)\]/
       end
     end
   end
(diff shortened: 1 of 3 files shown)


mojavelinux:
@Mogztter I'm a little unsure here what to do about backward compatibly. When we're parsing the short form using the `:attributes` content model, should we leave the target as nil?

The reason I'm unsure is because I've found many instances of extensions that are using the `:attributes` content model with the short form, but expecting the target to be the value between the square brackets. In other words:

```ruby
inline_macro do
  named :term
  using_format :short

  process do |parent, target, attrs|
    ...
  end
end
```

Example:

```
term:[application]
```

Technically, the target should be nil and the attributes `{ 1 => 'application' }`. However, if we do that we're going to break existing code.

What I'm proposing in this pull request is to set the target to the value between the brackets if the target is nil (which is the case for short inline macros). Perhaps we can document this and break it in 1.6.0. wdyt?

mojavelinux:
I'm ready to pull the trigger on this. I think I've massaged this code to death.

As a bonus, I added the ability to use named capture groups in a custom regexp. You can't use it yet in JavaScript, but it's kind of a preview of what's to come:

```ruby
match /@(?<target>\w+)/
```

PR #2139, opened 2017-04-21
resolves #2122 add resolves_attributes method to DSL for macros

### LINKED ISSUE #2122: Consolidate methods to configure attribute parsing in macro processors
Inspired by #2121, consolidate the methods on the macro processor extensions for configuring attribute parsing. Currently, there are three methods:

* content_model (aliased as parse_content_as)
* positional_attributes (aliased as name_positional_attributes, name_attributes, and pos_attrs)
* default_attrs (aliased as seed_attributes_with)

These methods feel clumsy. They can be reduced to a single method named `resolves_attribute` that can handle the following settings:

1. enable attribute parsing
2. map / bind positional attributes to attribute names
3. assign default attribute values

Here's an example that shows how this method can be used to disable attribute parsing:

```ruby
resolves_attributes false
```

(In this case, the content between the square brackets gets stored in the attribute named text).

Here's an example that shows how to enable attribute parsing:

```ruby
resolves_attributes
```

Here are examples that shows how to map positional attributes to attribute names:

```ruby
resolves_attributes 'size'
```

```ruby
resolves_attributes 'alt', 'width', 'height'
```

Here's an example that shows how to specify default attributes:

```ruby
resolves_attributes '@size' => 5, 'max-word-size' => 10
```

(The leading @, which gets dropped, marks an attribute name as positional).

mojavelinux:


lib/asciidoctor/extensions.rb
```diff
@@ -440,6 +440,71 @@ def default_attrs value
...
+          name, value = args.split '=', 2
+          if name.include? ':'
+            _, args = args.split ':', 2
+            option :pos_attrs, [args]
+          else
+            option :pos_attrs, []
+          end
+          option :default_attrs, { name => value }
+        else
+          _, args = args.split ':', 2 if args.include? ':'
```
mojavelinux: You are correct. I'm in the process of fixing this bug. The code got messed up when I moved some things around.

mojavelinux:
@Mogztter I just pushed a better version.

mojavelinux:
In the second commit, I added a few aliases for extension names and dropped some unused ones.

I'm shooting for a parallel (writing) structure:

```ruby
block do
  named :yell
  bound_to :paragraph
  parses_content_as :simple
  resolves_attributes '1:volume'
  process do |parent, reader, attrs|
    ...
  end
end
```

(this is not easy to get right)

mojavelinux:
A case could be made for writing that as:

```ruby
block do
  named :yell
  bound_to :paragraph
  parse_content_as :simple
  resolve_attributes '1:volume'
  process do |parent, reader, attrs|
    ...
  end
end
```

so we allow both.

Mogztter:
I think we should stick to a single rule.
In my opinion having aliases are not helping users to understand the naming rule. In other words, I don't think it improve the user/developer experience. But maybe that's just me 🤔

mojavelinux:
I'm definitely trying to converge towards a single way of writing it (or using aliases sparingly). Unfortunately, some of the aliases are needed right now to maintain backwards compatibility. When we find the right words, that will be what we'll document and we'll phase out the rest. But finding the right words is proving to be quite a challenge.

lib/asciidoctor/extensions.rb
```diff
@@ -173,13 +173,105 @@ def process *args, &block
...
+            idx, name = arg.split ':', 2
+            idx = idx == '@' ? names.size : idx.to_i
+            names[idx] = name
+          else
+            names << arg
+          end
+        end
+        option :pos_attrs, names.compact
+        option :default_attrs, defaults
+      when ::String # QUESTION should we accept Symbol too?
```
Mogztter: Positional attributes are not supported when using a `String` ? 
I'll admit that this is an edge case. But what if a user only want to resolve the second attribute ?
Eg:

```ruby
resolves_attributes '2:precision'
```

```adoc
deg:25[C,0].
```

Or (I don't know if this syntax is supported though):
```adoc
deg:25[units=C,0].
```

lib/asciidoctor/extensions.rb
```diff
@@ -173,13 +173,105 @@ def process *args, &block
...
+          name, value = args.split '=', 2
+          if name.include? ':'
+            _, name = name.split ':', 2
+            option :pos_attrs, [name]
+          else
+            option :pos_attrs, []
+          end
+          option :default_attrs, { name => value }
+        else
+          _, args = args.split ':', 2 if args.include? ':'
```
Mogztter: `pos_attrs` equals `[]` if the value doesn't contain `=` but equals `[args]` if the value contains `=`. Is this intended ?

```ruby
# Given
args = 'units=10'
# When resolves_attributes is called, Then...
#   :pos_attrs, []
#   :default_attrs, { 'units' => 10 }

# NOTE: pos_attrs is an empty array and default_attrs is defined (this is normal)

# Given
args = 'units'
# When resolves_attributes is called, Then...
#   :pos_attrs, ['units']
#   :default_attrs, {}

# QUESTION: post_attrs is defined, is this normal ?
# default_attrs is empty, this is normal.
```

mojavelinux: Yes, that is true. It's for backwards compatibility. The use of plain attribute names is no longer recommended. The second case should now be written as `'1:units'`.

lib/asciidoctor/extensions.rb
mojavelinux: Yes, that syntax is supported. There is a test to verify it.

mojavelinux: I found a more concise way of writing this so that we can support symbols. In other words, the two combinations are now recognized:

```ruby
resolve_attributes :precision
```

```ruby
resolve_attributes [:precision]
```

Although this is not the recommended syntax, at least the processor doesn't crash if you write it this way.

mojavelinux:
Let's do this!

PR #2149, opened 2017-04-25
CSS: use bold for menus.

### LINKED ISSUE #2148: CSS: use bold for menus
Menus seem to be a lot easier to read with `font-weight:bold`. Buttons are already bold, but in that case it is implemented by using a `<b>` tag in the HTML. I suppose CSS or the HTML `<strong>` tag would be more suitable in both cases.

An example:
Rendered HTML

The AsciiDoc source:
```
:experimental:

. _This is a short menu example with normal font weight._
+
Click on menu:File[Open], then save the file.

. _Here is #1 again, a short menu example, with bold font weight._
+
Click on *menu:File[Open]*, then save the file.

. _This is a longer menu example with normal font weight._
+
Click on menu:System & Hardware[Power Saving] and scroll to
*Displays*.

. _Here is #3 again, a longer menu example, with bold font weight._
+
Click on *menu:System & Hardware[Power Saving]* and scroll to
*Displays*.

```

owenh000:
Menus seem to be a lot easier to read with `font-weight:bold`. Buttons are already bold, but in that case it is implemented by using a `<b>` tag in the HTML. I suppose CSS or the HTML `<strong>` tag would be more suitable in both cases.

This commit adds `.menuseq,.menu{font-weight:bold}` to asciidoctor-default.css.

Closes #2148.


mojavelinux:
Superseded by #2164 and #2165. Thanks for getting the ball rolling!

PR #2176, opened 2017-05-04
resolves #2171 allow & to be used as the first character of the first segment of a menu

### LINKED ISSUE #2171: Allow target of menu macro to start with a character reference
Some menus begin with a symbol. However, the menu macro mandates that the target begin with a word character. The pattern should be modified to allow the target to begin with an &, which would allow any character reference to be used.

Here's an example for a reference to an item in the Chrome control menu, which is represented as a vertical ellipsis:

```
menu:&#8942;[More Tools > Extensions]
```

which can also be written as:

```
"&#8942; > More Tools > Extensions"
```

mojavelinux:


mojavelinux:
You are correct! I'm going to do that before I merge.

mojavelinux:
Tests added!

Mogztter:
:ok_hand:

PR #2178, opened 2017-05-06
use names instead of symbols in alias statements

mojavelinux:


mojavelinux:
This is what Ruby MRI uses internally, so we're going to stick with that as a convention.

PR #2179, opened 2017-05-06
split "treeprocessor" into two words; add aliases for compatibility

mojavelinux:


mojavelinux:
cc: @Mogztter

mojavelinux:
Seems like the right decision.

Issue #44, opened 2012-12-28
Create (or use an existing) message (logging) facility


mojavelinux:
We'll need to be able to report warnings (and other messages) to the user. Rather than printing these to stdout, we should pass them through a message facility which can queue them, print them with styling, etc.

jirutka:
What do you think about using core’s [Logger](http://www.ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html)? It seems that it’s quite flexible and extensible.

mojavelinux:
That has come up before, perhaps in an IRC chat. I'm not opposed to using Logger. I've just been conditioned to dislike default Logger implementations based on the disaster in the Java ecosystem. If Logger provides the flexibility we need, then it's definitely worth prototyping a solution that uses it.

Keep in mind we didn't have a Logger originally because...we'll, the processor wasn't doing sophisticated things back then :) Now, we need a proper facility for handling messages...so we're on the market for the cleanest solution.

mojavelinux:
One of the things I definitely want to do when we switch to the Logger interface is remove the need to hard code the application name with the message. The message formatter should allow us to insert this context into the message.

For an example of what I'm talking about, see https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/parser.rb#L275

lordofthejars:
Yes logger interface may be an option, for example we Can create a ruby
class which fits the Logger interface so from Asciidoctorj point of view we
won t need to do anything, rather than passing an instance of Logger into
ruby code, like we do in converters or extensions
El dt., 6 de gen., 2015 a les 7.50 Dan Allen <notifications@github.com> va
escriure:

jirutka:
> That has come up before, perhaps in an IRC chat.

Is there any other IRC channel than #asciidoctor on FreeNode IRC? I’m here almost all the time, but I’ve never seen there any core dev. :(

> I've just been conditioned to dislike default Logger implementations based on the disaster in the Java ecosystem.

JUL is indeed disaster, but Ruby is not Java and not every default logging facility is bad as in Java! It’s absolutely incomparable.

> Keep in mind we didn't have a Logger originally because...we'll, the processor wasn't doing sophisticated things back then :)

I understand that.

Well, I’ll look at it and send some suggestion how it can be possibly done.

mojavelinux:
[...]

PR #2193, opened 2017-05-13
resolves #1890 warn if doctype=inline and block has unexpected content model

### LINKED ISSUE #1890: Asciidoctor should issue a warning if there is a block but it's not a candidate for inline conversion
## Current state

``` javascript
var asciidoctor = require('asciidoctor.js')();
var processor = asciidoctor.Asciidoctor(true);

var content = "= This is a title!";
var options = {doctype: 'inline', attributes: ['showtitle']};
var html = processor.convert(content, options);
console.log(html); // returns Opal.nil
```

In the sample above, "This is a title!" is not a candidate for inline conversion and Asciidoctor returns "nil".
## Proposal

Asciidoctor should issue a warning message:

```
No candidate for inline conversion.
Inline document type should be used to convert a single paragraph. If you want to convert a complete document, use the default document type 'article' or simply remove 'doctype' from the options.
Read more about document types: http://asciidoctor.org/docs/user-manual/#document-types
```

I think this warning will help users to understand why the `convert` method returned `nil` (and outline the differences between each document type).

mojavelinux:


lib/asciidoctor/document.rb
```diff
@@ -1047,11 +1047,12 @@ def convert opts = {}
...
 
     if doctype == 'inline'
-      # QUESTION should we warn if @blocks.size > 0 and the first block is not a paragraph?
-      if (block = @blocks[0]) && block.content_model != :compound
-        output = block.content
-      else
-        output = nil
+      if (block = @blocks[0])
+        if block.content_model == :compound || block.content_model == :empty
+          warn %(asciidoctor: WARNING: no inline candidate; use the inline doctype to convert a single paragragh, verbatim, or raw block)
```
Mogztter: What do you think about adding a link on the documentation about how to use document types and/or making a suggestion on how to fix it (ie. remove doctype from options) ?

Maybe we should print this information only when verbose is turned on ?

mojavelinux: I'm not really a fan of adding links to warning messages. I think Maven tried that and it ended up being very misleading (though it could have been how they did it). I'd actually prefer the user to search because I think it leads them to very valuable information. But that doesn't mean I want to send them on a wild goose chase either. When we resolve #44, we can perhaps assign error codes that would match entries in a table in the documentation. That would get them to an explanation in the least amount of steps.

mojavelinux:
I added some tests.

lib/asciidoctor/document.rb
mojavelinux: In other words, let's discuss in #44.

PR #2194, opened 2017-05-13
resolves #1694 apply title substitutions to reftext

### LINKED ISSUE #1694: < > not escaped in anchor reftext definition
Hi there,

I have a small issue using anchor captions. When `<` or `>` are used in caption for anchor definition, the symbol are not translated in `&lt;` or `&gt;` resulting in rendering errors in HTML and parser warnings with the PDF backend.

E.g., this code:

```
See <<myuuid>>

[[myuuid, My ref title with <T>]]
== This is my real title
```

is translated as this:

```
<p>We can see the issue in <a href="#myuuid">My ref title with <T></a></p>
```

I use Asciidoctor 1.5.4

mojavelinux:
- apply title substitutions to reftext when parsing
- don't apply substitutions to value of reftext attribute enclosed in single quotes
- catalog inline anchors after creating block
- store substituted reftext in ID table and reftext attribute
- use more efficient code to scan for inline anchors
- remove dead code

### DIFF
--- a/lib/asciidoctor/attribute_list.rb
+++ b/lib/asciidoctor/attribute_list.rb
@@ -163,7 +163,7 @@ class AttributeList
           @attributes[%(#{value = value.strip}-option)] = ''
         end
         @attributes[name] = value
-      when 'title'
+      when 'title', 'reftext'
         @attributes[name] = value
       else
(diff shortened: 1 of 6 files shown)


mojavelinux:
This PR is superseded by #2220. Attribute references are substituted before the reftext is stored in the document catalog and reftext subs (a new substitution group) are applied when the reftext is inserted into the output at the location of a cross reference.

Issue #626, opened 2013-09-11
Support chunked (multi-page) HTML output in asciidoctor


gAmUssA:
Asciidoc can produce chunked html output via docbook toolchain http://www.methods.co.nz/asciidoc/chunked/ch05.html#_asciidoc_docbook_xsl_stylesheets_drivers

rhattersley:
> Here's a super quick hack using a custom converter.
> 
> https://gist.github.com/mojavelinux/d94372393950ca76d594

Invoke via:

    ruby multipage-html5-converter.rb <your-asciidoc-file>
Ensure you have the `output` directory already in place.

gAmUssA:
@mojavelinux I tried multipage-html5-converter.rb. it throws me an error

```
/usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor.rb:1362:in `rescue in load': asciidoctor: FAILED: <stdin>: Failed to parse source, undefined method `encoding' for nil:NilClass (NoMethodError)
	from /usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor.rb:1281:in `load'
	from /usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor.rb:1458:in `convert'
	from multipage-html5-converter.rb:33:in `document'
	from multipage-html5-converter.rb:18:in `convert'
	from /usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor/document.rb:1028:in `convert'
	from /usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor.rb:1499:in `convert'
	from /usr/local/opt/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/asciidoctor-1.5.2/lib/asciidoctor.rb:1562:in `convert_file'
	from multipage-html5-converter.rb:94:in `<main>'
```

gAmUssA:
disabled safe options, converted ends without errors, but `-chunked.html` Zero bytes

mojavelinux:
[...]

PR #2195, opened 2017-05-13
resolves #858 Add possibility to use figure number or text in link

### LINKED ISSUE #858: Add possibility to use figure number or text in link text

fap-:
- saves origin section of ids when set
- reftext gets subbed after document parsing is complete
  and sectnums are set correctly
- introduces new attribute :section-label: for reftext substitution
  :section-label: {sectnum} {secttitle}

I tried a different approach for the section numbers that doesn't requires to set the sectnums earlier than before.
@mojavelinux Please have a look and give me feedback, this would make https://github.com/asciidoctor/asciidoctor/pull/1921 obsolete.

### DIFF
--- a/lib/asciidoctor/document.rb
+++ b/lib/asciidoctor/document.rb
@@ -206,6 +206,7 @@ class Document < AbstractBlock
       @parent_document = nil
       @references = {
         :ids => {},
+        :id_origins => {},
         :footnotes => [],
         :links => [],
         :images => [],
@@ -562,6 +563,9 @@ class Document < AbstractBlock
       else
         @references[:ids][id] ||= reftext
       end
+    when :id_origins
+      id, origin = *value
+      @references[:id_origins][id] = origin
     when :footnotes, :indexterms
       @references[type] << value
     else
@@ -900,6 +904,19 @@ class Document < AbstractBlock
       $1 ? (apply_subs $2, (resolve_pass_subs $1)) : $2
     else
       apply_header_subs value
+    end
+  end
+
+  # Internal: apply subs for the section-label attribute
+  #
+  # This has to be an extra step because the substitution can only
+  # be done after the parsing is done and the section sectnums are
+  # set correctly.
+  def apply_section_label_subs
+    references[:id_origins].each do |id, section|
+      reftext = references[:ids][id]
+      reftext = section.sub_attributes(reftext, { :section_reftext => { :sectnum => section.sectnum, :secttitle => section.title }})
+      register(:ids, [id, reftext], true)
     end
   end
 
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -97,6 +97,7 @@ class Parser
         document << new_section if new_section
       end
     end
+    document.apply_section_label_subs
 
     document
   end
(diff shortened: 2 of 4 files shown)


mojavelinux:
While working on #1694, it became abundantly clear to me that we really need to store the node reference in the ID dictionary of the references table instead of just the text. This not only allows us to defer substitutions, it also gives us access to more information about the source.

I want to try to do this in the most minimally invasive way as possible. If that means going deeper and changing how we build the ID dictionary, then that's what we need to do. I want to avoid having "clean up" or "patch" method calls if possible.

I do like the idea of being able to refer to local attributes in the reftext. I like `sectnum`, though I'd like to use `title` instead of `secttitle` if possible.

I'm going to do a bit more research to see what impact it would have to change how we store IDs. As you have done, we may need a separate dictionary in the references table for compatibility reasons.

fap-:
Saving the node reference in the ID dictionary would also be a nice solution. I didn't do this since I can't really tell how much 3rd party tooling (non-core converter etc) this would break.
What is asciidoctors policy changes like that? I couldn't find anything about that on the [contribution page](https://github.com/asciidoctor/asciidoctor/blob/master/CONTRIBUTING.adoc).

Should I try out the suggestions you gave in https://github.com/asciidoctor/asciidoctor/issues/858 in this PR or wait for your prototype that you mentioned in https://github.com/asciidoctor/asciidoctor/pull/1921?

mojavelinux:
> Saving the node reference in the ID dictionary would also be a nice solution. I didn't do this since I can't really tell how much 3rd party tooling (non-core converter etc) this would break.

Let's leave that key alone and just use the key `:refs` instead. It's just an alternate name so we don't break anything that uses the old.

mojavelinux:
> What is asciidoctors policy changes like that?

As much as possible I try to break anything that's already using Asciidoctor. I tend to search for usages to know whether it is something that will break stuff. In this case, I know other extensions like Asciidoctor EPUB3 use the ID dictionary, so we can't touch that one for now.

mojavelinux:
> Would you like me to try out the suggestions you gave in #858 in this PR or should I wait for your prototype that you mentioned in #1921?

Give me a few days to give it a shot. If I get stuck or held up, I'll let you know so you can proceed. I just have some prototype code here already that I'd like you to see.

mojavelinux:
I think I've got something that's going to work. There are a lot of permutations at play here, but I think I found a path through the forest that should get us what we want most of the time and allow extensions to handle the rest. Stay tuned for a PR.

mojavelinux:
I'm getting closer. I've got tests written, but I ran out of time to push. I'm going to post a summary tomorrow of all the permutations that must be handled. This should get everyone thinking about whether we got the scenarios right and to make sure we are handling all the cases.

I found a few related issues that need to be solved before this patch can be applied, so I'm also working on those. In general, the foundation API was too squishy and I needed to strengthen it. But it also raised a lot of questions, such as when substitutions should be applied. We'll need to work through those step by step.

I'm also thinking that we should move the custom xrefstyle to a separate issue. I'm not quite sure the design is right on that and if we separate it, we can move this solution forward faster.

fap-:
> Let's leave that key alone and just use the key :refs instead. It's just an alternate name so we don't break anything that uses the old.

So we save the origin block to :refs for every reference?
This should also simplify the implementation of https://github.com/asciidoctor/asciidoctor/issues/626.

mojavelinux:
> So we save the origin block to :refs for every reference?

That is correct. It's what we should have done in the first place. But then again, we didn't have extensions at the time.

mojavelinux:
This PR is now superseded by #2220. I took the approach of storing the nodes in the reference table, thus gives us much greater control over how the cross reference text is generated. This eliminated the need to overload the reftext with contextual attributes. Among other benefits, this keeps a cleaner separation of content and presentation.

If we do decide that we still want contextual references scoped to the reftext, we can address that in a follow-up issue.

PR #2196, opened 2017-05-14
resolves #1516 add option to exclude tags when including file

### LINKED ISSUE #1516: Allow the option to exclude tags when using the include directive
Sometimes you want to include parts of a document, but exclude other parts (child tags). I'd like to see something like this:

include::presentation.asc[tag=content,exclude-tag=demo1]

or:

include::presentation.asc[tag=content,exclude-tags=demo1;demo4]

mojavelinux:
- in addition to including tags, add option to exclude tags using !name syntax
- introduce wildcard for including all (\*) or excluding all (!*) tags
- introduce wildcard for selecting non-tagged lines (**)
- don't select nested tags with wildcard if ancestor tag is excluded
- warn if a mismatched or unexpected end tag is found
- improve efficiency of detecting and processing tag directives
- maintain a tag stack for handling nested tags
- skip call to push_include if selected lines array is empty
- ignore tag and tags attribute if empty (previously caused an infinite loop)
- add tests for excluding tags

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -484,7 +484,7 @@ module Asciidoctor
     #     log(e);
     #   }
     #   // end::try-catch[]
-    TagDirectiveRx = /\b(?:tag|end)::\S+\[\]$/
+    TagDirectiveRx = /\b(?:tag|(end))::(\S+)\[\](?: -->)?$/
 
     ## Attribute entries and references
 
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -884,9 +884,7 @@ class PreprocessorReader < Reader
         path = PathResolver.new.relative_path include_file, @document.base_dir
       end
 
-      inc_lines = nil
-      tags = nil
-      attributes = {}
+      inc_lines, inc_tags, attributes = nil, nil, {}
       unless raw_attributes.empty?
         # QUESTION should we use @document.parse_attribues?
         attributes = AttributeList.new(raw_attributes).parse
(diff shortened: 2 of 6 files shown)


mojavelinux:
I'm still missing tests to verify that a warning is issued if an unexpected or misplaced end tag is found.

lib/asciidoctor/substitutors.rb
```diff
@@ -728,19 +728,23 @@ def sub_macros(source)
       result = result.gsub(LinkInlineRx) {
         # alias match for Ruby 1.8.7 compat
         m = $~
+        # NOTE if text is non-nil, this the macro syntax
```
Mogztter: I don't quite get this comment "this the macro syntax" ?

lib/asciidoctor/substitutors.rb
```diff
@@ -728,19 +728,23 @@ def sub_macros(source)
...
+        text = m[3]
         # honor the escape
         if m[2].start_with? RS
-          next %(#{m[1]}#{m[2][1..-1]}#{m[3]})
+          next %(#{m[1]}#{m[2][1..-1]}#{text})
```
Mogztter: I think we should aldo named `m[0]`, `m[1]` and `m[2]`

lib/asciidoctor/substitutors.rb
mojavelinux: The changes to substitutors.rb weren't supposed to be part of this PR. I'm going to pull them out into a separate PR because they are addressing something totally different.

mojavelinux:
This PR is now ready for review.

PR #2197, opened 2017-05-14
optimize link macro substitution and add additional tests

mojavelinux:
- make logic more clear
- use variable to track when we're processing a macro vs bare URL
- optimize logic
- don't parse link text of non-mailto link unless text starts with " or contains both , and =
- fix case of a nil pointer when attributes are parsed but there's no link text
- move registration of link after parsing logic
- add additional test cases


mojavelinux:
@Mogztter this is the refactoring that got stuck on the last PR.

PR #2199, opened 2017-05-15
resolves #1729 add support for include tags in languages with only circumfix comments

### LINKED ISSUE #1729: Enable tagging include regions in circumfix comments, beyond line comments

mojavelinux:


mojavelinux:
This PR depends on #2196.

mojavelinux:
Thanks for reviewing!

PR #2205, opened 2017-05-20
resolves #2204 use [ \t] instead of \p{Blank} to match spaces

### LINKED ISSUE #2204: Use [ \t] instead of \p{Blank} to match spaces
The regexp character property `\p{Blank}` matches a whole lot more than just spaces and tabs. The group includes characters such as no-break space and elongated spaces. We don't want to encourage the use of these characters as whitespace in the syntax because it's confusing and non-compliant. Therefore, most uses of `\p{Blank}` should be replaced with the simpler `[ \t]` group.

mojavelinux:
- use [ \t] (or just space) instead of \p{Blank}
- use CG_BLANK in front of bare URL
- use regexp to strip whitespace for text that follows label of admonition paragraph
- disallow tabs in certain places
- don't allow contents of block attribute list to begin with space
- optimize manpage header processing
- only match space, tab and newline when splitting attributes

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -389,13 +389,14 @@ module Asciidoctor
       CC_ALL   = '.'
       CC_ALNUM = '[:alnum:]'
       CG_ALNUM = '[[:alnum:]]'
-      CG_BLANK = '[[:blank:]]'
       CC_EOL   = '$'
       if ::RUBY_MIN_VERSION_1_9
+        CG_BLANK = '[[:blank:]]'
         CC_WORD = '[:word:]'
         CG_WORD = '[[:word:]]'
       else
         # NOTE Ruby 1.8 cannot match word characters beyond the ASCII range; if you need this feature, upgrade!
+        CG_BLANK = '[ \t]'
         CC_WORD = '[:alnum:]_'
         CG_WORD = '[[:alnum:]_]'
       end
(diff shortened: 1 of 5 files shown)


mojavelinux:
@Mogztter This is going to go a long way towards making Asciidoctor Ruby and Asciidoctor.js behave the same. I didn't realize that `\p{Blank}` was matching more than space (32) and tab.

lib/asciidoctor.rb
```diff
@@ -501,7 +500,7 @@ class << self; self; end.send :attr_accessor, key
     #                collapsing the line breaks and indentation to
     #                a single space.
     #
-    AttributeEntryRx = /^:(!?\w.*?):(?:#{CG_BLANK}+(.*))?$/
+    AttributeEntryRx = /^:(!?\w.*?):(?:\s+(.*))?$/
```
Mogztter: > /\s/ - A whitespace character: /[ \t\r\n\f]/

So if I'm reading this regular expression correctly, attribute value can be defined on a new line ?

```
:foo:
bar
```

mojavelinux: No because that regular expression is only ever run against a single line. Generally speaking, the only regular expressions that run on multiple lines are the ones that have "Inline" in the name (and often with a trailing "m"). It's not a perfect science, but that accounts for most cases.

Mogztter: Ok, thanks for the clarification

mojavelinux: We probably should use `[ \t]` here anyway since that's what we use for other line-oriented regexps. Better to be consistent.

mojavelinux:
It turns out we have to keep the CG_BLANK constant to allow any space character in front of a bare URL. That may go away in the future, but for now we need it.

mojavelinux:
I discovered that `\s` in JavaScript is the same as `\p{Space}`, which matches all blank characters (space, tab, and all space variations such as no-break space and hair space) and newlines. In Ruby, `\s` only matches space, tab, and newlines (not no-break space). So we should be very careful using `\s`.

mojavelinux:
I think this is ready to go.

I want to point out that we are dropping support for using no-break space in a number of places in the syntax (such as to offset a list marker or indent a literal paragraph), but that's consistent with AsciiDoc Python. I did a quick scan of some AsciiDoc in the wild and I'm amazed at how often no-break spaces are used unintentionally. For any documentation site I run, I'd fail the build if these characters try to creep in (except for when preceding punctuation marks in languages like French).

PR #2209, opened 2017-05-23
resolves #2208 store number for formal block on node

### LINKED ISSUE #2208: Store number assigned to formal block on node object
Currently, the number assigned to a formal block (i.e., a block that have both a title and a label) is not stored independently on the node. Instead, it is generated, then immediately joined with the label and stored in the value of the caption attribute. This puts it out of reach of extensions (unless they crudely extract it from the value of the caption attribute).

The number property should be pulled up from Section to AbstractBlock since it's relevant for various block nodes. For nodes which it is not relevant or set, the value should remain nil.

This is a prerequisite for solving #858.

mojavelinux:
- pull number property up from Section to AbstractNode
- store number for block when assigning caption
- consolidate logic in assign_caption method

### DIFF
--- a/lib/asciidoctor/abstract_block.rb
+++ b/lib/asciidoctor/abstract_block.rb
@@ -19,6 +19,11 @@ class AbstractBlock < AbstractNode
   # Public: Get/Set the caption for this block
   attr_accessor :caption
 
+  # Public: Get/Set the number of this block (if section, relative to parent, otherwise absolute)
+  # Only assigned to section if automatic section numbering is enabled
+  # Only assigned to formal block (block with title) if corresponding caption attribute is present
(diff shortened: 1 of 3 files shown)


mojavelinux:
@fap- This is a key step in being able to do labeled block references. I was kind of surprised to find out that the only place the number was stored was inside the caption. That obviously doesn't offer a lot of flexibility. I was able to extract it without breaking anything, so this is a straightforward change.

mojavelinux:
You'll notice we still have some terminology problems, but we'll likely deal with those in the 1.6.x timeframe. (What AsciiDoc calls the caption is really the label. The caption is the label + title. And label is label name + number).

PR #2216, opened 2017-05-26
drop redundant non-greedy flag in regular expressions

mojavelinux:
- when performing a negated closing delim, no need for non-greedy flag


mojavelinux:
Although this flag is redundant, I'm going to leave it because it makes the regular expression slightly faster (for some reason). I might then use it in other places as well.

mojavelinux:
Actually, I was wrong. I don't think the performance difference is measurable. Therefore, I'll proceed with #2277.

Issue #61, opened 2013-01-08
Parse inline markup during the parse phase (use recursive descent parser)


mojavelinux:
Currently, the parsing of inline content happens during rendering. This limits the information you have in the document after parsing. Instead, Asciidoctor should parse all text extents into inline nodes during parsing.

This requires moving the substitutions from the rendering phase to the parsing phase.

It also means that each line in the buffer will become an array of inline nodes that represent the chunked text, toggling between plain text and elements like links and images.

Since this has the potential to slow down a single pass parse/render, we may want to have a flag which controls the phase in which inline text is parsed.

benignbala:
Should I have a go at this using ANTLR4 (https://rubygems.org/gems/antlr4/versions/0.9.2) ? Thanks

benignbala:
Or rather, shall we try it with ANTLR4 in Java in the asciidoctorj project ?

mojavelinux:
There's a project setup w/ ANTLR to do this experimentation. See https://github.com/asciidoctor/asciidoc-grammar-prototype

mojavelinux:
Test case:

The following should not be matched as emphasis (i.e., italics):

```
*_id* word_
```

elextr:
I provided advice to another project that was trying to develop an Asciidoc implementation in C++ for an embedded project.  That project has now been dropped (the embedded project and therefore the Asciidoc implementation :disappointed: ), but here are some lessons for future implementers of this parser.

Using the traditional parser tools is problematical, they all assume context free grammers, and Asciidoc is definitely not a CFG.  Even the tools various ways of handling contextual dependence did not seem sufficient for this case (YMMV but don't waste too much time on it, so far nobody made it work).  The example in the post above is a simple context dependence (parsing the `_id` as italic should stop if it hits the end of any quotes it is nested in, thats the context it depends on).

The big context dependence is of course the the different parsing between blocks (and changing that with the `subs=` option, since it changes the context dependence dynamically during parsing).  So you have to recognise the block type or `subs` option early and apply it to the appropriate part of the input, even before you have parsed the appropriate input.  

An alternative that was explored was to parse everything and then just revert the parses that were not required back to text.  Unfortunately what is parsed affects how other things parse, eg if quotes are not parsed because the opening is inside a macro target and the ending is outside that will change if subsequently its discovered that macros are not in the `subs=` list, eg
```
[subs=-macros]
the italicised part of http:xxx__yyy[]__ frobnicates the foo bar
```

[...]

PR #2220, opened 2017-05-28
resolves #858 add support for formal xref text

### LINKED ISSUE #858: Add possibility to use figure number or text in link text

mojavelinux:
- store referenceable nodes under refs key in document catalog
- when registering a ref, store entry in both refs table and deprecated ids table
- update tests to use consolidated register method
- add xreftext method to AbstractBlock, Section, and Inline to produce formatted text for xref
- add API docs to the xreftext methods
- format xref text according to xrefstyle attribute, if set (full, short, or basic)
- move reftext? and reftext methods down to AbstractBlock and Inline; contextualize
- apply reftext substitutions (specialchars, quotes, replacements) to value returned by reftext method
- drop support for reftext document attribute
- don't catalog inline anchor if reftext containing attribute reference resolves to empty string
- use reftext method instead of text method for :ref in DocBook converter
- add feature tests for explicit reftext
- add feature tests for generated reftext using xrefstyle
- introduce language attributes for chapter, section, and appendix reference signifier
  * name attributes chapter-refsig, section-refsig, and appendix-refsig, respectively
- enable chapter-label attribute in translations file; map chapter-refsig attribute to it
- map appendix-refsig attribute to value of appendix-caption attribute in translations file
- add placeholders in translations file for chapter-refsig where missing and section-refsig
- handle refsig document attribute values consistently (only use if attribute is non-nil)
- update link and xref tests to populate refs key in document catalog
- encode double quote and strip XML tags in value of xreflabel attribute in DocBook converter
- don't store return value of target in inline_anchor method of HTML converter

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -631,7 +631,7 @@ class Parser
           block = next_item_list(reader, :ulist, parent)
           if (style || (Section === parent && parent.sectname)) == 'bibliography'
             attributes['style'] = 'bibliography' unless style
-            block.items.each {|item| catalog_inline_biblio_anchor item.instance_variable_get(:@text), document }
+            block.items.each {|item| catalog_inline_biblio_anchor item.instance_variable_get(:@text), item, document }
           end
           break
 
@@ -726,8 +726,6 @@ class Parser
             return
           end
 
-          catalog_inline_anchors lines * LF, document
-
           # NOTE don't check indented here since it's extremely rare
           #if text_only || indented
           if text_only
@@ -768,6 +766,8 @@ class Parser
             adjust_indentation! lines if indented && style == 'normal'
             block = Block.new(parent, :paragraph, :content_model => :simple, :source => lines, :attributes => attributes)
           end
+
+          catalog_inline_anchors lines * LF, block, document
         end
 
         break # forbid loop from executing more than once
@@ -904,16 +904,16 @@ class Parser
       block.source_location = source_location if source_location
       # FIXME title should have already been assigned
       block.title = attributes.delete 'title' if attributes.key? 'title'
+      #unless attributes.key? 'reftext'
+      #  attributes['reftext'] = document.attributes['reftext'] if document.attributes.key? 'reftext'
+      #end
       # FIXME caption should have already been assigned
       block.caption ||= attributes.delete 'caption'
       # TODO eventually remove the style attribute from the attributes hash
       #block.style = attributes.delete 'style'
       block.style = attributes['style']
-      # AsciiDoc Python always use [id] as the reftext in HTML output,
-      # but I'd like to do better in Asciidoctor
       if (block_id = (block.id ||= attributes['id']))
-        # TODO sub reftext
-        document.register(:ids, [block_id, (attributes['reftext'] || (block.title? ? block.title : nil))])
+        document.register :refs, [block_id, block, attributes['reftext'] || (block.title? ? block.title : nil)]
       end
       # FIXME remove the need for this update!
       block.attributes.update(attributes) unless attributes.empty?
@@ -1192,36 +1192,42 @@ class Parser
     found
   end
 
-  # Internal: Catalog any inline anchors found in the text, but don't process them
+  # Internal: Catalog any inline anchors found in the text (but don't convert)
   #
   # text     - The String text in which to look for inline anchors
-  # document - The current document in which the references are stored
+  # block    - The block in which the references should be searched
+  # document - The current Document on which the references are stored
   #
   # Returns nothing
-  def self.catalog_inline_anchors text, document
+  def self.catalog_inline_anchors text, block, document
     text.scan(InlineAnchorScanRx) do
       if (id = $1)
-        reftext = $2
+        if (reftext = $2)
+          next if (reftext.include? '{') && (reftext = document.sub_attributes reftext).empty?
+        end
       else
         id = $3
-        if (reftext = $4) && (reftext.include? ']')
-          reftext = reftext.gsub '\]', ']'
+        if (reftext = $4)
+          reftext = reftext.gsub '\]', ']' if reftext.include? ']'
+          next if (reftext.include? '{') && (reftext = document.sub_attributes reftext).empty?
         end
       end
-      document.register :ids, [id, reftext]
+      document.register :refs, [id, (Inline.new block, :anchor, reftext, :type => :ref, :target => id), reftext]
     end if (text.include? '[[') || (text.include? 'or:')
     nil
   end
 
-  # Internal: Catalog any bibliography inline anchors found in the text, but don't process them
+  # Internal: Catalog the bibliography inline anchor found in the start of the list item (but don't convert)
   #
-  # text     - The String text in which to look for inline bibliography anchors
-  # document - The current document in which the references are stored
+  # text     - The String text in which to look for an inline bibliography anchor
+  # block    - The ListItem block in which the reference should be searched
+  # document - The current document in which the reference is stored
   #
   # Returns nothing
-  def self.catalog_inline_biblio_anchor text, document
+  def self.catalog_inline_biblio_anchor text, block, document
     if InlineBiblioAnchorRx =~ text
-      document.register :ids, [$1, %([#{$2 || $1}])]
+      # QUESTION should we sub attributes in reftext (like with regular anchors)?
+      document.register :refs, [(id = $1), (Inline.new block, :anchor, (reftext = %([#{$2 || id}])), :type => :bibref, :target => id), reftext]
     end
     nil
   end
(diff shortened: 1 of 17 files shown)


mojavelinux:
This is just a WIP. The next step is to extract the logic from the converter and settle which substitutions are applied to the reftext and when.

cc: @fap-

lib/asciidoctor/parser.rb
```diff
@@ -1234,7 +1235,10 @@ def self.catalog_inline_anchors text, document
     text.scan(InlineAnchorRx) do
       # lead with assignments for Ruby 1.8.7 compat
       id, reftext = $1 || $3, $2 || $4
-      document.register :ids, [id, reftext] unless $&.start_with? '\\'
+      unless $&.start_with? '\\'
+        document.register(:ids, [id, reftext])
+        document.register(:refs, [id, reftext])
```
fap-: Shouldn't this save the block instead of the reftext?
The method should be called in Parser#next_block when a block is already initialized, so we can pass the block instead of document.

fap-: After reading more of the code, it feels the `Inline` object would be the right thing to save but those are only created during substitution in the conversion phase.
I would probably try to introduce a value object that gets saved into :refs. It just wraps the block/section and in the case of an inline anchor, it knows from which index of lines the ref is coming. Just so we have a pointer to the origin of the ref in all cases.

mojavelinux: Unfortunately, until #61 is resolved, we don't have a node to store at this phase. We could consider using special type of Inline node that could hold the reftext and the parent node. I'll see what I can do.

mojavelinux: Okay, I was able to use the Inline node for this. It's actually emulating the node that is later created during conversion, so logically it fits.

mojavelinux:
@fap- I've removed the [wip] label from this PR, so it's now ready for final review.

There are some notable advancements achieved here:

* references are stored in the :refs table in the document catalog in addition to the :ids table
  * the :refs table is now the preferred location for storing :refs moving forward since it stores the target node
  * however, we have to support both tables for backwards compatibility (and may continue to use the :ids table just as a reftext index)
* attributes references in reftext are substituted before being stored in the document catalog
* in the HTML output, normal substitutions (minus attributes) are applied to reftext when used at the location of an xref (aka cross reference)
  * this functionality is provided by the reftext method on AbstractBlock and Inline
* the HTML converter uses the xreftext method on AbstractBlock, Section, and Inline to generate the text at an xref location (this is the primary enhancement introduced by this change)
  * this method first calls reftext, then generates text if the target node does not define reftext
* in the DocBook output, normal substitutions (minus attributes) are applied to reftext when used in the xreflabel attribute of the node (DocBook uses the value of the xreflabel to generate the text at the xref location)

mojavelinux:
I still may need to add some additional tests. That's what I'll be working on tomorrow.

mojavelinux:
Oh yeah. I was also about to figure out how to apply formatting to the segments of the xreftext without putting the code in the converter (where it would be harder to reuse). I apply substitutions to a template string, then insert the formatted title. I should be able to reuse this now in Asciidoctor PDF.

mojavelinux:
@fap- I'm wondering if we should just name the chapter and section attributes `chapter-caption` and `section-caption`, respectively. The reason I switched to the `label` suffix is because `caption` isn't really the right word. But perhaps we should change them all at once (in 1.6) rather than have a mixed set. wdyt?

fap-:
@mojavelinux The code looks good to me, but you've touched parts of asciidoctor that I'm still unfamiliar with, so take it for what it's worth :sweat_smile: 

I'm in favor of the renaming to use the same descriptor for everything. I would even like to call everything {name}-label but that would probably too big of a change.

I understand that this is probably personal preference but I find those deeply nested if/else cascade quite hard to read. What about multiple/early returns?
AbstractBlock:
```
  def xreftext xrefstyle = nil
    return val if (val = reftext) || !@title
    # NOTE formal block == title + caption
    # FIXME admonition blocks abuse the caption property; exclude for now
    return title unless xrefstyle && @caption && (ctx = @context) != :admonition
    case xrefstyle
    when 'full'
      quoted_title = sprintf sub_quotes(@document.compat_mode ? %q(``%s'') : '"`%s`"'), title
      if @number && (label = @document.attributes[ctx == :image ? 'figure-caption' : %(#{ctx}-caption)])
        %(#{label} #{@number}, #{quoted_title})
      else
        %(#{@caption.chomp '. '}, #{quoted_title})
      end
    when 'short'
      if @number && (label = @document.attributes[ctx == :image ? 'figure-caption' : %(#{ctx}-caption)])
        %(#{label} #{@number})
      else
        @caption.chomp '. '
      end
    else
      title
    end
  end
```

Another suggestion would be to introduce a formal_block? method so the comment could be removed (my understanding is that we would return nil in case of !@title, which should also be the case with this new version):
AbstractBlock:
```
  def xreftext xrefstyle = nil
    return val if (val = reftext)
    # FIXME admonition blocks abuse the caption property; exclude for now
    return title unless xrefstyle && formal_block? && (ctx = @context) != :admonition
    case xrefstyle
    when 'full'
      quoted_title = sprintf sub_quotes(@document.compat_mode ? %q(``%s'') : '"`%s`"'), title
      if @number && (label = @document.attributes[ctx == :image ? 'figure-caption' : %(#{ctx}-caption)])
        %(#{label} #{@number}, #{quoted_title})
      else
        %(#{@caption.chomp '. '}, #{quoted_title})
      end
    when 'short'
      if @number && (label = @document.attributes[ctx == :image ? 'figure-caption' : %(#{ctx}-caption)])
        %(#{label} #{@number})
      else
        @caption.chomp '. '
      end
    else
      title
    end
  end

  def formal_block?
    @title && @caption
  end
```

mojavelinux:
> I'm in favor of the renaming to use the same descriptor for everything. I would even like to call everything {name}-label but that would probably too big of a change.

That's really the question at hand. We want to move to the {name}-label convention, but that's too big of a change right now (and will only further delay this release). Thus the question is whether we want to a) use the old convention for the new attributes and change them all later, or whether we want to b) use the new name for the new attributes and change the old ones later.

I'm leaning towards (a), staying consistent and switching all the names later. That way, remapping the old to new names is easier because we can apply a common rule.

> I understand that this is probably personal preference but I find those deeply nested if/else cascade quite hard to read.

I'm more interested in getting feedback on the user experience right now than the code. This level of nesting is nothing compared to other places in core. Although the dense logic doesn't bother me, the real reason it's done is because method call lookup is expensive in Ruby. I try to avoid unnecessary method calls, esp for internal logic. That's a trade-off we have to make in order for Asciidoctor to be as fast as it is.

> Another suggestion would be to introduce a formal_block? method

This is an interesting idea. I will consider that change. Thanks.

mojavelinux:
> I'm leaning towards (a), staying consistent and switching all the names later. That way, remapping the old to new names is easier because we can apply a common rule.

This is the approach I've decided to take.

mojavelinux:
> Another suggestion would be to introduce a formal_block? method

I decided not to do this because it's actually more than just a formal block. A formal block is a block with a title. What we're looking for is a block with a title and a number. There's not really a name for that...so for now we're just going to look for those properties.

mojavelinux:
I thought about the translation attribute names more, and the idea of using section-caption and chapter-caption really bothers me. While I want to remain consistent, those names just don't make any sense. What we are trying to expression is a prefix, specifically a number prefix. The first ideas I had were:

* chapter-number-prefix
* chapnum-prefix
* chapter-caption-prefix
* chapter-prefix

However, I did some digging into LaTeX and I found that the key it uses to control this language string is `chaptername`. Since the terminology for this is so poorly defined, we may get mileage out of aligning with LaTeX. Therefore, I propose:

* chapter-name

(and hence section-name and eventually part-name).

Are we okay with chapter-name (at least for the time being) or do you like one of the other proposals better?

mojavelinux:
I did some more research into the terminology and determined that the concept we are shooting for is a "signifier". But more specifically, a reference signifier. We want to signify that the reference is referring to a chapter, section, or appendix. With that in mind, I've updated my proposal to the following:

* chapter-refsig (where refsig is short for "reference signifier")

The benefit of this attribute pattern is that it's now possible to control the text used in a reference independent of the text used in the heading (which is not yet implemented).

I went ahead and mapped the attribute for appendix as well so that it can be controlled independently of what is currently the appendix-caption. It also means that all section types follow the same attribute name pattern.

* chapter-refsig
* section-refsig
* appendix-refsig
* part-refsig (future)

mojavelinux:
I'm prepared to roll this out.

fap-:
Great idea. I'm not really sure that I like the latex naming convention. It feels just as wrong as "caption".
I would still prefer "label" for the actual heading but if you think that are large part of the user base is familiar with the latex terminology it's better to follow their convention.

Also congratulations on merging this! :tada: :smile:

mojavelinux:
Thank you for your help and support!

> I'm not really sure that I like the latex naming convention.

In the end, I didn't like it either, which is why I decided to introduce new terminology. See http://asciidoctor.org/docs/user-manual/#customizing-the-cross-reference-text

PR #2230, opened 2017-05-31
resolves #2217 only drop known AsciiDoc extensions from inter-document xref path

### LINKED ISSUE #2217: Extensionless inter-document cross reference is truncated at period
If the path of an inter-document cross reference contains a period, but does not include the AsciiDoc file extension, the path is truncated at the location of the period.

Example:

```
<<using-.net-web-services#,Using .NET web services>>
```

The remainder after the final period is interpreted as a file extension and stripped away. The inter-document cross reference logic should only remove known AsciiDoc file extensions.

The workaround is to include the file extension in the target.

```
<<using-.net-web-services.adoc#,Using .NET web services>>
```

mojavelinux:


test/links_test.rb
```diff
@@ -295,6 +295,16 @@
     assert_xpath '//a[@href="tigers.html"][text() = "[tigers]"]', doc.render, 1
   end
 
+  test 'inter-document xref should not truncate after period if path has no extension' do
+    result = render_embedded_string '<<using-.net-web-services#,Using .NET web services>>'
+    assert_xpath '//a[@href="using-.net-web-services.html"][text() = "Using .NET web services"]', result, 1
+  end
+
+  test 'inter-document xref should only truncate file extension if path contains period' do
```
Mogztter: "if path contains a known AsciiDoc extensions" ?
In my opinion, the title of this test is a bit misleading since the path contains two periods `using-.net-web-services.adoc` (ie. `.net` and `.adoc`)

mojavelinux: I'll update to make it more clear.

PR #2231, opened 2017-05-31
resolves #2227 add the remove_attr method to AbstractNode

### LINKED ISSUE #2227: Add remove_attr method to AbstractNode
Add the method remove_attr to AbstractNode (to complement set_attr). This method removes the named key from the attributes on the current node. It should not trigger any side effects.

Consider aliasing this method to delete_attr or unset_attr. Note that "unset" is consistent with the terminology in the documentation.

mojavelinux:


Mogztter:
Should we remove `Document.removeAttribute` function in Asciidoctor.js (and add a new function `AbstractNode.removeAttribute`) ? 

https://github.com/asciidoctor/asciidoctor.js/blob/91fb33c0868ba921afa70661905f8e949f0b0b76/src/asciidoctor-core-api.js#L504-L507

mojavelinux:
I would say yes. If we did want to keep the one on document, it should be renamed to "deleteAttribute", though really that's an internal method as we previously discussed. I realize it's still a little confusing, but until we decide what do with those attribute-related APIs on the Document object, it's best to just not expose them.

PR #2241, opened 2017-06-07
no double quotes in xref macro

mojavelinux:
- don't allow double quotes around id
- don't strip double quotes around reftext
- update relevant tests


mojavelinux:
For context, there was one case of this in the AsciiDoc User Guide, but the behavior was never specified. I can't find any other use of this syntax, so I think we should just drop it. The quotes serve absolute no purpose (other than the slow down parsing).

PR #2282, opened 2017-06-27
Fix for #1121 - Introduce four tilde syntax for open blocks and deprecate '--'

### LINKED ISSUE #1121: Allow open blocks to be nested
I'm using _asciidoctor_ to replace a website which is made in a _wysiwyg html editor_  to be responsive. I have to use _openblocks_ frequently and it may be cool to nest those in to each other. I think a signing character can be solve this, for example:

```
[.parent]
--A
[.child]
--B
content
--B
[.another_child]
--B
content
--B
--A
```

Wdyt?

To know what i mean, check the [old site](http://www.szizaltrend.hu/nadfonatok.html) and the [new](http://www.ro.naturtrend.com/ratanandthonet.html) _asciidoctor_ version. I need nested `div`s to vertically align the tables on the left and the text on the right side, like this:

```
+--------------------+
|Container   +------+|
|+----------+|Text  ||
||Table div ||div   ||
|+----------+|      ||
|            +------+|
+--------------------+
```

TobiasHector:
Doing this as a monkey patch right now, so getting it into the core of asciidoctor would be good. Don't know what your take on deprecation is though - should there be a warning? I've added one for now, but can take it out.

Fixes #1121

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -287,7 +287,9 @@ module Asciidoctor
   VERBATIM_STYLES = ['literal', 'listing', 'source', 'verse'].to_set
 
   DELIMITED_BLOCKS = {
+    # DEPRECATED: The two hyphen syntax for open blocks is deprecated - tildes should be used instead as these allow nesting of open blocks.
     '--'   => [:open, ['comment', 'example', 'literal', 'listing', 'pass', 'quote', 'sidebar', 'source', 'verse', 'admonition', 'abstract', 'partintro'].to_set],
+    '~~~~' => [:open, ['comment', 'example', 'literal', 'listing', 'pass', 'quote', 'sidebar', 'source', 'verse', 'admonition', 'abstract', 'partintro'].to_set],
     '----' => [:listing, ['literal', 'source'].to_set],
     '....' => [:literal, ['listing', 'source'].to_set],
     '====' => [:example, ['admonition'].to_set],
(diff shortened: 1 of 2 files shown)


mojavelinux:
patch by TobiasHector

mojavelinux:
Scheduled for 2.0.0.

elextr:
Why remove the two dash open block?

Ok, fine #1121 has some use-cases for adding variable length open block delimiters, and it will make it like the other delimited blocks, but it wouldn't be any problem to leave the two-dash version too.  

Quite frankly the four plus tilde version is ugly and using open blocks a lot in list item extensions it looks good with short delimiters.

Its fine to have the ugly four tilde version for when its needed, but leave the short two dash version for most cases where nesting isn't required.

mojavelinux:
> Why remove the two dash open block?

I should have clarified. That won't be removed. This PR isn't going is as is. We're just introducing the 4+ character version.

> Its fine to have the ugly four tilde version for when its needed, but leave the short two dash version for most cases where nesting isn't required.

100%

(I don't agree it's ugly, though, but otherwise :+1:)

elextr:
> > Why remove the two dash open block?

> I should have clarified. That won't be removed. 

Ok, no worries then.

> I don't agree it's ugly, though

Maybe its the Github font, it does look better in my editor.

> I might do a compatibility flag that you can disable if you really want the forbid the two-character delimiter.

Wouldn't that be an in-compatibility flag [sorry] :grin:  

Your call but seems effort for not much gain.

mojavelinux:
> Wouldn't that be an in-compatibility flag [sorry] grin

Yes, it would be incompatible if switched off.

mojavelinux:
Closed in favor of #3010.

PR #2286, opened 2017-06-30
fix the no bullet list styles

mojavelinux:
- none means hide the bullet / number, but preserve indentation
- no-bullet or unnumbered (olist only) means hide the bullet / number and remove only list item indentation
- unstyled means hide the bullet / number and remove all indentation
- allow none and no-bullet to be used on ulist or olist


mojavelinux:
This aligns with the behavior of Asciidoctor PDF.

Issue #1957, opened 2016-12-01
Add helper method in extensions to create new section


mojavelinux:
Add the convenience method `create_section` to the base Processor class for the purpose of creating sections inside an extension.

mojavelinux:
Here's what I'm imagining:

```ruby
def create_section parent, title, attrs, opts = {}
  ...
end
```

For example:

```ruby
create_section parent, 'Section Title', {}
```

The level can be specified as an option (`:level`). Otherwise, it's assumed to be one more than the parent. The id attribute will be generated automatically from the title if not specified unless the value is `false`. Numbered can be specified as an option (`:numbered`), otherwise it defaults to whether the `sectnums` attribute is specified.

PR #2304, opened 2017-07-09
enhance create_section helper in extensions to set up section properly

mojavelinux:
- set up section in the same way the parser does
- set sectname
- set numbered
- set special
- use default level based on parent
- add rdoc
- add tests


mojavelinux:
This is a follow-up to #1957 to incorporate the changes to the model that were added to support formal cross reference text.

PR #2305, opened 2017-07-09
Adds a Rake task to trigger build on dependent projects

Mogztter:


mojavelinux:
Great start! I made some revision and started to hook it into the Travis CI build. I managed to get it working locally with a Travis token for the asciidoctor-docbot. What's left is to figure out how to make it only run once in the build matrix.

mojavelinux:
It's kind of a hack, but I think I got it working. When this is merged, it should work.

If you'd like, it's possible to customize the request so it only executes certain parts of the asciidoctor.js build. Right now, it will just trigger a full build against master.

Mogztter:
We are doing the following in Asciidoctor.js to run task only once:

```
after_success:
 - >
  if ! [ "$AFTER_SUCCESS_RUN" ]; then
    export AFTER_SUCCESS_RUN=1;
    bundle exec rake build:dependents
  fi
````

mojavelinux:
Clever!

mojavelinux:
Let's give it a shot!

mojavelinux:
Thanks!

mojavelinux:
I managed to get this working...too well at first. It was triggering hundreds of jobs on Asciidoctor.js. Now we're down to just one per core build.

However, it's confusing to see the commit message of the last commit be reused for the triggered build. I think we should force a custom message. Let's open a separate issue for that.

Mogztter:
> However, it's confusing to see the commit message of the last commit be reused for the triggered build. I think we should force a custom message. Let's open a separate issue for that.

👍

PR #2308, opened 2017-07-10
resolves #2244 warn if duplicate ID is detected

### LINKED ISSUE #2244: Warn if a duplicate ID is detected
When explicit IDs are used in a document, there's a chance the author has assigned the same ID to different nodes. Asciidoctor should detect this scenario and report a warning. The warning indicates that the output document (e.g., HTML) will have a duplicate ID, which is invalid (though usually ignored by the browser).

mojavelinux:


mojavelinux:
Due to how lines are scanned, it's only possible to report the line number for a section. If we keep the start line of a block, then we can use it in the report.

mojavelinux:
@fap- and @vogella, I figured you might appreciate this one. It was made possible because of the refactoring done for the formal cross reference text.

vogella:
Thanks, @mojavelinux. That is definitely helpful.

Issue #2314, opened 2017-07-12
Use a custom commit message when triggering dependent builds


mojavelinux:
When the core build triggers dependent builds (build:dependents), it should use a custom commit message that indicates why the build is being rerun instead of using the message from the last commit in the dependent repository. This means customizing the payload that is sent to Travis CI.

See https://docs.travis-ci.com/user/triggering-builds/

mojavelinux:
I want to add other projects to the list, such as AsciidoctorJ and Asciidoctor Diagram, but I'd like to fix this before we do so that it's clear what is happening in those downstream projects.

mojavelinux:
cc: @Mogztter

mojavelinux:
Resolved by f07edd8b3575bb9e264042e1111052c02cfe69e4.

PR #2315, opened 2017-07-13
Trigger upstream build of AsciidoctorJ with every Travis build

robertpanzer:


mojavelinux:
This is great. The only thing I'd like to do is get it to update the commit message of the job to something like:

"build triggered by change to asciidoctor/asciidoctor"

Otherwise, it's not clear why the build is running. I think it's just a matter of passing an additional value in the JSON payload.

See https://docs.travis-ci.com/user/triggering-builds/#Customizing-the-commit-message.

Can you add that? That would solve #2314.

mojavelinux:
I'm also not sure if it's possible to trigger a build on the 1.6. branch as well. Can you see how? If not, we can just do master for now.

robertpanzer:
Yes, good point! Totally forgot about 1.6.0. 
I'll check that tomorrow.

mojavelinux:
I'm going to merge this then push a change to customize the commit message. I'd still like to explore how we can trigger the build on the 1.6 branch. It probably means we need to update the configuration set to include a target branch.

Mogztter:
The SHA1 of the commit could be nice to quickly understand why a build is failing.

mojavelinux:
ah, good point. would you put that in the subject of the message? or in the body?

Mogztter:
It will be useful in the message, especially when browsing the build history on Travis:

sha1

With this view we can quickly see when the build start failing and get the corresponding sha1:
"Build triggered by change to Asciidoctor core (29bd89d)"

But we need to give it a try because I don't know if Travis will try to automatically turn it into a link on Asciidoctor.js or AsciidoctorJ (instead of Asciidoctor).

I wonder if the body text is shown on the build history ? :thinking:

mojavelinux:
I like this idea. I guess the only way we'll know is to try, so I'll just start throwing some data at it and see what comes out.

To save space, I'll drop the word "core" as that is understood.

mojavelinux:
> I wonder if the body text is shown on the build history ?

It's shown on the build detail page, but it's kind of faded out. We'll see if we can squeeze it into the subject.

mojavelinux:
Okay, the URL is sort of grayed out, but you can triple click to select it.

https://travis-ci.org/asciidoctor/asciidoctor.js/builds/253796777

I'll also include the textual commit hash in the subject line.

mojavelinux:
Let me know if you like that.

Mogztter:
It's nice, can we use the short format for SHA1 (7 characters) ?

mojavelinux:
Unfortunately, Travis doesn't put the short form of the SHA1 into an environment variable. So we'd have to either calculate it ourselves or use git to calculate it. But I think it's better just to have the full value. Why discard information? If you feel strongly we should use the short format, I'd need to know what strategy you'd like to use to retrieve it.

mojavelinux:
I tested and the URL works if we just truncate the SHA1 value to 10 characters. So perhaps we should just do that. wdyt?

mojavelinux:
We'll truncate to 8 chars. We technically only need 4 to be unique in the core project, so that should be plenty.

Mogztter:
8 chars is perfect 👌

Mogztter:
It looks great on mobile:

screenshot_20170721-094550

mojavelinux:
This is nice! Thanks for sharing!

PR #2344, opened 2017-07-29
Pass env var BUILD_REASON=UPDATE_UPSTREAM to downstream builds

robertpanzer:
If I understood correctly we are still looking for a way to let downstream builds only run against the latest upstream version when Asciidoctor is built. (https://github.com/asciidoctor/asciidoctor.js/issues/363)

Therefore this PR adds an env variable to the downstream builds `BUILD_REASON=UPDATE_UPSTREAM` that would allow the downstream builds to detect that only a build against the current upstream is necessary.

I think in AsciidoctorJ it should be easily possible to skip the "normal" build when this var is set.
I didn't find how the upstream build is done in Asciidoctor.js, but I guess it should be possible there as well to consider this.

I could not test this, therefore I hope that it works and that I didn't add any error.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -154,7 +154,18 @@ desc 'Trigger builds for all dependent projects on Travis CI'
       if (commit_hash = ENV['TRAVIS_COMMIT'])
         commit_memo = %( (#{commit_hash.slice 0, 8})\\n\\nhttps://github.com/#{ENV['TRAVIS_REPO_SLUG'] || 'asciidoctor/asciidoctor'}/commit/#{commit_hash})
       end
-      payload = %({ "request": { "branch": "#{branch}", "message": "Build triggered by Asciidoctor#{commit_memo}" } })
+      payload = %({ 
+                    "request": { 
+                      "branch": "#{branch}", 
+                      "message": "Build triggered by Asciidoctor#{commit_memo}", 
+                      "config": {
+                        "merge_mode": "deep_merge",
+                        "env": {
+                          "BUILD_REASON": "UPDATE_UPSTREAM"
+                        }
+                      }
+                    } 
+                  })
       (http = Net::HTTP.new 'api.travis-ci.org', 443).use_ssl = true
       request = Net::HTTP::Post.new %(/repo/#{org}%2F#{name}/requests), header
       request.body = payload


mojavelinux:
This is exactly what I had in mind. In theory, it should be possible to veto builds in the matrix based on environment variable...so it's not something you'd have to check for in the task itself.

I'd prefer the following variable name and reason:

* TRIGGER=upstream

That reads a bit better and is not so verbose.

mojavelinux:
Actually, you know what would work better? If we used YAML references. Imagine that one or matrices are defined using a YAML reference (or perhaps excludes are defined). Then, the YAML defined in this build script would just reference that YAML stanza, thus activating it. That would separate the configuration from the intent to use it.

mojavelinux:
Here's my idea. We pass the following payload:

```ruby
payload = %({
  "request": {      
    "branch": "#{branch}", 
    "message": "Build triggered by Asciidoctor#{commit_memo}", 
    "config": {       
      "matrix": "*upstream_only_matrix"
    } 
  } 
})
```

This relies on a matrix to be defined as follows:

```yaml
matrix: &upstream_only_matrix
  exclude:
  - env: ASCIIDOCTOR_CORE_VERSION=v1.5.5
  - env: ASCIIDOCTOR_CORE_VERSION=v1.5.6.1
matrix: {}
```

See https://github.com/asciidoctor/asciidoctor.js/pull/366

That essentially defined a YAML reference, then resets the matrix (to whatever the default matrix would be).

Of course, this would require that the upstream_only_matrix reference is defined in the config.

wdyt?

cc: @Mogztter

mojavelinux:
The problem with setting an env variable is that it would interfere with the build configuration, esp in Asciidoctor.js. The matrix idea is the cleanest way to redefine what gets built. But I'm open to other ideas if you have them.

mojavelinux:
What I'm not sure is whether Travis will expand the reference when it is defined this way. I guess we'll need to test it.

mojavelinux:
Actually, if the env was defined as a global env, then it wouldn't interfere with the matrix. So that is another option.

```
payload = %({
  "request": {
    "branch": "#{branch}",
    "message": "Build triggered by Asciidoctor#{commit_memo}",
    "config": {
      "env": {
        "global": ["TRIGGER=upstream"]
      }
    }
  }
})
```

mojavelinux:
My idea doesn't work because Travis doesn't honor the YAML reference when merging the config.

mojavelinux:
We could store a separate file in the repository that contains the upstream only config (.travis-upstream-only.yml). Then we could read it if it exists, merge it in, and pass it on. That would put all control in the downstream repo.

mojavelinux:
Yep, this idea is totally going to work and is the most flexible choice. It puts the responsibility in the right place.

mojavelinux:
This is now open for use. Just create a file named .travis-upstream-only.yml in your repository and it will be used as the config that is passed through to the payload. You can even control the merge_mode since the merge_mode is a child of the config key.

Here's an example in action:

https://travis-ci.org/asciidoctor/asciidoctor.js/builds/264019035

PR #2355, opened 2017-08-06
Resolves #2349, check mandatory attributes when creating an image block

### LINKED ISSUE #2349: create_image_block method should check that alt and target attribute are defined
> I do think this should be checked in the create_image_block method. In fact, it should check for target too, which is also required. Let's pursue that change in core.

https://github.com/asciidoctor/brackets-asciidoc-preview/issues/40#issuecomment-319449094

Mogztter:
`target` and `alt` attributes are mandatory to create an image block.

### DIFF
--- a/lib/asciidoctor/extensions.rb
+++ b/lib/asciidoctor/extensions.rb
@@ -147,9 +147,9 @@ module Extensions
       Block.new parent, context, { :source => source, :attributes => attrs }.merge(opts)
     end
 
-    # Public: Creates an image block attached to the parent.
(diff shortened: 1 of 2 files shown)


mojavelinux:
I made a slight change. I decided to generate the alt text if not specified, just like the core parser does.

I also updated the example to be a more playful, since an idea struck me.

mojavelinux:
Oh, and I also set default-alt, which is recommended if the alt is generated.

mojavelinux:
Thanks!

PR #2356, opened 2017-08-06
resolves #2272 use the quote element in DocBook converter

### LINKED ISSUE #2272: Use <quote> element in DocBook backend
Currently quotes are written unmodified ('"' stays '"') into DocBook. In my understanding, these should be replaced with proper `<quote>...</quote>` elements: http://tdg.docbook.org/tdg/5.0/quote.html
Otherwise it is currently not possible to the downstream processors to properly handle the quotes (according to the selected language).

bk2204:
Fixes #2272.


mojavelinux:
Thank you for making this fix! I can't believe I was practically staring at this the whole time and didn't see it.

PR #2371, opened 2017-08-17
resolves #2369 add XML circumfix comments to .js and .ts files

### LINKED ISSUE #2369: Potential regression on tag inclusion in 1.5.6.1
Previously, using `1.5.5`, we could do something like:

```asciidoc
= Hello, AsciiDoc!
include::include.ts[tags=h1]
```

with `include.ts` looking like:

```typescript
const template = `
<body>
  <!-- tag::h1[] -->
  <h1>Hello</h1>
  <!-- end::h1[] -->
</body>`;
```

which was producing the expected result:
capture d ecran 2017-08-15 a 18 14 04

Upgrading to `1.5.6.1` leads to the following error:

```
asciidoctor: WARNING: issue.adoc: line 2: tag 'h1' not found in include file: /Users/cedric/Code/tests/ascidoctor-tag-issue/include.ts
```

and the HTML is no longer displayed.

This is a minimal reproduction, but I've seen it on every project I tried to bump to the latest release.

mojavelinux:
- JavaScript / TypeScript often have embedded HTML, so treat as XML files

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -363,7 +363,7 @@ module Asciidoctor
   CIRCUMFIX_COMMENTS = {
     ['/*', '*/'] => ['.css'],
     ['(*', '*)'] => ['.ml', '.mli', '.nb'],
-    ['<!--', '-->'] => ['.html', '.xhtml', '.xml', '.xsl'],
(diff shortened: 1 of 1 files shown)


mojavelinux:
We're going in a different direction. See #2683.

PR #2379, opened 2017-08-21
always activate extension registry passed to processor

mojavelinux:


test/extensions_test.rb
```diff
@@ -554,6 +555,19 @@ def process parent, reader, attributes
       assert_equal 0, Asciidoctor::Extensions.groups.size
     end
 
+    # NOTE I'm not convinced we want to continue to support this use case
+    test 'can provide extension registry created without any groups as option' do
+      registry = Asciidoctor::Extensions.create
+      registry.tree_processor SampleTreeProcessor
```
Mogztter: What about:

```ruby
registry.register :sample do
  tree_processor SampleTreeProcessor
end
```

Full comment: https://github.com/asciidoctor/asciidoctor.js/pull/372#issuecomment-323742003

mojavelinux: Let's take that up separately since it's different in scope to the regression.

mojavelinux:
After merging, Asciidoctor.js still has one test failure. Is this related to the regression?

Mogztter:
> After merging, Asciidoctor.js still has one test failure. Is this related to the regression?

Yes, previously extensions were enabled by default.
So I had to change this test: https://github.com/asciidoctor/asciidoctor.js/pull/372/files#diff-ab51605a3a8c43a6518adc56c2e0fed9R164

test/extensions_test.rb
Mogztter: Ok.
I'm not sure how to label this feature though. I think we should stick to the DSL and prevent users from creating "detached" objects. But on the other side we want to access those objects to be able to perform actions on them (register, unregister, activate...).

I will create one issue per "use case" so we can discuss on each "use case" if we want to support it and how we should implement it.

PR #2380, opened 2017-08-21
update nokogiri to 1.8.0 for ruby >= 2.1

miltador:
This fixes dependencies installation on Windows with ruby >= 2.4.

### DIFF
--- a/asciidoctor.gemspec
+++ b/asciidoctor.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'erubis', '~> 2.7.0'
   # haml is needed for testing custom templates
   s.add_development_dependency 'haml', '~> 5.0.0'
-  s.add_development_dependency 'nokogiri', '~> 1.7.0'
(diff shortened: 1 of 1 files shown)


mojavelinux:
I updated it to apply to Ruby 2.1 as well since that is still compatible.

mojavelinux:
Thanks!

miltador:
[x64-mingw32](https://rubygems.org/gems/nokogiri/versions/1.8.0-x64-mingw32) and [x86-mingw32](https://rubygems.org/gems/nokogiri/versions/1.8.0-x86-mingw32) versions require Ruby 2.2 while [general](https://rubygems.org/gems/nokogiri/versions/1.8.0) requires 2.1. That's why I put the higher denominator of 2.2 version.

mojavelinux:
that's kind of insane. hmm, I'd rather make an exception for that one case then.

mojavelinux:
see if that works

miltador:
It works, thanks!

PR #2418, opened 2017-09-17
disable jruby.cext.enabled flag in CI

mojavelinux:


mojavelinux:
It turns out the flag is already disabled. JRuby 9k just happens to complain about the flag being set at all. So we'll just ignore.

PR #2446, opened 2017-10-08
resolves #2445 put id attribute on tag around phrase instead of preceding anchor

### LINKED ISSUE #2445: Put id on phrase instead of as anchor in HTML output
When an ID is assigned to a phrase (span of formatted text), put the the id attribute on the tag that encloses the phrase instead of as an anchor before it.

Example:

```
[#bond]_007_
```

Output:

```
<em id="bond">007</em>
```

instead of:

```
<a id="bond"></a><em>007</em>
```

The proposed output is the more intuitive result.

mojavelinux:


mojavelinux:
@lordofthejars wdyt?

PR #2453, opened 2017-10-13
resolves #1572 set max-width on .imageblock a.image for Firefox

### LINKED ISSUE #1572: Images with link should not overflow
Hi,
In my documentation source file, I am declaring a picture with a link, like this:

```
image::Dockerized_Environment.svg[role="text-center", link="Images/Dockerized_Environment.svg"]
```

However, with the HTML output, the image overflows if there is not enough space.
This issue disappears when removing the `link` attribute, and I did not notice this behavior in 1.5.2 .

After digging a little, I found that `img` has a `max-width:100%;`, but `.image` doesn't, which seems to cause this issue.

mojavelinux:


mojavelinux:
Closing since this is no longer a problem as of the current release of Firefox (and several before it). But we can revisit if it comes back or another browser has a problem.

PR #2467, opened 2017-11-02
require core if not loaded when requiring extensions

mojavelinux:
- force core to load if extensions are loaded first

### DIFF
--- a/lib/asciidoctor/extensions.rb
+++ b/lib/asciidoctor/extensions.rb
@@ -1,3 +1,5 @@
+require 'asciidoctor'.to_s unless defined? Asciidoctor
+
 # encoding: UTF-8
 module Asciidoctor
 # Extensions provide a way to participate in the parsing and converting


mojavelinux:
This is specifically for Jekyll, which loads core lazily. Therefore, extensions could get loaded before core, so we want to guarantee that order.

lib/asciidoctor/extensions.rb
```diff
@@ -1,3 +1,5 @@
+require 'asciidoctor'.to_s unless defined? Asciidoctor
```
Mogztter: @mojavelinux just for my curiosity, why are you using the `to_s` on a String ? As far as I know, `to_s` returns the String representation of an Object, so `'asciidoctor'` and `'asciidoctor'.to_s` should be equivalent no ?

mojavelinux:
It's a trick to hide the import from Opal since it's never needed in that
environment.

Mogztter:
Oh I see, thanks for explaining!

jirutka:
This really should be documented using a comment, because no one else can know the purpose of this weird hack and even you will surely not remember it after few weeks. Just a common sense…

mojavelinux:
Fair enough. I added a note above each usage. Opal only supports static require statements, which is why this works.

PR #2468, opened 2017-11-04
promote Font Awesome version to a constant

mojavelinux:


mojavelinux:
We should probably do the same for highlight.js.

Issue #2471, opened 2017-11-07
First heading should determine if Markdown compliance should be on or not


ds26gte:
If the first header of the document is AsciiDoc-style (i..e, with `=`s), do _not_ turn on Markdown compliance. This allows writers to continue using `#`, `*`, `-`, `>` without having to work around to avoid special behavior.

Mentioned in #2074.

ds26gte:
Implementing this is easy enough: In `parser.rb`, just after

```
    if Compliance.markdown_syntax ? ((line1.start_with? '=', '#') && ExtAtxSectionTitleRx =~ line1) :
        ((line1.start_with? '=') && AtxSectionTitleRx =~ line1)
```

add

```
      if (line1.start_with? '=')
          Compliance.markdown_syntax = false
      end
```

However, a bunch of tests (`bunch exec rake`) that check for Markdown compliance need to be rewritten, because Markdown compliance won't happen anymore without an `=` header.

ds26gte:
My PR (#2482) was OK on a `bundle exec rake`, but seems to fail the Travis CI build. Any thoughts, @mojavelinux?

ds26gte:
Any movement on this? I offered a PR: the implementation itself is straightforward, but some tests need a rewrite and I don't know to temporarily force Markdown-compliance in the tests for Markdown-compliance.

mojavelinux:
I apologize for the delay, @ds26gte. I've been insanely busy recently getting Antora into shape.

I think this change takes the idea too far. It's perfectly legitimate to use an AsciiDoc-style document title and still use Markdown compliance. I do that all the time.

We also can't have a document change the static state of Asciidoctor. That violates thread safety and would introduce some hard to understand bugs in programs. Compliance settings are meant to be changed when you load Asciidoctor.

The reason I use a static property (essentially a constant) is because it's extremely fast and we can access it from anywhere. If we changed to using a document attribute (kind of like compat-mode), then it's slightly slower and could be out of reach at times (though I have to check that).

Is it possible for you to change the compliance setting manually when loading Asciidoctor?

ds26gte:
Thanks, I see what you're saying and have probably learned to work around any surprises by now. 

[...]

PR #2482, opened 2017-11-16
Have parse_section_title turn off Markdown compliance on seeing = header

ds26gte:
This addresses #2471.  Tests updated.


mojavelinux:
This is not something we're going to do right now in Asciidoctor. Besides, you can't change a global setting from the context of a single document as that is not thread safe and would introduce unexpected race conditions when converting a lot of documents.

PR #2489, opened 2017-11-19
Mention Homebrew in README-fr

Mogztter:
> To install Asciidoctor on OS X with Homebrew, first follow the instructions at https://brew.sh/[brew.sh] to install Homebrew.

In French the first sentence was a bit repetitive so I've decided to remove the first part.

> Homebrew installs the `asciidoctor` gem into an exclusive prefix that's independent of system gems.

I do understand the meaning but it's hard to translate in French because you don't install something into a "prefix".
So I replaced "into" by "avec" (with) but it's not accurate because the gem is not install with a prefix...

@zmwangx Could you please elaborate on this point so I can find a better translation ? The gem is installed in a dedicated directory that's independent of system gems ?

### DIFF
--- a/README-fr.adoc
+++ b/README-fr.adoc
@@ -144,7 +144,7 @@ Asciidoctor fonctionne mieux lorsque vous utilisez UTF-8 partout.
 
 == Installation
 
-Asciidoctor peut être installé en utilisant la commande (a) `gem install`, (b) Bundler ou (c) les gestionnaires de paquets pour les distributions Linux populaires.
+Asciidoctor peut être installé en utilisant la commande (a) `gem install`, (b) Bundler, (c) les gestionnaires de paquets pour les distributions Linux populaires, ou (d) Homebrew pour OS X.
 
 TIP: L'avantage d'utiliser le gestionnaire de paquets pour installer la gemme est que l'installation englobe celle des librairies Ruby et RubyGems si elles ne sont pas déjà installés.
 L'inconvénient est que le paquet n'est pas forcément mis à jour immédiatement après la mise à disposition de la gemme.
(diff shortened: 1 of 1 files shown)


zmwangx:
> In French the first sentence was a bit repetitive so I've decided to remove the first part.

Yeah, I was kind of following the style of other subsections like dnf and apt-get... They are indeed repetitive.

> because you don't install something into a "prefix".

I'm loosely using the term "prefix" in the sense of the traditional `configure --prefix`; think of the standard [`prefix` variable](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) from GNU coding standards. For `gem` this is basically `GEM_HOME`; in fact, here's what we do inHomebrew:

```rb
  def install
    ENV["GEM_HOME"] = libexec
    system "gem", "build", "asciidoctor.gemspec"
    system "gem", "install", "asciidoctor-#{version}.gem"
    bin.install Dir[libexec/"bin/*"]
    bin.env_script_all_files(libexec/"bin", :GEM_HOME => ENV["GEM_HOME"])
  end
```

where `libexec` translates to `$HOMEBREW_PREFIX/Cellar/asciidoctor/$version/libexec`. So the resulting structure of installation is

```console
$ tree -L 3 /usr/local/Cellar/asciidoctor/1.5.6.1/
/usr/local/Cellar/asciidoctor/1.5.6.1/
├── CHANGELOG.adoc
├── INSTALL_RECEIPT.json
├── LICENSE.adoc
├── README.adoc
├── bin
│   ├── asciidoctor
│   └── asciidoctor-safe
└── libexec
    ├── bin
    │   ├── asciidoctor
    │   └── asciidoctor-safe
    ├── build_info
    ├── cache
    │   └── asciidoctor-1.5.6.1.gem
    ├── doc
    │   └── asciidoctor-1.5.6.1
    ├── extensions
    ├── gems
    │   └── asciidoctor-1.5.6.1
    └── specifications
        └── asciidoctor-1.5.6.1.gemspec

11 directories, 10 files
```

Mogztter:
@zmwangx Thanks for the detailed explanation, that makes sense 👍

mojavelinux:
> In French the first sentence was a bit repetitive so I've decided to remove the first part.

Same for English. I think the sentence got mangled during the editing process.

mojavelinux:
Thanks @Mogztter! Looks good!

PR #2505, opened 2017-12-08
fixed wrong command name.

aokiplayer:
The correct command name is 'asciidoctor', not 'asciidoctorjs'.


mojavelinux:
Nice catch! Thanks!

PR #2514, opened 2017-12-18
store URI include as URI object in include stack

mojavelinux:


mojavelinux:
@Mogztter Is this use of URI going to cause a problem in Asciidoctor.js? I think all we need is a stub for 'uri' that does the following:

```ruby
module URI
  def self.parse str
    str.extend URI
  end
end
```

We don't really care about the parsing. It's more for the identity.

mojavelinux:
This is aside from whether Asciidoctor.js supports remote includes since URI will be used in other places when this is merged.

Mogztter:
> Is this use of URI going to cause a problem in Asciidoctor.js? I think all we need is a stub for 'uri' that does the following

:+1: 
https://github.com/asciidoctor/asciidoctor.js/pull/398

Mogztter:
Actually to implement OpenURI in Opal, we will need (at some point) to implement URI :wink:

mojavelinux:
> Actually to implement OpenURI in Opal, we will need (at some point) to implement URI

That's true. I'm just want to be sure we don't break Asciidoctor.js in the interim.

Now we just need to figure out how to coordinate this change. I guess if you merge the change in Asciidoctor.js, then I merge in core, the integration build will continue to work.

mojavelinux:
Let's give this a shot!

PR #2530, opened 2018-01-08
resolves #2529 don't throw exception if text of dd node is nil

### LINKED ISSUE #2529: text method of description details with only block content shouldn't fail
If the description details node (dd) only consists of blocks (not text), then the text method throws an exception. It should instead return nil.

This happens because the text method invokes apply_subs without first checking whether the value of the text property is nil. While the text property is never nil for normal list items, it can be nil for a description details node that only contains blocks.

Example:

```
term::
+
block content
```

This exception does not arise in the built-in converters since they always check `has_text?` first.

mojavelinux:


mojavelinux:
Merged as e6dbe4a4606b8852296bb53c546d7898587284c7

PR #2547, opened 2018-01-28
Fix wrong `Asciidoctor::AbstractBlock#lineno` of nested lists and tables

seikichi:
`Asciidoctor::AbstractBlock#lineno` returns wrong line number for nested lists
and nested tables. This commit fixes the problem  (save `reader.cursor` as a
local variable before doing lookahead).

Example:

```ruby
require 'asciidoctor'
doc = Asciidoctor.load(<<EOF, sourcemap: true)
* foo
** bar
EOF
nested_list = doc.find_by(context: :ulist).last
puts nested_list.lineno # => 3 (expect 2)
```

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -853,7 +853,8 @@ class Parser
         block = build_block(block_context, :compound, terminator, parent, reader, attributes)
 
       when :table
-        block_reader = Reader.new reader.read_lines_until(:terminator => terminator, :skip_line_comments => true), reader.cursor
+        cursor = reader.cursor
+        block_reader = Reader.new reader.read_lines_until(:terminator => terminator, :skip_line_comments => true), cursor
         # NOTE it's very rare that format is set when using a format hint char, so short-circuit
         unless terminator.start_with? '|', '!'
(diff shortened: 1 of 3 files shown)


seikichi:
See https://github.com/asciidoctor/asciidoctor/pull/2548 for travis error.

seikichi:
I noticed that there are still bugs ... 

```ruby
require 'asciidoctor'

doc = Asciidoctor.load(<<EOS, sourcemap: true)
|===
a|
foo
|===
EOS
foo_para = doc.blocks[0].rows.body[0][0].inner_document.find_by(context: :paragraph)[0]
puts foo_para.lineno # => 2 (expect 3)
```

mojavelinux:
Thanks!

I'm seeing more and more cases where we really need to bookmark the cursor at the start of next_block. I'm going to experiment with that approach and see if it gives us the results we want. If that's the case, I'll apply it to this PR.

mojavelinux:
The changes from this PR have been incorporated in #2701. The code change was exactly right. I only made minor updates to the code style.

mojavelinux:
And thanks!

PR #2558, opened 2018-02-11
base_dir can be a uri

Mogztter:
If `base_dir` is an URI, use `URI.parse` instead of `::File.expand_path`. Otherwise the current working directory will be prepended to the URI (thus making the `base_dir` invalid and unusable).

Also remove the special condition for Opal and `xmlhttprequest ` since it's not required anymore.

Ref: 
- https://github.com/asciidoctor/asciidoctor-browser-extension/issues/210
- https://github.com/asciidoctor/asciidoctor.js/pull/384
- https://github.com/asciidoctor/asciidoctor.js/issues/382

### DIFF
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -836,11 +836,6 @@ class PreprocessorReader < Reader
       if @include_stack.size >= abs_maxdepth
         warn %(asciidoctor: ERROR: #{line_info}: maximum include depth of #{@maxdepth[:rel]} exceeded)
         return
-      elsif ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
-        # NOTE resolves uri relative to currently loaded document
-        # NOTE we defer checking if file exists and catch the 404 error if it does not
-        target_type = :file
-        inc_path = relpath = @include_stack.empty? && ::Dir.pwd == @document.base_dir ? target : %(#{@dir}/#{target})
       elsif (Helpers.uriish? target) || ((::URI === @dir) && (target = %(#{@dir}/#{target})))
         unless @document.attributes.key? 'allow-uri-read'
           replace_next_line %(link:#{target}[])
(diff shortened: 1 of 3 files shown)


lib/asciidoctor/reader.rb
```diff
@@ -836,11 +836,6 @@ def preprocess_include_directive raw_target, raw_attributes
       if @include_stack.size >= abs_maxdepth
         warn %(asciidoctor: ERROR: #{line_info}: maximum include depth of #{@maxdepth[:rel]} exceeded)
         return
-      elsif ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
-        # NOTE resolves uri relative to currently loaded document
-        # NOTE we defer checking if file exists and catch the 404 error if it does not
-        target_type = :file
-        inc_path = relpath = @include_stack.empty? && ::Dir.pwd == @document.base_dir ? target : %(#{@dir}/#{target})
       elsif (Helpers.uriish? target) || ((::URI === @dir) && (target = %(#{@dir}/#{target})))
```
Mogztter: The next step is to use `URI.join` instead of `%(#{@dir}/#{target})` so the following will also work:

```ruby
base_dir = 'http://localhost:1234'
target = './name/asciidoctor'
resolved_target = 'http://localhost:1234/name/asciidoctor'
```

```ruby
base_dir = 'http://localhost:1234/name'
target = '../name/asciidoctor'
resolved_target = 'http://localhost:1234/name/asciidoctor'
```

But I don't want to introduce two changes at once 😉

Mogztter: To make it work on Asciidoctor.js we will also need to implement `URI.join` in Opal.

mojavelinux:
When I looked over my notes, I found I had written to myself "allow base_dir to be a URI", so we are definitely on the same page here.

mojavelinux:
I've opened https://github.com/asciidoctor/asciidoctor/issues/2592 to take a deeper look at what we have to do to allow base_dir to be a URI. There are multiple places in the code where the assumption is made that base_dir is an absolute path...so we'll need to cover all those cases to do it right.

Mogztter:
I'm closing this pull request since the code has greatly evolved and we need to revisit what needs to be done.

PR #2573, opened 2018-02-21
Small CSS fix.

sindzicat:
Pseudo-elements `::before` and `::after` require two dots before them.
[Mozilla docs for ::before](https://developer.mozilla.org/en-US/docs/Web/CSS/::before)


mojavelinux:
I agree, the correct modern style is two colons. The minification tool in the stylesheet factory was reducing it to one. I think that's because IE8 only supported the single form, but IE8 is no longer supported ("Effective January 12, 2016, Internet Explorer 8 is no longer supported on any client or server version of Windows"), so we don't have to worry about it.

There are a lot more places to update, but I'll allow the script that generates the stylesheet to handle that, which I'll merge after merging this change.

sindzicat:
Is there unminified version of `asciidoctor-default.css` in repo? I wasn't able to find such file...

mojavelinux:
https://github.com/asciidoctor/asciidoctor-stylesheet-factory#asciidoctor-stylesheet-factory

PR #2575, opened 2018-02-23
resolves #2574 record timings when calling convert and write on Document

### LINKED ISSUE #2574: Record timings correctly when calling load, convert, write sequence
Record the timings correctly when using the sequence load, convert, write instead of convert or convert_file:

```rb
doc = Asciidoctor.load_file 'path/to/doc.adoc', safe: :safe, (t = Asciidoctor::Timings.new)
output = doc.convert
doc.write output, 'out.html'
t.print_report
```

This outputs something like:

```
  Time to read and parse source: 0.25263
  Time to convert document: 0.00000
  Total time (read, parse and convert): 0.25263
```

Currently, the timings for the convert (and write) steps are not recorded.

The reason this happens is because the timings for convert and write are computed inside the Asciidoctor.convert API instead of inside the convert and write methods on Document. Saving the timings option in the document, and using it inside the Document's convert method will solve this issue.

mojavelinux:
- store timings object on Document
- record timing for convert in Document#convert
- record timing for write in Document#write

### DIFF
--- a/lib/asciidoctor/timings.rb
+++ b/lib/asciidoctor/timings.rb
@@ -7,11 +7,11 @@ module Asciidoctor
     end
 
     def start key
-      @timers[key] = ::Time.now
+      @timers[key] = now
     end
 
     def record key
-      @log[key] = (::Time.now - (@timers.delete key))
+      @log[key] = (now - (@timers.delete key))
     end
 
     def read_parse
@@ -35,6 +35,17 @@ module Asciidoctor
       to.puts %(  Time to read and parse source: #{'%05.5f' % read_parse.to_f})
       to.puts %(  Time to convert document: #{'%05.5f' % convert.to_f})
       to.puts %(  Total time (read, parse and convert): #{'%05.5f' % read_parse_convert.to_f})
+    end
+
+    if ::Process.respond_to? :clock_gettime
+      CLOCK_ID = ::Process::CLOCK_MONOTONIC
+      def now
+        ::Process.clock_gettime CLOCK_ID
+      end
+    else
+      def now
+        ::Time.now
+      end
     end
   end
 end
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -1287,6 +1287,7 @@ module Asciidoctor
   # Returns the Document
   def load input, options = {}
     options = options.dup
+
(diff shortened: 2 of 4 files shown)


lib/asciidoctor/timings.rb
```diff
@@ -36,5 +36,16 @@ def print_report to = $stdout, subject = nil
       to.puts %(  Time to convert document: #{'%05.5f' % convert.to_f})
       to.puts %(  Total time (read, parse and convert): #{'%05.5f' % read_parse_convert.to_f})
     end
+
+    if ::Process.respond_to? :clock_gettime
+      CLOCK_ID = ::Process::CLOCK_MONOTONIC
```
Mogztter: @mojavelinux `clock_gettime` is defined in an Opal environment but not `::Process::CLOCK_MONOTONIC`:

```
CLOCK_MONOTONIC: uninitialized constant Process::CLOCK_MONOTONIC
```

https://travis-ci.org/asciidoctor/asciidoctor.js/jobs/345568804

mojavelinux: Dang. What are the chances? That's okay, we can just check for both.

mojavelinux: JavaScript can use Time.now with good accuracy, so this is just a Ruby thing.

mojavelinux: I see a reference to CLOCK_MONOTONIC (which is mapped to performance.now()). 

https://github.com/opal/opal/blob/478f5f15e41ef8ba398d995846532b17aed82be1/opal/corelib/process.rb#L30

Is that not in a release yet?

Mogztter: Indeed but the constant was removed: https://github.com/opal/opal/commit/319f9c07ea0111e2f8488903cfc0f585d28c2847

> Is that not in a release yet?

I need to check if it's available in Asciidoctor.js.

> I see a reference to CLOCK_MONOTONIC (which is mapped to performance.now()).

I don't know if it's safe to use this implementation because `performance.now` and `process.hrtime` may not be available in a Nashorn environment.

mojavelinux: Good point. We can always put an Opal guard on it if necessary.

Issue #2497, opened 2017-11-26
Allowing newlines in asciimath stem block


hgiesel:
This is basically a continuation of #2484. At the moment a stem block which utilizes AsciiMath (which is the default) like this:

```
[stem]
++++
f: bbb"N" -> bbb"N"
f: x |-> x + 1
++++
````

...is parsed like this:

```html
<div class="stemblock">
<div class="content">
\$f: bbb"N" -&gt; bbb"N"
f: x |-&gt; x + 1\$
</div>
</div>
```

However this will display both lines on the same line. [AsciiMath does not support linebreaks in formulas](https://github.com/asciimath/asciimathml/issues/49). However this is confusing to the AsciiDoc user, because stem blocks aim to facilitate multiline input for code segments, or similiar.
I could think of three solutions for this:

1. 
We simply parse the above AsciiDoc like this:
```html
<div class="stemblock">
<div class="content">
\$f: bbb"N" -&gt; bbb"N"\$<br/>
\$f: x |-&gt; x + 1\$
</div>
</div>
```
..., which will display two lines. This is the simplest solution and fits with the minimalist approach of AsciiMath, however users being used to this will then most likely be surprised, if they do this:

```
[stem,latexmath]
++++
f: \mathbb{N] \to \mathbb{N}
f: x \mapsto x + 1
++++
```

... and get everything on a single line (because you need the `\\` to mark a newline in MathJax)

2.
We extend the AsciiMath by a token like `\\`, however I don't know how much work this will incur, and it seems to me it's the least appealing approach.

3.
Ask the team behind MathJax to implement something like AsciiMath blocks. However this does not really seem necessary, because in the issue I linked above, the AsciiMath team basically says it does not want to concern itself with the layout aspects. This is easily done in HTML, because you can use `br` tags, however you can't do this in AsciiDoc, because it is abstracted away.

dimztimz:
Double backslash `\\` is taken in Asciimath. Single newline is deliberately ignored.

I was thinking about this and come up to 2 possible solutions.

1. **Double new line** should mark a new line in asciimath block. With double new line I think that the intent for new line is clearly stated. This is analogous to writing double new line for new paragraph.
2. If we want a more tightly packed ascii input, allow **backlash newline** to break. Currently that translates to "backslash space" in the html which is rendered as forced space (e.g. if you want multiple spaces not to get collapsed into one, you prefix them with backslash to get forced space).

[...]

PR #2577, opened 2018-02-25
Support newline in asciimath stem block in HTML5 output.

dimztimz:
Newline `<br />` will be emitted with double newline or backslash
and newline in the input.

Related to #2497 .


mojavelinux:
I started with your PR and built on it to create #2579.

Here's what I added or changed.

* promoted regexp to a constant
* pre-compute entire replacement string
* use guard to prevent running gsub on a single line
* used \n instead of \R in the regexp since we can assume all AsciiDoc content is normalized
* look for double backslash, not single backslash
* strip spaces in front of double backslash
* add tests

mojavelinux:
The alternate proposal was merged as 2034a5f3f5a67ddb2bd549957abb547064ae8c87.

PR #2587, opened 2018-03-02
drop is_ prefix from methods in PathResolver

mojavelinux:


mojavelinux:
Must have been before I learned all the Ruby idioms ;)

PR #2588, opened 2018-03-03
resolves #1365 allow table stripes to be configured using stripe option on table

### LINKED ISSUE #1365: Allow table striping to be controlled by role or attribute

mojavelinux:
- valid values are even, odd, all, and none
- not set by default; implies stripes on even rows
- requires support from stylesheet

### DIFF
--- a/data/stylesheets/asciidoctor-default.css
+++ b/data/stylesheets/asciidoctor-default.css
@@ -257,6 +257,8 @@ table.grid-all>tbody>tr:last-child>.tableblock,table.grid-all>thead:last-child>t
 table.frame-all{border-width:1px}
 table.frame-sides{border-width:0 1px}
 table.frame-topbot{border-width:1px 0}
+table.stripe-all tr,table.stripe-odd tr:nth-of-type(odd){background:#f8f8f7}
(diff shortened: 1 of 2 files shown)


mojavelinux:
I think the option should be named "stripe" to be congruent with the frame and grid options. The CSS classes should also follow suite.

It's possible to control this behavior either using the option or the role directly. After all, the option just sets the role.

* stripe=even|odd|all|none
* .stripe-even, .stripe-odd, .stripe-all, .stripe-none

I don't add the role by default to reduce bloat. If we decide it should be added by default (like with frame and grid), I can update the PR.

PR #2596, opened 2018-03-07
resolves #2595 PathResolver#root? returns true for absolute URL in browser env

### LINKED ISSUE #2595: PathResolver#root? should return true for absolute URL when IO module is xmlhttprequest
The PathResolve#root? method should return true for an absolute URL (in addition to file://, which is already supported) when the IO module is xmlhttprequest (Opal environment only). This allows the root? method to work correctly with the URL-based filesystem used by the browser extensions.

mojavelinux:


lib/asciidoctor/path_resolver.rb
```diff
@@ -135,16 +135,17 @@ def initialize file_separator = nil, working_dir = nil
...
   # path - the String path to check
   #
   # returns a Boolean indicating whether the path is an absolute root path
   if RUBY_ENGINE == 'opal'
     def root? path
       (path.start_with? SLASH) ||
-          (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (path.start_with? 'file://')) ||
-          (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
+          (@file_separator == BACKSLASH && (WindowsRootRx.match? path)) ||
+          (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (path.start_with? 'file://', 'http://', 'https://'))
```
Mogztter: I didn't know that `start_with?` accepted a list of arguments! It's good to know :smile:

mojavelinux: I only discovered that recently and was like :champagne:

Issue #2592, opened 2018-03-04
Allow value of base_dir option (or docdir) to be a URI


mojavelinux:
While the base_dir option was originally designed to be the starting point for system references on the local filesystem, there are cases when it should be allowed to be a URI. One such example is when the document source is a string that is the contents of a URI. In this case, references such as includes should be resolved from the URI instead of from a path on the local filesystem. This use case comes up in the browser extensions, which may begin by processing a remote document.

There are a few places were we assume the base_dir option is a local directory. However, the Reader class is already URI-aware, so we just need a broader alignment.

diguage:
I think it shoud add a `base_url` attribute, is stands for the site url of the document. 

The `base_dir` attribute still stands for the working directory.

mojavelinux:
It's not exposed for security reasons. And it's really internal state anyway.

PR #2598, opened 2018-03-08
resolves #2587 don't expand value of base_dir option in browser env if value is a URL

mojavelinux:


lib/asciidoctor/document.rb
```diff
@@ -383,16 +383,20 @@ def initialize data = nil, options = {}
...
 
-    # If the base_dir option is specified, it overrides docdir and used as the root for relative
-    # paths. Otherwise, the base_dir is the directory of the source file (docdir), if set, or else
+    # If the base_dir option is specified, it overrides docdir and is used as the root for relative
+    # paths. Otherwise, the base_dir is the directory of the source file (docdir), if set, otherwise
     # the current directory.
-    if options[:base_dir]
-      @base_dir = attr_overrides['docdir'] = ::File.expand_path(options[:base_dir])
+    if (base_dir_val = options[:base_dir])
+      if ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (Helpers.uriish? base_dir_val)
```
Mogztter: I think this change is too conservative.
`File.expand_path` does not play well with URIs and we will have issues down the line:

```ruby
p Dir.pwd # /home/guillaume
p File.expand_path('file:///path/to/asciidoctor.js') # /home/guillaume/file:/path/to/asciidoctor.js
p File.expand_path('https://asciidoctor.org') # /home/guillaume/https:/asciidoctor.org
```

In my opinion, if `base_dir_val` is an URI we should not use `File.expand_path` no matter what the runtime is.

mojavelinux: I agree with you, but we're not ready for that until #2592, which is a much larger change. Right now we are focused just on the xmlhttprequest environment.

Outside of the xmlhttprequest environment, base_dir must be an absolute path. This is a deep assumption in Asciidoctor that is going to take some effort to weed out.

mojavelinux: (we can assume it's okay in the browser environment since it doesn't do system_path operations).

Mogztter: > I agree with you, but we're not ready for that until #2592, which is a much larger change. Right now we are focused just on the xmlhttprequest environment.

Makes sense.

> (we can assume it's okay in the browser environment since it doesn't do system_path operations).

I think it's not needed because this is an intermediate step and browser environment almost always use xmlhttprequest anayway

mojavelinux: When we do fix #2592, then we'll store a URI value as a URI type. That way, we can easily detect what type of value we have an act appropriately. That will be something we'll do in 1.5.7.

mojavelinux: What we're saying is that until #2592, only the browser environment (xmlhttprequest) is allowed to set the base_dir to a URI (URL or file://). We may or may not need it, but I think it's worth starting there at least.

mojavelinux: Note that with the current changes in both branches, the other option is to set the docdir attribute from the API. That value is not modified. That may even void the need for the exception in the base_dir logic. It was needed for Antora, which has a virtual filesystem and therefore expanding docdir against the physical filesystem makes no sense.

PR #2601, opened 2018-03-09
resolves #2599 don't modify URI or absolute path include target (xmlhttprequest IO module)

### LINKED ISSUE #2599: If the include target is a URL, don't modify it when the IO module is xmlhttprequest
If the target of the include directive is a URL, use the value as is when the IO module is xmlhttprequest. Don't attempt to prepend the value of the `@dir` instance variable from the reader.

mojavelinux:
- add PathResolver#absolute_path? method
- isolate PathResolver#absolute_path? and PathResolver#root? impls for xmlhttprequest IO module
- don't modify URI target (xmlhttrequest IO module)
- enforce allow-uri-read restriction for remote URI (aka URL) (xmlhttrequest IO module)
- add file:// scheme to absolute path (xmlhttrequest IO module)


mojavelinux:
@Mogztter I tested the three scenarios on Asciidoctor.js that we have so far and all is looking good:

1. file:// URI
2. Absolute URL (requires allow-uri-read attribute is set)
3. Absolute path (file:// scheme is automatically added to value)

mojavelinux:
It doesn't matter whether the include is nested or not. These always work the same.

mojavelinux:
Eventually, I can see pulling this code out into a dedicated function that lives in asciidoctor.js...but for 1.5.6.2, I'm planning to have it in core to bake.

Mogztter:
> I tested the three scenarios on Asciidoctor.js that we have so far and all is looking good:

Awesome news, I need to check it out :+1:

mojavelinux:
This will now be consolidated into #2603.

PR #2603, opened 2018-03-10
resolves #2599 and #2602 correctly resolve includes in browser environment

### LINKED ISSUE #2599: If the include target is a URL, don't modify it when the IO module is xmlhttprequest

### LINKED ISSUE #2602: Resolve relative include path correctly when IO module is xmlhttprequest
Now that the Asciidoctor.js implementation has matured, we need to revisit the logic for how a relative include path is resolved / processed in the browser environment. Here's the new proposed logic for this scenario:

* if the target is relative and this is a top-level include
   * if base_dir equals `.`, return the target unmodified
   * else if base_dir does not equal `.`, prepend the base_dir value to target
* if the target is relative and this is a nested include
   * if the `@dir` property starts with `file://`, prepend the `@dir` value to target
   * else if the `@dir` property is a URL and the allow-uri-read attribute is set, prepend the `@dir` value to target
   * else if the `@dir` property is a URL and the allow-uri-read attribute is not set, replace w/ link
   * else if the `@dir` property is not `.`, prepend the `@dir` value to target (when is this the case?)

When prepending, always separate the two values by a forward slash.

In the case of a nested include, we only have to check the allow-uri-read attribute in the case that an extension used push_include. There's no way in normal operation that the `@dir` property is a URL and the allow-uri-read attribute wouldn't have already been checked.

mojavelinux:
- add PathResolver#absolute_path? method
- isolate PathResolver#root? implementation for xmlhttprequest IO module
- add PathResolver#descends_from? helper
- initialize PathResolver to handle both both forward and backslashes in path
- always posixify target
- handle target with file:// prefix first
- don't modify target if it's a URL 
- enforce allow-uri-read restriction if target is a URL and doesn't descend from base_dir
- add file:// scheme to absolute path (xmlhttrequest IO module)
- resolve include path for top-level relative include
- resolve include path for nested relative include
- check allow-uri-read if nested include comes from a document w/ different origin as base_dir

### DIFF
--- a/lib/asciidoctor/path_resolver.rb
+++ b/lib/asciidoctor/path_resolver.rb
@@ -131,26 +131,42 @@ class PathResolver
     @_partition_path_sys, @_partition_path_web = {}, {}
   end
 
-  # Public: Check if the specified path is an absolute root path.
-  # This operation considers posix paths, windows paths, and file URLs.
+  # Public: Check whether the specified path is an absolute path.
   #
-  # Unix absolute paths and UNC paths start with slash. Windows roots can start
-  # with a drive letter. Absolute paths may start with file://, http://, or
-  # https:// when the IO module is xmlhttprequest (Opal runtime only).
+  # This operation considers both posix paths and Windows paths. It does not
+  # consider URIs.
+  #
+  # Unix absolute paths and UNC paths both start with slash. Windows roots can
+  # start with a drive letter.
   #
   # path - the String path to check
   #
   # returns a Boolean indicating whether the path is an absolute root path
-  if RUBY_ENGINE == 'opal'
+  def absolute_path? path
+    (path.start_with? SLASH) || (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
+  end
+
+  # Public: Check if the specified path is an absolute root path (or, in the
+  # browser environment, an absolute URI as well)
+  #
+  # This operation considers both posix paths and Windows paths. If the JavaScript IO
+  # module is xmlhttprequest, this operation also considers absolute URIs.
+  #
+  # Unix absolute paths and UNC paths start with slash. Windows roots can
+  # start with a drive letter. When the IO module is xmlhttprequest (Opal
+  # runtime only), an absolute (qualified) URI (starts with file://, http://,
+  # or https://) is also considered to be an absolute path.
+  #
+  # path - the String path to check
+  #
+  # returns a Boolean indicating whether the path is an absolute root path (or
+  # an absolute URI when the JavaScript IO module is xmlhttprequest)
+  if RUBY_ENGINE == 'opal' && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
     def root? path
-      (path.start_with? SLASH) ||
-          (@file_separator == BACKSLASH && (WindowsRootRx.match? path)) ||
-          (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (path.start_with? 'file://', 'http://', 'https://'))
+      (absolute_path? path) || (path.start_with? 'file://', 'http://', 'https://')
     end
   else
-    def root? path
-      (path.start_with? SLASH) || (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
-    end
+    alias root? absolute_path?
   end
 
   # Public: Determine if the path is a UNC (root) path
@@ -169,6 +185,18 @@ class PathResolver
   # returns a Boolean indicating whether the path is an absolute (root) web path
   def web_root? path
     path.start_with? SLASH
+  end
+
+  # Public: Determine whether path descends from base.
+  #
+  # If path equals base, or base is a parent of path, return true.
+  #
+  # path - The String path to check. Can be relative.
+  # base - The String base path to check against. Can be relative.
+  #
+  # returns If path descends from base, return the offset, otherwise false.
+  def descends_from? path, base
+    base == path ? 0 : ((path.start_with? base + '/') ? base.length + 1 : false)
   end
 
   # Public: Normalize path by converting any backslashes to forward slashes
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -836,10 +836,40 @@ class PreprocessorReader < Reader
         warn %(asciidoctor: ERROR: #{line_info}: maximum include depth of #{@maxdepth[:rel]} exceeded)
         return
       elsif ::RUBY_ENGINE_OPAL && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
-        # NOTE resolves uri relative to currently loaded document
-        # NOTE we defer checking if file exists and catch the 404 error if it does not
-        target_type = :file
-        inc_path = relpath = @include_stack.empty? && ::Dir.pwd == @document.base_dir ? target : %(#{@dir}/#{target})
+        # NOTE when the IO module is xmlhttprequest, the only way to check if the file exists is to catch a 404 response
+        p_target = (@path_resolver ||= PathResolver.new '\\').posixify target
+        target_type, base_dir = :file, @document.base_dir
+        if p_target.start_with? 'file://'
+          inc_path = relpath = p_target
+        elsif Helpers.uriish? p_target
+          unless (@path_resolver.descends_from? p_target, base_dir) || (@document.attributes.key? 'allow-uri-read')
+            return replace_next_line %(link:#{target}[])
+          end
+          inc_path = relpath = p_target
+        elsif @path_resolver.absolute_path? p_target
+          inc_path = relpath = %(file://#{(p_target.start_with? '/') ? '' : '/'}#{p_target})
+        elsif (ctx_dir = (top_level = @include_stack.empty?) ? base_dir : @dir) == '.'
+          # WARNING relative include won't work in the Firefox web extension (unless we use the second line)
+          inc_path = relpath = p_target
+          #inc_path = %(#{File.dirname `window.location.href`}/#{relpath = p_target})
+        elsif (ctx_dir.start_with? 'file://') || !(Helpers.uriish? ctx_dir)
+          # WARNING relative nested include won't work in the Firefox web extension if base_dir is '.'
+          inc_path = %(#{ctx_dir}/#{p_target})
+          if top_level
+            relpath = p_target
+          elsif base_dir == '.' || !(offset = @path_resolver.descends_from? inc_path, base_dir)
+            relpath = inc_path
+          else
+            relpath = inc_path.slice offset, inc_path.length
+          end
+        elsif top_level
+          inc_path = %(#{ctx_dir}/#{relpath = p_target})
+        elsif (offset = @path_resolver.descends_from? ctx_dir, base_dir) || (@document.attributes.key? 'allow-uri-read')
+          inc_path = %(#{ctx_dir}/#{p_target})
+          relpath = offset ? (inc_path.slice offset, inc_path.length) : p_target
+        else
+          return replace_next_line %(link:#{target}[])
+        end
       elsif (Helpers.uriish? target) || ((::URI === @dir) && (target = %(#{@dir}/#{target})))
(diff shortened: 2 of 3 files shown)


mojavelinux:
@Mogztter I think this is a solid start. I tested it against the asciidoctor.js test suite and it's all working. There are a few open questions, but they aren't super important.

mojavelinux:
It's kind of amazing how much logic is in there...but we're really covering a lot of scenarios. And with these identified, it will be easier to refactor this whole processor later.

mojavelinux:
Btw, this is based on #2601, so there's no need to review that one separately. I'll just squash these together.

lib/asciidoctor/reader.rb
```diff
@@ -837,10 +837,42 @@ def preprocess_include_directive raw_target, raw_attributes
...
+            replace_next_line %(link:#{target}[])
+            return true
+          end
+          inc_path = relpath = target
+        elsif @path_resolver.absolute_path? target
+          inc_path = relpath = ((target.start_with? '/') ? 'file://' : 'file:///') + target
+        elsif (ctx_dir = (top_level = @include_stack.empty?) ? base_dir : @dir) == '.'
+          # NOTE won't work in the Firefox web extension (unless we use the second line)
+          inc_path = relpath = target
+          #inc_path = (File.dirname `window.location.href`) + '/' + (relpath = target)
```
mojavelinux: @Mogztter If we want to allow the base_dir option to be unset in the browser extension, we will need to enable this line. wdyt? Is it better just to require the base_dir to be set in the browser extension?

Mogztter: I think the base_dir should be defined in the browser extension.

Mogztter:
> It's kind of amazing how much logic is in there...but we're really covering a lot of scenarios. 

Indeed, I will do a review as soon as I can

lib/asciidoctor/path_resolver.rb
```diff
@@ -131,26 +131,46 @@ def initialize file_separator = nil, working_dir = nil
...
+  def absolute_path? path
+    (path.start_with? SLASH) || (@file_separator == BACKSLASH && (WindowsRootRx.match? path))
+  end
+
   if RUBY_ENGINE == 'opal'
-    def root? path
-      (path.start_with? SLASH) ||
-          (@file_separator == BACKSLASH && (WindowsRootRx.match? path)) ||
-          (::JAVASCRIPT_IO_MODULE == 'xmlhttprequest' && (path.start_with? 'file://', 'http://', 'https://'))
+    if ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
```
Mogztter: We could use a single `if` right ? Is there a particular reason why you choose to use two `if` ?

```rb
if RUBY_ENGINE == 'opal' && ::JAVASCRIPT_IO_MODULE == 'xmlhttprequest'
  def root?
    # ...
  end
else
  alias root? absolute_path?
end
```

lib/asciidoctor/reader.rb
```diff
@@ -837,10 +837,42 @@ def preprocess_include_directive raw_target, raw_attributes
...
-        # NOTE resolves uri relative to currently loaded document
-        # NOTE we defer checking if file exists and catch the 404 error if it does not
-        target_type = :file
-        inc_path = relpath = @include_stack.empty? && ::Dir.pwd == @document.base_dir ? target : %(#{@dir}/#{target})
+        # NOTE the only way to check if the file exists is to catch a 404 response in the IO module
+        target, target_type, base_dir = ((@path_resolver ||= PathResolver.new '\\').posixify target), :file, @document.base_dir
+        if target.start_with? 'file://'
+          inc_path = relpath = target
+        elsif Helpers.uriish? target
+          # NOTE only require allow-uri-read if cross origin
```
Mogztter: Nice improvement ✨

lib/asciidoctor/reader.rb
```diff
@@ -837,10 +837,42 @@ def preprocess_include_directive raw_target, raw_attributes
...
+          unless (@path_resolver.descends_from? target, base_dir) || (@document.attributes.key? 'allow-uri-read')
+            replace_next_line %(link:#{target}[])
+            return true
+          end
+          inc_path = relpath = target
+        elsif @path_resolver.absolute_path? target
+          inc_path = relpath = ((target.start_with? '/') ? 'file://' : 'file:///') + target
+        elsif (ctx_dir = (top_level = @include_stack.empty?) ? base_dir : @dir) == '.'
```
Mogztter: In my opinion this test (+ assignment) is hard to read.
We are covering the case when the target is relative and we are on the top level include (ie. not on a nested include), that's right ?

lib/asciidoctor/reader.rb
```diff
@@ -837,10 +837,42 @@ def preprocess_include_directive raw_target, raw_attributes
...
+          # NOTE won't work in the Firefox web extension (unless we use the second line)
+          inc_path = relpath = target
+          #inc_path = (File.dirname `window.location.href`) + '/' + (relpath = target)
+        elsif (ctx_dir.start_with? 'file://') || !(Helpers.uriish? ctx_dir)
+          # WARNING a relative ctx_dir won't work in the Firefox web extension
+          inc_path = ctx_dir + '/' + target
```
Mogztter: `ctx_dir` can be relative in the Firefox web extension ? could you please elaborate ?

mojavelinux: It won't work unless you set base_dir to an absolute URI for the reason in the comment a few lines above. If we didn't want that requirement, we'd have to expand `.` to the dirname of window.location.href.

lib/asciidoctor/reader.rb
mojavelinux: This assignment allows us to handle both a top-level include and a nested include in all remaining conditions.

mojavelinux: ...and then it immediately tests the `.` condition.

lib/asciidoctor/path_resolver.rb
mojavelinux: Because the second constant is not defined in the Ruby environment and thus will fail.

mojavelinux: Nevermind, I see what you're saying. You're right.

mojavelinux: I used to have another condition in there, but then I removed it. So now we can collapse it.

mojavelinux:
What we really need before we wrap this up is a table that shows all the permutations that this block handles...for our own sanity. We shouldn't focus on stuff the block does not handle because that will just make it confusing.

lib/asciidoctor/reader.rb
mojavelinux: I tend to agree. I'll leave this comment in here in case we decide to revisit it...then we'll already know where we would put it.

mojavelinux:
The other question is whether we should delegate this logic to a method that is actually provided by Asciidoctor.js. We'd just need to come up with a little internal project-to-project API like we do with the regex stuff.

Mogztter:
> What we really need before we wrap this up is a table that shows all the permutations that this block handles...for our own sanity. We shouldn't focus on stuff the block does not handle because that will just make it confusing.

Yes! :+1: 
The code handles all the case at once (I guess for performance) and therefore it's a bit harder to read. But basically we handle absolute paths first, then relative and nested paths/includes.

> The other question is whether we should delegate this logic to a method that is actually provided by Asciidoctor.js. We'd just need to come up with a little internal project-to-project API like we do with the regex stuff.

It could be a good idea but for now it can stay here.
The main benefit would be that Asciidoctor.js could handle more JavaScript environments without any impact on Asciidoctor core.

lib/asciidoctor/reader.rb
Mogztter: I see... but I think Asciidoctor.js is best suited to expand/resolve a relative path in a browser environment

mojavelinux: I agree, that's what we're going to depend on here. I'm just noting that here so I remember what the assumption is.

mojavelinux:
> The main benefit would be that Asciidoctor.js could handle more JavaScript environments without any impact on Asciidoctor core.

Exactly. It also means the code is tested when it is changed. Right now, the test suite skips this entire branch because there's no way to test it from the Ruby side.

mojavelinux:
> So we need to update the include directive support page in Asciidoctor.js add code comments ?

I think it needs to be more detailed. That table is good for users, but we need something that looks more like a specification. I'll draft something up.

mojavelinux:
> The code handles all the case at once (I guess for performance) and therefore it's a bit harder to read.

I'd argue it's consistent with how Asciidoctor is written. It's just the style I prefer. I don't like to break the conditional chain if I don't have to. And I usually avoid checking the same condition more than once in the same flow.

Mogztter:
> Exactly.  It also means the code is tested when it is changed. Right now, the  test suite skips this entire branch because there's no way to test it  from the Ruby side.

Oooh, another solid argument :moyai: 

> I think it needs to be more detailed. That table is good for users, but we need something that looks more like a specification. I'll draft something up.

Excellent!

> I'd argue it's consistent with how Asciidoctor is written. It's just the style I prefer. I don't like to break the conditional chain if I don't have to. And I usually avoid checking the same condition more than once in the same flow.

Fair enough :wink:

mojavelinux:
As per the spec, here's what I had in mind.

https://gist.github.com/mojavelinux/9a8fab76a0833eaa2c65914b412a1241

I put it in the gist so that I can update and so that we can eventually move it to live somewhere in the Asciidoctor.js project. Feel free to snag it whenever you want.

mojavelinux:
I'm prepared to merge at this point. I haven't yet had a chance to test the previously broken scenarios in the web extension yet.

mojavelinux:
> I'm prepared to merge at this point.

Once we merge, and before we release, we can decide whether we want to move the code to Asciidoctor.js and replace it with a stub. But I want it to be in the history of asciidoctor core for auditing purposes.

mojavelinux:
I tested with the web extension and everything seems to work fine. You will need to set the base_dir option to match the pseudo-code `path.dirname(window.location.href)`. I see you're already extracting the filename, so it's just a matter of getting the dirname too.

mojavelinux:
I'm also thinking about one other possible to change to the v1.5.6.x branch, which is a partial rollback from what we added. I had already allowed the `docdir` attribute to be passed through unmodified. Until we have full support for using a URL as the base_dir, I wonder if we should set the `docdir` attribute in the web extension instead of the base_dir and remove the exception from the base_dir processing.

Here's the code I'm referring to:

https://github.com/asciidoctor/asciidoctor/blob/8b6a5046642f390b696af21aa5eb991536cb7371/lib/asciidoctor/document.rb#L389-L400

As you can see, if base_dir is not set, but docdir is, docdir is used unmodified (even in the Ruby version). So we can remove the special case for the xmlhttprequest IO module in the first condition and just set the docdir for now. wdyt?

mojavelinux:
Setting the docdir without setting the base_dir is intended only for advanced use cases (of which this is the case). (It's also used in Antora, which uses a virtual filesystem).

Mogztter:
> Setting the docdir without setting the base_dir is intended only for advanced use cases (of which this is the case). (It's also used in Antora, which uses a virtual filesystem).

Then I think it's fine to remove the special condition and define the `docdir` in the browser extension :+1:

mojavelinux:
Great. Let's proceed with that plan until we have complete support for a URI value.

mojavelinux:
I'm happy with where we are now. I'm going to file another issue to discuss the API.

mojavelinux:
See #2610.

Mogztter:
The updated "include directive matrix" is looking good: https://github.com/asciidoctor/asciidoctor.js/pull/384/files?short_path=9426fa4#diff-9426fa44d3cde6c36db133748626a9b2

:ok_hand: :ok_hand: :ok_hand:

mojavelinux:
:+1:

PR #2612, opened 2018-03-16
resolves #1133 show error if delimited block is not terminated

### LINKED ISSUE #1133: code blocks with mismatched start/end lines never end
Given an input like:

```
-------
some preformatted content
--------

some regular content
```

where the start and end lines are not identical (the first has seven hyphens, while the latter has eight), we don't end the block at the ending line (and `some regular content` is part of the preformatted block). AFAIK, the behavior here is not explicitly documented by AsciiDoc, but regular AsciiDoc does end the block.

Git's documentation has run into this. It renders fine using AsciiDoc, but not on the git-scm.com site, which uses AsciiDoctor. A specific example is in https://github.com/git/git-scm.com/issues/447.

mojavelinux:


mojavelinux:
OPEN QUESTION: should we warn if a comment block is open? In my experience, I tend to use unterminated comment blocks to comment out the rest of the document, so leaving them open seems like a valid use case.

mojavelinux:
> should we warn if a comment block is open?

We've decided the answer to this question is yes.

PR #2613, opened 2018-03-16
resolves #2610 extract method to resolve include path

### LINKED ISSUE #2610: Delegate include target resolution to method that can be overridden in Asciidoctor.js
As a follow-up to #2599 and #2601, delegate the logic of resolving the include target to a method provided by Asciidoctor.js. Doing so allows the implementation to be managed in the Asciidoctor.js code base, thus allowing any changes to be tested immediately.

Here's the proposed API, which will be a member method on the PreprocessorReader class.

```rb
# An internal contract with Asciidoctor.js to resolve the target of an include when
# using the xmlhttprequest IO module. It must return a tuple containing the resolved
# (absolute) path and the path relative to the root document. Alternately, it can return
# a boolean to halt processing, which is used as the return value of the calling method.
def resolve_include_target_xmlhttprequest target
  ...
end
```

**OPEN QUESTION:** Should we use the term `browser` instead of `xmlhttprequest`? (i.e., `resolve_include_target_browser`)

The method can assume that attribute references in the target value have already been resolved.

The method is expected to return a tuple with the following values:

* inc_path - the resolved include path, which will later be passed to the top-level open function.
* relpath - the relative path of the target starting from the directory of the root document (i.e., root-relative)

If the method wants to halt processing of the include (for example, after replacing it with a link), it should instead return a boolean value.

mojavelinux:


mojavelinux:
@Mogztter I think this one's ready to go. I was going back and forth about whether to pass both the attrlist (raw string) and attributes (parsed attributes), but from what I can see we need both values.

I'm mostly interested in whether you think we got the method signature right, since this is the primary contract between the two projects.

PR #2623, opened 2018-03-22
resolves #2622 raise clearer exception when extension class cannot be resolved

### LINKED ISSUE #2622: Report clearer error message if extension class cannot be resolved from string
Under certain circumstances, when a string is passed to an extension registration method, and the extension class cannot be resolved, Asciidoctor allows a low-level Ruby error message to bubble up.

Specifically, if the value is not a valid constant name (e.g., the value starts with a lowercase letter), the exception is not properly caught and, as a result, the user sees a cryptic message.

```rb
  block 'foobar'
end
```

In this case, since "foobar" begins with a lowercase, the user sees:

```
NameError: wrong constant name foobar
```

This exception should be caught so that the user sees this message instead:

```
Could not resolve class for name: foobar
```

Additionally, the resolver method is finding classes in the top-level scope even if specified with a specific scope:

```rb
class CustomBlockMacro < Asciidoctor::Extensions::BlockMacroProcessor
  use_dsl
  named :foobar
  def process parent, target
    puts %(processing #{target} macro...)
  end
end

  block_macro 'Asciidoctor::Extensions::CustomBlockMacro'
end
```

The CustomBlockMacro extension should not be resolved.

mojavelinux:
- split impl of class_for_name by Ruby version
- add tests to verify correct error message is thrown when passed an illegal constant name
- add tests to verify top-level class is not resolved when namespaced class is requested

### DIFF
--- a/lib/asciidoctor/extensions.rb
+++ b/lib/asciidoctor/extensions.rb
@@ -1481,17 +1481,36 @@ module Extensions
     # Public: Resolves the Class object for the qualified name.
     #
     # Returns Class
-    def class_for_name qualified_name
-      resolved = ::Object
-      (qualified_name.split '::').each do |name|
-        unless name.empty? || ((resolved.const_defined? name) && ::Module === (resolved = resolved.const_get name))
-          raise ::NameError, %(Could not resolve class for name: #{qualified_name})
-        end
+    if RUBY_MIN_VERSION_2
+      def class_for_name qualified_name
+        resolved = ::Object.const_get qualified_name, false
+        raise unless ::Class === resolved
+        resolved
+      rescue
+        raise ::NameError, %(Could not resolve class for name: #{qualified_name})
       end
-      raise ::NameError, %(Could not resolve class for name: #{qualified_name}) unless ::Class === resolved
-      resolved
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/extensions.rb
```diff
@@ -1481,17 +1481,36 @@ def resolve_class object
...
-          raise ::NameError, %(Could not resolve class for name: #{qualified_name})
+    if RUBY_MIN_VERSION_2
+      def class_for_name qualified_name
+        resolved = ::Object.const_get qualified_name, false
+        raise unless ::Class === resolved
+        resolved
+      rescue
+        raise ::NameError, %(Could not resolve class for name: #{qualified_name})
+      end
+    elsif RUBY_MIN_VERSION_1_9
```
Mogztter: Ruby 1.9 is not EOL yet ?

mojavelinux: Yes. But we need a dedicated release and announcement before we can drop support. Until then, I have to keep the test suite running as is.

mojavelinux:
I know, I was so surprised to find out this is finally built-in as of Ruby 2...and that Opal supports it too.

PR #2636, opened 2018-03-28
Add source and manual attrs for manpage

tiwai:
Define source and manual refmiscinfo entries if the corresponding
attributes are defined.  For source, we need to define a dumb one
always; otherwise xmlto puts ugly "FIXME".


mojavelinux:
Thanks! This is a great addition!

PR #2667, opened 2018-04-14
Allow the DATA_PATH variable to be picked from an environment variable

aerostitch:
Hi guys,

In the Debian packaging process we maintain a patch to allow the DATA_PATH variable to be set from another location as the location we have during the debian package build process (and in the final installation) are different from an usual gem context.

The current patch in Debian is: https://salsa.debian.org/ruby-team/asciidoctor/blob/master/debian/patches/data_path.patch

I'd like to avoid patching from within the package so I thought it could be more flexible to allow a variable (named here `ASCIIDOCTOR_DATA_PATH`) to be set to define the path where the `data` folder is located.

Would that be an acceptable solution?

Thanks for your help,
Joseph

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -196,7 +196,7 @@ module Asciidoctor
   #LIB_PATH = ::File.join ROOT_PATH, 'lib'
 
   # The absolute data path of the Asciidoctor RubyGem
-  DATA_PATH = ::File.join ROOT_PATH, 'data'
+  DATA_PATH = ::ENV['ASCIIDOCTOR_DATA_PATH'] || (::File.join ROOT_PATH, 'data')
 
   # The user's home directory, as best we can determine it
   # NOTE not using infix rescue for performance reasons, see: https://github.com/jruby/jruby/issues/1816


mojavelinux:
Seems reasonable to me.

aerostitch:
Sorry I pushed too fast and got distracted.
This will work for the build process but once installed on someone's laptop the environment variable won't be set anymore and the subfolders of data will be installed inside `/usr/share/ruby-asciidoctor/`.
I'm not sure yet how to manage that use case.

mojavelinux:
In that case, my advice would be not to move the data folder when packaging. For all intents and purposes, this is source code, not shared data. The reason it's a file and not a string in the source code is simply to make it easier to manage.

aerostitch:
having a data folder inside `/usr/bin` will break the Filesystem hierarchy standards that we have to respect in Debian. The architecture-independent data have to be in `/usr/share` and in a folder named after the package it belongs to.
If we don't do that the package will be rejected.

More about FHS in linux in general: http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html 
Debian summary of it: https://wiki.debian.org/FilesystemHierarchyStandard

mojavelinux:
> having a data folder inside /usr/bin will break the Filesystem hierarchy standards that we have to respect in Debian.

I'm not suggesting it should be in /usr/bin. I'm suggesting it should be where ever lib/asciidoctor.rb ends up.

Debian can't expect that if it changes the structure of a package that the package will continue to work properly. Debian is violating the packaging contract that gem authors have defined.

mojavelinux:
> The architecture-independent data have to be in /usr/share and in a folder named after the package it belongs to.

The fact that Debian is imposing a rule on gem authors needs to be rethought. This software is not owned by Debian and therefore it cannot mandate how a library author creates a package when the package structure is already well defined. I strongly oppose Debian's position here.

aerostitch:
That's fine, don't worry about it.
I know the rules imposed by debian are sometimes generating conflicts with upstream and I know it's sometimes better to keep the patches in the package. I thought it would have fixed my issue in a smart way when I opened the PR but I opened it too soon and realized only afterwards that it wouldn't work, so closing that one. Sorry for the noise.

mojavelinux:
I appreciate you raising the issue and the discussion about it. I know you don't make the packaging rules. I just think it's important to point out that Debian is overstepping here. Hopefully that message can be passed on.

aerostitch:
Lol!
Those kind of disagreements have been around for a long time, believe me. Some thinks hard like stone that the only way to avoid chaos is to stick to the FHS recommendations that have been setup for linux almost 25 years ago.
It's part of my role to make sure the package we ship in the distro works when bent to those rules, so here I am, annoying the upstream projects trying to find solutions to shorten the amount of stuffs I have to do on each release! :wink:

mojavelinux:
I certainly appreciate your efforts. Hopefully that spirit doesn't get lost in our little debates ;)

aerostitch:
Lol!
Don't worry, I love to annoy people, especially when they can answer! :smile:

elextr:
>  This software is not owned by Debian and therefore it cannot mandate how a library author creates a package when the package structure is already well defined.

And the world has moved on from 1970s Linux, maybe the standards should too.  I know at one stage Asciidoc Python was not allowed to put asciidoc (actually asciidoc.py without the extension) in bin "because its not binary".  Clearly the existance of interpreters and the fact that exec interprets shebangs hadn't impinged on that commenter.  I am not sure what the result of this was, maybe @aerostitch knows.

aerostitch:
I've been managing the asciidoc debian package for several years now, so yes i do know. :smile: 
But Debian has evolved a lot and having interpreted scripts without extension in `/usr/bin` is common now provided that you have the proper shebang in place.

As a system administrator & DBA, I won't deny that having a norm to follow when you are looking for a certain kind of file gives you a certain comfort, but some people are more strict that others.
I found out that those are generally the ones that have a ton of different packages to manage, so having a uniform way to manage the files provides them more efficiency especially for packages they don't update often (and some old habits of rules they've been using for 10 to 20 years take time to change).
And sometimes that generates bad relationships with other people and package developers unfortunately, giving bad press to the rest of the project. :disappointed:

mojavelinux:
> Don't worry, I love to annoy people, especially when they can answer!

I see it as a challenge, and I like challenges. It makes me think harder about it. In fact, sometimes, it plants a seed that leads to a change much further down the line. That's why it's good to talk through it.

mojavelinux:
> And sometimes that generates bad relationships with other people and package developers unfortunately

The way I see it, the only bad relationship is no relationship at all. We are (passionately) sharing our perspectives, and that leads to understanding. The better I can understand the constraints / standards, the more consideration I can give them when development. Just because I ask to make it easier on me doesn't mean I always get my way (just as the reverse is true). But we can find that common ground, do some compromising, and we'll get a better package in the end. That's winning.

PR #2668, opened 2018-04-15
Allow the tests to run when asciimath is not installed

aerostitch:
Hi,

As asciimath is an optional gem (and not packaged in Debian yet), we maintain a patch that skips the tests that fail when asciimath is not installed: https://salsa.debian.org/ruby-team/asciidoctor/blob/master/debian/patches/skip-asciimath-test.patch
But if you only do the assert when the warning message about the gem is not present we would be able to avoid this.
Does that seem reasonable?

Thanks
Joseph

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -116,6 +116,7 @@ Improvements / Refactoring::
   * only warn if thread_safe gem is missing when using built-in template cache
   * rename enumerate_section to assign_numeral; update API docs
   * drop deprecated compact option from CLI; remove from manpage
+  * use more robust mechanism for lazy loading the asciimath gem
   * purge render method from test suite (except to verify alias)
 
 Documentation::
--- a/test/blocks_test.rb
+++ b/test/blocks_test.rb
@@ -1525,7 +1525,8 @@ sqrt(3x-1)+(1+x)^2 < y
       assert_equal '\$sqrt(3x-1)+(1+x)^2 &lt; y\$', nodes.first.to_s.strip
     end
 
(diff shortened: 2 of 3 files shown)


mojavelinux:
Is there an environment variable we can check that indicates that we're in the packaging process?

mojavelinux:
What I'm concerned about is this could weaken the test. It would be better to know what environment we are running in so that we know what we need to assert.

I'll have a closer look once I get the logging PR merged. That will necessitate rebasing this patch anyway as now we will look in the logger for the message instead of the stderr stream.

aerostitch:
Sounds good.
My assumption was, as the gem is optional anyway, if one try to run the tests in an environment that don't have the optional gems, we could make the tests run anyway.

Basically the unless statement I'm adding here tests that the gem works unless the proper warning is issued.

mojavelinux:
I have an idea, but I need to experiment. Details to come.

mojavelinux:
Good news. The experiment panned out. Update on the way.

mojavelinux:
Okay, let me know if this works out for you.

I did a few things. First, I test if the gem is available before starting the test. That determines how to assert. But we assert either way, so it's just as robust.

I also added a test that fakes the gem not being available so we can test what happens even when it is.

Now it should be possible to accomplish your goal. I'll be interested to hear if it works out.

aerostitch:
The logic looks good to me, the code looks great, the test work in with and without asciimath installed.
Looks like it's good to go.

mojavelinux:
Fantastique !

PR #2672, opened 2018-04-16
resolves #2433 warn if nested sections are used where not allowed

### LINKED ISSUE #2433: Only preface, appendix, and abstract special sections are allowed to have subsections
When converting to DocBook, only the preface, appendix, and abstract special sections are permitted to have subsections. If subsections are added to other special sections when converting to DocBook, Asciidoctor should emit a warning (or error).

mojavelinux:
- optimize section level validation
- warn if subsections are used where not allowed
- add tests to verify warnings when section title is out of sequence

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -88,6 +88,7 @@ Bug fixes::
 Improvements / Refactoring::
 
   * log a warning message if an unterminated delimited block is detected (#1133, PR #2612)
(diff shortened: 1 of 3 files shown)


elextr:
Just noticed that the first example [here](https://asciidoctor.org/docs/user-manual/#user-preface) jumps from level 1 to level 3.

mojavelinux:
That's actually correct. Subsections of a special section always start at level 2. A special section in a book can be written as a level 0 section to make it a sibling of a part. But it gets pushed down to level 1 when processed.

PR #2677, opened 2018-04-16
Fix a few typos in extensions.rb

vivien:
This fixes the ChromeInineMacro -> ChromeInlineMacro and missing `do` keyword in the extensions.rb file. Only comments are affected, no functional changes.


mojavelinux:
Thanks @vivien!

PR #2683, opened 2018-04-18
resolves #2369 change strategy for finding tags inside circumfix comments

### LINKED ISSUE #2369: Potential regression on tag inclusion in 1.5.6.1

mojavelinux:
- allow any content after a tag directive if offset by a space
- add test to verify tag is found inside HTML inside a string in a source file

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -54,7 +54,8 @@ Enhancements::
 Bug fixes::
 
   * set :to_dir option correctly when output file is specified (#2382)
-  * using the shorthand syntax to set block attributes (id, roles, options) doesn't reset block style (#2174)
+  * using the shorthand syntax on the style to set block attributes (id, roles, options) no longer resets block style (#2174)
+  * match include tags anywhere on line as long as offset by word boundary on left and space or newline on right (#2369, PR #2683)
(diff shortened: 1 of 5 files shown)


mojavelinux:
cc: @cexbrayat we're dropping the strategy of looking at the file extension (except to determine it's not an AsciiDoc file) and just matching circumfix comment pairs. This will allow using any supported circumfix pair in any source file, which will address the case of XML inside of a string in another language.

PR #2701, opened 2018-04-22
Accurately track source location for lists, list items, and tables

mojavelinux:
- fix incorrect line number on lists and tables (@seikichi)
- populate source_location on list items if sourcemap is enabled (@Mogztter)
- consolidate cursor data in Parser.next_block into an array
- consolidate definition of Cursor class
- split Reader#cursor and Reader#cursor_at methods
- introduce a mark system for tracking line numbers more accurately
- add Reader#mark method to mark position of cursor (store internally)
- add Reader#cursor_at_mark method to build Cursor from previously saved mark
- mark position of cursor after reading block attribute lines in Parser.next_block
- use Reader#cursor_at_mark to retrieve/record source location in Parser.next_block
- add cursor option to Reader#read_lines_until so unterminated block warning reports start location of block
- allow value of cursor option passed to Reader#read_lines_until to be :at_mark to read cursor from mark
- mark and save cursor data when creating parsing context for table
- use cursor at mark when creating cells
- rename cursor_at to cursor_at_line
- rename prev_line_cursor to cursor_at_prev_line

This PR supersedes #2547 and #2069. The changes from those PRs have been incorporated into this one.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -95,6 +95,8 @@ Bug fixes::
 
 Improvements / Refactoring::
 
+  * use cursor marks to track lines more accurately; record cursor at the start of each block, list item, or table cell (PR #2701, PR #2547) (@seikichi)
+  * populate source_location property on list items when sourcemap option is set on document (PR #2069) (@mogztter)
   * log a warning message if an unterminated delimited block is detected (#1133, PR #2612)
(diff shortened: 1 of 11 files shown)


mojavelinux:
I've updated this substantially to handle even more cases. Now lines are reported accurately even for some pretty crazy combination of blocks.

PR #2706, opened 2018-04-24
resolves #2381 allow xrefstyle to be overridden per xref using the xrefstyle attribute

### LINKED ISSUE #2381: Inline xref macro should support attributes
Like the link macro, the xref macro should support attributes. Among other uses, this would allow a role to be specified on the link created for the xref. It will also be needed to allow the style of the xref to be controlled for a specific xref.

mojavelinux:
- allow xrefstyle to be overridden per xref by assigning xrefstyle attribute on xref macro
- allow xrefstyle to be overridden in different parts of the document

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,13 +18,14 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 Enhancements::
 
   * BREAKING: drop XML tags, character refs, and non-word characters (except hyphen and space) when generating ID for section (#794)
-  * parse text of xref macro as attributes if attribute signature found (equal sign) (#2381)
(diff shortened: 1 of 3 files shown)


cirosantilli:
Awesome! I could not find it documented in the user guide, would be good to add a mention at https://asciidoctor.org/docs/user-manual/#customizing-the-cross-reference-text if not yet done!

mojavelinux:
If you'd like to see it in the documentation, feel free to submit a PR to add it to this file: https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/xrefstyle.adoc

cirosantilli:
site build failed with:

```
You must use Bundler 2 or greater with this lockfile.
Run `rake update` to install additional required gems from RubyGems.
```

 `rake update` failed, and I don't have the time to figure it out now :-)

PR #2707, opened 2018-04-24
resolves #2059 enable linkattrs behavior implicitly (except in compat mode)

### LINKED ISSUE #2059: Enable attributes in the link macro by default (set linkattrs attribute)
It's strange that the link macro behaves differently than other macros. It only accepts attributes if the `linkattrs` document attribute is set, which is only a feature of Asciidoctor. The link macros should have attributes on by default.

We now have smart parsing of the attribute text so as not to catch cases when the alt text contains a comma (a common scenario). Since we have this parsing, we should be able to safely enable this feature.

One compromise is to enable link attributes by default when the link prefix is used, but not for raw URLs. Personally, I think this would just create more confusion than it's worth, but we can discuss.

This feature can remained disabled by default when compat mode is set.

mojavelinux:
- parse text of link macro as attributes if text contains an equal sign
- parse text of mailto macro if text contains a comma
- remove use of linkattrs attributes from test suite

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -804,11 +804,10 @@ module Substitutors
         attrs, link_opts = nil, { :type => :link }
         unless text.empty?
           text = text.gsub ESC_R_SB, R_SB if text.include? R_SB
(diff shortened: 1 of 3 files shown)


mojavelinux:
This change did not introduce a single failure in the test suite, so I'm confident it is very benign.

PR #2719, opened 2018-04-25
Translate Copyright and Licensing, Authors, and Changelog in README into Japanese

wotsushi:
I've tried translating the sections in README into Japanese.


mojavelinux:
Thank you for your contribution, @wotsushi!

PR #2747, opened 2018-05-02
make PathResolver instance available via path_resolver property on Document

mojavelinux:


lib/asciidoctor/document.rb
```diff
@@ -218,6 +218,9 @@ def to_s
   # Public: Get the Reader associated with this document
   attr_reader :reader
 
+  # Public: Get/Set the PathResolver instance used to resolve paths in this Document.
+  attr_reader :path_resolver
```
Mogztter: Should we expose this attribute in the Asciidoctor.js API ?

mojavelinux: I'm not sure yet. I'm putting it here as a test. It's primarily there so that the inner document can access it. But we might decide later it's useful to make public-public.

PR #2748, opened 2018-05-02
resolves #2108 use Pathname to resolve relative path for file outside base directory

### LINKED ISSUE #2108: xrefs vs files included from different directories does not detect internal/external xrefs correctly
Let's say you include `../doc-b/doc-b.adoc` from `/tmp/doc-a/doc-a.adoc` and then try to xref `doc-b` from `doc-a`. Since `doc-b` is included, it should automatically detect it and make only internal link rather than external. The problem is how `document.references[:includes]` is built -- it does not contain full paths, so when we check this in `substitutors.sub_inline_xrefs`, it does not always work:
```@document.attributes['docname'] == path || @document.references[:includes].include?(path)```
...because `path` can be pointing to the same file but can be relative from somewhere else.

The problem with relative paths is addressed in #2107 but is not enough alone.

I did an experiment modifying the test as follows:
```
require 'pathname'
includes_full = @document.references[:includes].map! {|x| (Pathname.new x).cleanpath.to_s}
path_full = (Pathname.new path).cleanpath.to_s
if @document.attributes['docname'] == path || @document.references[:includes].include?(path) || includes_full.include?(path_full)
```
...what worked but I'd rather see a method like `@document.include?(path)` that would do this on the background -- I am not sure if we can start storing full paths in `@document.references[:includes]` as it is used for some other purposes but we should be able to have at least another set like `:includes_full` or store something like tuple (relative and absolute paths) or so.

mojavelinux:
- autoload pathname
- use Pathname#relative_path_from to resolve path outside base directory
- add test

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -61,6 +61,7 @@ Fixes::
   * don't eagerly apply subs to inline attributes in general
   * make sure encoding of output file is UTF-8
   * prevent warning about invalid `:asciidoc` option when using custom templates with Slim 4 (#2928)
+  * use Pathname#relative_path_from to compute relative path to file outside of base directory (#2108)
 
 Improvements::
 
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -15,6 +15,7 @@ else
   autoload :Base64, 'base64'
   autoload :URI, 'uri'
   autoload :OpenURI, 'open-uri'
+  autoload :Pathname, 'pathname'
   autoload :StringScanner, 'strscan'
 end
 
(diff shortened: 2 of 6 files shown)


PauloFrancaLacerda:
I'm having problems in v1.5.8 because of `relative_path_from`.
The same docs that used to compile fine with v1.5.7 now throw this error:

**`Failed to load AsciiDoc document - different prefix: "O:/" and "B:/books/the-adoc-file"`**

NOTE:
"O:/" is the drive of the Asciidoctor root:
       `:asciidoctor-root: O:/Progs/Asciidoc`

Obviously, the relative path cannot be determined in this case, since the paths involved are in different drives.

PR #2758, opened 2018-05-06
Allow to specify the lib path via environment variable for the tests or load the gem directly

aerostitch:
Hi,

Context: I am packaging the new version of asciidoctor for Debian and am trying to reduce the amount of custom stuffs we do there.

The idea of this PR is to be able to specify the tests when running them against an already installed environment.
With this PR we could either:
* use what we currently do if `ASCIIDOCTOR_LIB_PATH` is not set and the usual directory exists
* specify the `ASCIIDOCTOR_LIB_PATH` environment variable to say that the `lib/asciidoc.rb` is located in a given build directory (or in `/usr/lib/ruby/vendor_ruby/asciidoctor.rb` for example)
* try to load the gem if the usual location does not exist

Up to now we were changing the following line to `require 'asciidoctor'`:
```ruby
require File.join(ASCIIDOCTOR_PROJECT_DIR, 'lib', 'asciidoctor')
```
Having the environment variable would give us even more flexibility.

Would that be an acceptable change?

Thanks
Joseph

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,6 +1,7 @@
 # encoding: UTF-8
 ASCIIDOCTOR_TEST_DIR = File.expand_path File.dirname __FILE__
 ASCIIDOCTOR_PROJECT_DIR = File.dirname ASCIIDOCTOR_TEST_DIR
+ASCIIDOCTOR_LIB_PATH = ENV['ASCIIDOCTOR_LIB_PATH'] || File.join(ASCIIDOCTOR_PROJECT_DIR, 'lib', 'asciidoctor')
 Dir.chdir ASCIIDOCTOR_PROJECT_DIR
 
 if RUBY_VERSION < '1.9'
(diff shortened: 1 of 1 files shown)


mojavelinux:
I'm totally fine with making this change as it's very much in the spirit of what the environment variables in test_helper.rb aim to do. However, I have two questions:

* Can we set the variable to the lib directory instead? i.e., ASCIIDOCTOR_LIB_DIR
* Is it necessary to check if the file exists? Or can we simply assume that if the environment variable is set, then we should use it?

aerostitch:
Hi @mojavelinux 

> Can we set the variable to the lib directory instead? i.e., ASCIIDOCTOR_LIB_DIR

Yep, no problem, PR updated accordingly.

> Is it necessary to check if the file exists? Or can we simply assume that if the environment variable is set, then we should use it?

That was just to provide the ability for a person to put a dummy value in the variable and fallbacl on a `require 'asciidoctor'` but it is true that the value is quite limited or nil. PR updated to reflect that.

Thanks for the review
Joseph

mojavelinux:
Thanks Joseph!

mojavelinux:
Let me know how that works out. If you need other changes to make it work, don't hesitate to propose them.

PR #2760, opened 2018-05-06
Allow the usage of ASCIIDOCTOR_LIB_DIR in gemspec and Rakefile for loading the version

aerostitch:
Hi,

This is the same idea as #2758.
Context: if you are in a build/autopkgtest environment that don't have the exact same directory structure as this, you end up having to write patch like https://salsa.debian.org/ruby-team/asciidoctor/blob/c7e30fd0010a0c6f90df492695e6d9731c0b2ce9/debian/patches/package-version.patch to change the lib directory.
This patch would allow us to use the environment variable `ASCIIDOCTOR_LIB_DIR` to specify the lib directory in those specific use cases.

Thanks
Joseph

### DIFF
--- a/asciidoctor.gemspec
+++ b/asciidoctor.gemspec
@@ -1,5 +1,10 @@
 # encoding: UTF-8
-require File.expand_path '../lib/asciidoctor/version', __FILE__
+begin
+  require File.expand_path '../lib/asciidoctor/version', __FILE__
+rescue LoadError
+  require 'asciidoctor/version'
+end
+
 require 'open3' unless defined? Open3
 
 Gem::Specification.new do |s|


aerostitch:
No, I've just tested it. The problem is that we have the following:
* In the build environment it works without any issue because we have the full package
* in the autopkgtest, the tests are done against the packaged fully installed and in the directory you're in, you only have the Rakefile, gemspec and a few other files necessary for the tooling to load what it needs, so the file `lib/asciidoctor/version` doesn't exist there.

The trick that was done with the patch we have (I just realized it) is that it adds to the LOAD_PATH even if the directory does not exist and then it does the require which will get the library needed from another directory in the LOAD_PATH (the system library path).

mojavelinux:
Here's the funny thing. We don't even need to load the version file in the rake build. The task it was used for isn't used anymore, so I can just gut it.

That leaves us with the gemspec. But there we could attempt to load it from the load path and fallback to the relative require...or even just fallback to version 0.0.0.

aerostitch:
That's a good idea, give me a minute I should be able to do just that. I'll update the PR once I've tested (maybe 30mins or so).
Thanks for the idea! :)

mojavelinux:
I dropped the obsolete stuff from the Rakefile.

aerostitch:
Awesome thanks, I'll rebase my patch and test.

asciidoctor.gemspec
```diff
@@ -1,5 +1,10 @@
 # encoding: UTF-8
-require File.expand_path '../lib/asciidoctor/version', __FILE__
+begin
+  require 'asciidoctor/version'
+rescue LoadError
+  require_relative 'lib/asciidoctor/version'
```
mojavelinux: We still need to use `require File.expand_path '../lib/asciidoctor/version', __FILE__` to keep the 1.8.7 build running (though we'll be dropping that support shortly).

aerostitch: Ok, cool. I amended the commit. Thanks! :)

mojavelinux:
Fantastic!

mojavelinux:
One minor request. Can we swap the order of the two require statements?

aerostitch:
Yeah that's what I was about to say, if you build in an environment that already has asciidoctor installed, that will take the wrong version.

aerostitch:
Commit amended. Time to go to bed for me. Talk to you later! ;)

mojavelinux:
Thanks Joseph!

aerostitch:
Thank you! :)

PR #2763, opened 2018-05-07
resolves #2738 use part-signifier and chapter-signifer, if set

### LINKED ISSUE #2738: Use the chapter-label and part-label attributes, if set
If the chapter-label and part-label attributes are set, and the corresponding numbering attributes are set (sectnums and partnums), prepend the value of these attributes to the chapter and part title, respectively, in the HTML output. This change should be applied in both the main content area and the table of contents.

Example:

```
= Book Title
:doctype: book
:sectnums:
:partnums:
:chapter-label: Chapter
:part-label: Part

= First Part

== First Chapter
```

Should produce:

```
Part I. First Part

Chapter 1. First Chapter
```

mojavelinux:
- add value of part-signifier attribute to beginning of title for numbered part
- add value of chapter-signifier attribute to beginning of title for numbered chapter
- change trailing delimiter of part number from period to colon

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 Enhancements::
 
+  * if set, add value of part-signifier and chapter-signifier attributes to part and chapter titles (#2738)
   * allow position (float) and alignment (align) to be set on video block (#2425)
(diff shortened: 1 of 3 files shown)


mojavelinux:
Let's switch to part-sig (or part-signifier) and chapter-sig (or chapter-signifier).

PR #2774, opened 2018-05-10
resolve #1860 multiple revision inserted in docbook format.

### LINKED ISSUE #1860: Multiple RevisionInfo entries are processed as content
Per the [Asciidoc Document Structure Specification](http://www.methods.co.nz/asciidoc/chunked/ch08.html), the header may contain zero or more `RevisionInfo` blocks, each containing an optional revision number, a required revision date, and an optional revision remark.

When creating an asciidoc with a revision history (as is quite common in DocBook land) and rendering with asciidoctor, the first entry is parsed correctly with the version and the date, and the remaining entries get lumped into the content after the title page with no additional formatting. They are smashed together in a single paragraph block.

This additionally causes my `:toc:` attribute to be missed (and other attributes too).

I would expect that asciidoctor would parse all RevisionInfo lines, select the numerically largest revision (which is not necessarily the first) for the rendered Title Page/Block, and make the rest available to the rendering backend for a revision table, if desired.

miko53:
Hello, 

This pull request inserts a revision history section with several revisions in a docbook output.
To be able to manage several authors, I put also a author section inside revision line.
This pach keeps compatibility with previous format.

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -437,7 +437,8 @@ module Asciidoctor
     #   v1.0, 2013-01-01: Ring in the new year release
     #   1.0, Jan 01, 2013
     #
-    RevisionInfoLineRx = /^(?:\D*(.*?),)? *(?!:)(.*?)(?: *(?!^),?: *(.*))?$/
+    #RevisionInfoLineRx = /^(?:\D*(.*?),)? *(?!:)(.*?)(?: *(?!^),?: *(.*))?$/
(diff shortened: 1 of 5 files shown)


mojavelinux:
I like where this is headed. We'll aim to integrate this in 2.0.0.

I haven't decided whether I like the author as another field. Perhaps the author could be bracketed, like:

```
v0.9.0, 2016-06-03: (John Doe) We're getting there
```

It would also have to be clear that the author won't be parsed in this case.

miko53:
yes that sounds good in this format. The fields author can be ignored.
The drawback is in pdf (converted with dblatex) by default the name is displayed in revision history. How can we manage this ?

PR #2778, opened 2018-05-12
should not set option on node if already set

mojavelinux:


mojavelinux:
For now, we're going to rely on the presence of the `<name>-option` attribute to check whether the option is already set. In the future, options will be stored in a set, so we'll revisit this assumption then.

Issue #242, opened 2013-04-16
Create a new HTML5 converter that maximizes semantic constructs


mojavelinux:
The HTML 5 backend in AsciiDoc is acceptable, but certainly not ideal. It's kind of a stretch to call it HTML 5. We should create an alternative HTML 5 backend that maximizes the use of semantic constructs and keeps the markup to a minimum.

I can cite a few examples of where improvements could be made.
### Paragraphs

I don't see a reason for having paragraphs wrapped in a `<div>`.

```
<div class="paragraph">
  <p>text</p>
</div>
```

This breaks a lot of CSS frameworks, which assume the `<p>` tag will be adjacent to other elements.

It should be handled instead by always using `<p>` tag as the outer element:

```
<p class="paragraph">text</p>
```

and using a nested `<span>` tag for the title of a format paragraph:

```
<p class="paragraph"><strong class="head">Title</strong> text</p>
```

(This is the same approach as used in the [Asciidoctor EPUB3 converter](https://github.com/asciidoctor/asciidoctor-epub3/blob/c313448ecaab2dc0f4f51406226b214058b5bb2d/lib/asciidoctor-epub3/converter.rb#L253-L265)).

### Admonition blocks

Admonition blocks are rendered as a table, which makes styling them very difficult. They should be done instead as layered `<div>` tags that can accommodate a wider range of layouts.
### Sidebars

`<aside>` element. Nuff said.
### Header and footer

`<header>` and `<footer>` elements. Nuff said.
### Sections

`<section>` element. Nuff said.
### Block images

`<figure> and`<figcaption>` element. Nuff said.
### Content

`<article>` element if doctype is article, or perhaps the `<main>` element.
### Dates and times

`<time>` element. Nuff said.

Shadowfiend:
What's the status here, out of curiosity? Is there something I can do to help? The current asciidoctor output is rather painful (if fully functional :)), and I'd love to get a proper HTML5 conversion off the ground quickly. It's not really clear from this issue (and the ones linked from it) if there's been progress made already, if there's already a place (repo or subdirectory) where that progress might take place, etc.

mojavelinux:
One of my goals for 2017 is to see that this happens.

[...]

PR #2802, opened 2018-06-28
Extract stylesheet handling into separate class

martijnversluis:
To extract functionality and responsibility parts of the code can be
extracted into isolated classes that return an array containing lines
of HTML. These classes could even be unit tested if necessary and
desirable.

I'm really curious what you think of this approach.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -2,190 +2,193 @@ require "asciidoctor/converter/html5/stylesheets"
 
 # encoding: UTF-8
 module Asciidoctor
+  module Converter
   # A built-in {Converter} implementation that generates HTML 5 output
   # consistent with the html5 backend from AsciiDoc Python.
-  class Converter::Html5Converter < Converter::BuiltIn
-    (QUOTE_TAGS = {
-      :monospaced  => ['<code>',   '</code>',   true],
-      :emphasis    => ['<em>',     '</em>',     true],
-      :strong      => ['<strong>', '</strong>', true],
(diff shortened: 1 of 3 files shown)


martijnversluis:
The Travis build is failing in a really weird way for two JRuby versions. Could you help me debugging that?

martijnversluis:
@mojavelinux Any thoughts on this?

martijnversluis:
Hi @mojavelinux. Thanks for your reply. The main reason for creating this PR is the fact that you explicitly welcome refactoring PRs in the README. I tried to propose an approach to slowly improve the readability and maintainability of the code. I agree that having a quick conversion is important, but I think it might affect code quality too much.

So I would like to know whether or not you think this approach makes sense. I agree the naming collision with the existing `Stylesheet` class should be resolved.

mojavelinux:
I'm willing to discuss code quality anywhere but the built-in HTML converter. As I stated, I view the built-in converter as compiled code. It's meant to be fast, not readable. Given I'm using Asiidoctor to process tens of thousands of documents, we cannot afford to lose any speed. (I've also become very accustomed to working with it the way it is, so for me it's readable enough given what we are trying to achieve).

At this point, I don't want to modify the current built-in HTML converter because I don't want to continue to invest in it. What I would l like to do is start making headway on a next generation HTML converter that maximizes semantics. See #242. That's where I'd be open to rethinking how we write and maintain the converter. I'll still be concerned about performance, but since it's new territory, we won't have the same expectations to uphold. And a new structure might be more efficient anyway.

martijnversluis:
Ok, fair enough. Then this PR simply does not fit your roadmap I guess. Thanks for your replies anyway.

PR #2813, opened 2018-07-16
resolves #2812 catalog inline anchors in ordered lists

### LINKED ISSUE #2812: Inline anchors in ordered lists are not cataloged
For issue #2257, pull request #2700 was intended to implement cataloging of inline anchors in lists, but the necessary code for ordered lists was missed.

For example, consider this `test.adoc`:

```
* Unordered list, item 1, without anchor
* [[item-2,Item 2]]Unordered list, item 2, with anchor

. Ordered list, item 1, without anchor
. [[step-2,Step 2]]Ordered list, item 2, with anchor

Here is a cross-reference to <<item-2>> (works).
Here is a cross-reference to <<step-2>> (does not work).
```

The resulting HTML is:

```
Here is a cross-reference to <a href="#item-2">Item 2</a> (works).
Here is a cross-reference to <a href="#step-2">[step-2]</a> (does not work).
```

owenh000:
See issue #2812.

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -1309,16 +1309,8 @@ class Parser
         end
       elsif list_type == :olist
         list_item.marker = (sibling_trait ||= resolve_ordered_list_marker(match[1], list_block.items.size, true, reader))
-        if item_text.start_with?('[')
-          if style && style == 'bibliography'
-            if InlineBiblioAnchorRx =~ item_text
-              catalog_inline_biblio_anchor $1, $2, list_item, reader
-            end
-          elsif item_text.start_with?('[[')
-            if LeadingInlineAnchorRx =~ item_text
-              catalog_inline_anchor $1, $2, list_item, reader
-            end
-          end
+        if item_text.start_with?('[[') && LeadingInlineAnchorRx =~ item_text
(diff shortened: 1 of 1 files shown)


lib/asciidoctor/parser.rb
```diff
@@ -1287,25 +1287,26 @@ def self.next_list_item(reader, list_block, match, sibling_trait = nil, style =
       has_text = true
       list_item = ListItem.new(list_block, (item_text = match[2]))
       list_item.source_location = reader.cursor if list_block.document.sourcemap
+      if item_text.start_with?('[')
```
mojavelinux: A lot of this code only applies to :ulist, so moving it outside of the conditional block will cause it to be evaluated too often and even used in cases it shouldn't. Instead, the change should be as follows:

1. Find elsif list_type == :olist
2. Add these lines below it:

```ruby
if item_text.start_with?('[[') && LeadingInlineAnchorRx =~ item_text
  catalog_inline_anchor $1, $2, list_item, reader
end
```

We'll deal with :colist in a separate issue since I was intentionally not supporting that case.

owenh000: Okay, I guess it is quite simple then in that case. I have updated the pull request accordingly. Thanks!

PR #2815, opened 2018-07-20
Issue 2818: catalog inline anchors in callout lists

owenh000:
See issue #2812 comments.


owenh000:
See pull request #2819, which has been updated for the new issue #2818.

PR #2822, opened 2018-07-26
Replace thread_safe with concurrent-ruby

junaruga:
Because thread_safe is deprecated, as it was merged to concurrent-ruby.

> https://github.com/ruby-concurrency/thread_safe
> [DEPRECATED] Thread-safe collections for Ruby (merged with concurrent-ruby)

> https://github.com/ruby-concurrency/concurrent-ruby
> Collection classes that were originally part of the (deprecated) thread_safe gem:
> ...
> Hash A thread-safe subclass of Ruby's standard Hash.
> ...

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -39,6 +39,10 @@ Improvements::
   * use shorthands %F and %T instead of %Y-%m-%d and %H:%M:%S to format time
   * read file in binary mode whenever contents are being normalized
 
+Build / Infrastructure::
+
+  * replace thread_safe with concurrent-ruby (PR #2822) (*@junaruga*)
+
 // tag::compact[]
 == 1.5.7.1 (2018-05-10) - @mojavelinux
 
--- a/lib/asciidoctor/converter/factory.rb
+++ b/lib/asciidoctor/converter/factory.rb
@@ -23,7 +23,7 @@ module Asciidoctor
     # DocBook 5} and {DocBook45Converter DocBook 4.5}, as well as any custom
     # converters that have been discovered or explicitly registered.
     #
-    # If the {https://rubygems.org/gems/thread_safe thread_safe} gem is
+    # If the {https://rubygems.org/gems/concurrent-ruby concurrent-ruby} gem is
     # installed, access to the default factory is guaranteed to be thread safe.
     # Otherwise, a warning is issued to the user.
     class Factory
@@ -32,8 +32,8 @@ module Asciidoctor
 
         # Public: Retrieves a singleton instance of {Factory Converter::Factory}.
         #
-        # If the thread_safe gem is installed, the registry of converters is
-        # initialized as a ThreadSafe::Cache. Otherwise, a warning is issued and
+        # If the concurrent-ruby gem is installed, the registry of converters is
+        # initialized as a Concurrent::Hash. Otherwise, a warning is issued and
         # the registry of converters is initialized using a normal Hash.
         #
         # initialize_singleton - A Boolean to indicate whether the singleton should
@@ -44,14 +44,14 @@ module Asciidoctor
         # Returns the default [Factory] singleton instance
         def default initialize_singleton = true
           return @__default__ || new unless initialize_singleton
-          # FIXME this assignment is not thread_safe, may need to use a ::Threadsafe helper here
+          # FIXME this assignment is not concurrent-ruby, may need to use a ::Threadsafe helper here
           @__default__ ||= begin
             # NOTE .to_s hides require from Opal
-            require 'thread_safe'.to_s unless defined? ::ThreadSafe
-            new ::ThreadSafe::Cache.new
+            require 'concurrent'.to_s unless defined? ::Concurrent
+            new ::Concurrent::Hash.new
           rescue ::LoadError
             include Logging unless include? Logging
-            logger.warn 'gem \'thread_safe\' is not installed. This gem is recommended when registering custom converters.'
+            logger.warn 'gem \'concurrent-ruby\' is not installed. This gem is recommended when registering custom converters.'
             new
           end
         end
--- a/lib/asciidoctor/converter/factory.rb
+++ b/lib/asciidoctor/converter/factory.rb
@@ -44,10 +44,12 @@ module Asciidoctor
         # Returns the default [Factory] singleton instance
         def default initialize_singleton = true
           return @__default__ || new unless initialize_singleton
-          # FIXME this assignment is not concurrent-ruby, may need to use a ::Threadsafe helper here
+          # FIXME this assignment itself may not be thread safe; may need to use a helper here
           @__default__ ||= begin
-            # NOTE .to_s hides require from Opal
-            require 'concurrent'.to_s unless defined? ::Concurrent
+            unless defined? ::Concurrent::Hash
+              # NOTE .to_s hides require from Opal
+              require ::RUBY_MIN_VERSION_1_9 ? 'concurrent/hash'.to_s : 'asciidoctor/core_ext/1.8.7/concurrent/hash'.to_s
+            end
             new ::Concurrent::Hash.new
           rescue ::LoadError
             include Logging unless include? Logging
--- a/Gemfile
+++ b/Gemfile
@@ -4,25 +4,30 @@ source 'https://rubygems.org'
 gemspec
 
 group :development do
-  if (ruby_version = Gem::Version.new RUBY_VERSION) < (Gem::Version.new '2.1.0')
-    if ruby_version < (Gem::Version.new '2.0.0')
-      gem 'haml', '~> 4.0.0'
-      if ruby_version < (Gem::Version.new '1.9.3')
-        gem 'cucumber', '~> 1.3.0'
-        gem 'nokogiri', '~> 1.5.0'
-        gem 'slim', '~> 2.1.0'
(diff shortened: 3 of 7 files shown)


junaruga:
The Travis test was failed for only Ruby 1.8.
https://travis-ci.org/asciidoctor/asciidoctor/builds/408586518

Because `concurrent-ruby` latest version 1.0.5 requires Ruby >= 1.9.3.

To run `concurrent-ruby` on Ruby 1.8, we have to use the old version 0.5.0.
https://rubygems.org/gems/concurrent-ruby/versions/0.5.0
However the version 0.5.0 does not have `Concurrent::Hash` class. it might be implemented at different class.

How do you think to implement to fix it?
Do you like to implement to use use `thread_safe` only if Ruby < 1.9.3 ?

mojavelinux:
Thanks for raising this issue.

Since we're not technically supporting Ruby 1.8.7 anymore, and this is just for tests, you can add an exception in the Gemfile.

Find:

```
if ruby_version < (Gem::Version.new '1.9.3')
```

and add:

```
gem 'thread_safe', '~> 0.3.0'
```

mojavelinux:
Can you also add an entry to the CHANGELOG.adoc when you update the PR? Please use the category "Build / Infrastructure". Thanks!

lib/asciidoctor/converter/factory.rb
```diff
@@ -44,14 +44,14 @@ class << self
...
           return @__default__ || new unless initialize_singleton
-          # FIXME this assignment is not thread_safe, may need to use a ::Threadsafe helper here
+          # FIXME this assignment is not concurrent-ruby, may need to use a ::Threadsafe helper here
           @__default__ ||= begin
             # NOTE .to_s hides require from Opal
-            require 'thread_safe'.to_s unless defined? ::ThreadSafe
-            new ::ThreadSafe::Cache.new
+            require 'concurrent'.to_s unless defined? ::Concurrent
+            new ::Concurrent::Hash.new
           rescue ::LoadError
```
junaruga: > this is just for tests
`thread_safe` is not only for test. It is implemented.
Adding `gem 'thread_safe', '~> 0.3.0'` to `Gemfile` is meaningless, if we do not implement the conditional requiring `thread_safe` logic here.

junaruga:
> Since we're not technically supporting Ruby 1.8.7 anymore, and this is just for tests, you can add an exception in the Gemfile.

## Possible solution 1:

Add only `concurrent-ruby` to `Gemfile`, removing `thread_safe` from `asciidoctor.gemspec`.

```
group :development do
...
  if ruby_version < (Gem::Version.new '1.9.3')
  ...
    else
..
      s.add_development_dependency 'concurrent-ruby', '~> 1.0.5'
    end
```

## Possible solution 2:

Add only `concurrent-ruby` and `thread_safe` to `Gemfile`, removing `thread_safe` from `asciidoctor.gemspec`.

In this solution, we need implementation to use `concurrent-ruby` on Ruby >= 1.9.3, and `thread_safe` on only Ruby < 1.9.3.

## Possible solution 3:

Just remove Ruby 1.8 test case from Travis.

lib/asciidoctor/converter/factory.rb
mojavelinux: Aha! I see what you are saying. Let me think about it. We might be able to put stubs in place for Ruby 1.8.7 or we can just make this part of the 2.0.0 release.

junaruga: Okay, I will wait you :)

junaruga:
By the way, though it's not related to this PR, seeing current `Gemfile`,

> if ruby_version < (Gem::Version.new '1.9.3')

This kind of brackets style `(Gem::Version.new 'X.Y.Z')` is used in the file.
But maybe I think that `Gem::Version.new('X.Y.Z')` style is more popular like below code.

> https://github.com/rubygems/rubygems/blob/v2.7.7/lib/rubygems/specification.rb#L2570
> if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then

Just I would share you the information :)

mojavelinux:
I'm very strongly in favor of lisp-style brackets when writing Ruby. I use it everywhere. I've documented this as part of the style for Asciidoctor projects here: https://github.com/asciidoctor/jekyll-asciidoc/blob/master/coding-style-guide.adoc. That will get moved into the documentation once we roll out 2.0.

junaruga:
Okay, it is great for this project to have a coding style guide.

mojavelinux:
I've proposed a revision which maps Concurrent::Hash to ThreadSafe::Cache in Ruby 1.8.7 so tests will pass. We really aren't worrying about Ruby 1.8.7 at this point, so as long as the CI job runs, it's fine.

mojavelinux:
I recognize this solution is not pretty, but it's just temporary until Asciidoctor 2.0.0, when we'll be dropping support for all these old Ruby versions.

junaruga:
All right. I like the solution. The effort to support Ruby 1.8 until Asciidoctor 2.0.0.

By the way, you pushed your commit to my repository to add your commit in this PR.
I did not  know that was possible. Did I give you the permission in the past time?
https://github.com/junaruga/asciidoctor/commits/feature/replace_thread-safe-to-concurrent-ruby

Anyway, thank you for the working.

CHANGELOG.adoc
```diff
@@ -39,6 +39,10 @@ Improvements::
   * use shorthands %F and %T instead of %Y-%m-%d and %H:%M:%S to format time
   * read file in binary mode whenever contents are being normalized
 
+Build / Infrastructure::
+
+  * replace thread_safe with concurrent-ruby (PR #2822) (*@junaruga*)
+
```
junaruga: This part should be changed? This PR is not only for Build / Infrastructure?

junaruga:
Should I move my comment to another category in `CHANGELOG.adoc`?

CHANGELOG.adoc
mojavelinux: Since it changes the APIs used by this project, you're right. It's not just a dependency change. Therefore, it should now be under Improvements.

mojavelinux:
> By the way, you pushed your commit to my repository to add your commit in this PR.
I did not know that was possible. Did I give you the permission in the past time?

It's a rather recent feature of GitHub and it's the default setting when contributing to this project. This workflow makes it easier to collaborate on changes as you can communicate via code instead of words. It's also very helpful in cases when the contributor is not familiar with git (obviously not the case here).

mojavelinux:
> The effort to support Ruby 1.8 until Asciidoctor 2.0.0.

It's not so much of an effort as just something that was never clearly defined. By moving to semantic versioning, we can do it.

CHANGELOG.adoc
junaruga: Okay, I moved it to Improvement.

junaruga:
> It's a rather recent feature of GitHub and it's the default setting when contributing to this project.

That's great feature for the collaboration. I did not know that.

> you can communicate via code instead of words

I agree. This is really helpful.

CHANGELOG.adoc
mojavelinux: Thanks!

mojavelinux:
Looks good! I'll get this merged tonight.

Mogztter:
We might need to patch Asciidoctor.js and more specifically: https://github.com/asciidoctor/asciidoctor.js/blob/master/lib/asciidoctor/js/opal_ext/thread_safe.rb

I need to check but since we are now using `::Concurrent::Hash`, we may need to define a "fake" class.

lib/asciidoctor/converter/factory.rb
```diff
@@ -44,14 +44,14 @@ class << self
         # Returns the default [Factory] singleton instance
         def default initialize_singleton = true
           return @__default__ || new unless initialize_singleton
-          # FIXME this assignment is not thread_safe, may need to use a ::Threadsafe helper here
+          # FIXME this assignment is not concurrent-ruby, may need to use a ::Threadsafe helper here
```
Mogztter: i think we really want to say that the assignment is not "thread safe" right ?

mojavelinux:
@Mogztter yes, we'll need something mapped in Opal. Something like:

```
module Concurrent
  class Hash < ::Hash
  end
end
```

That will replace the existing patch, though it doesn't hurt to keep both for now.

lib/asciidoctor/converter/factory.rb
mojavelinux: right you are!

mojavelinux:
I made some updates. It should be sound now. I had forgotten about the case in factory.rb. (I am so looking forwarded to be rid of Ruby 1.8.7 from this code base).

junaruga:
@Mogztter thanks for the checking! @mojavelinux thanks for the updates! I checked it.

mojavelinux:
Thanks again! We can use this opportunity to test on AsciidoctorJ and Asciidoctor.js.

cc: @robertpanzer

Mogztter:
@mojavelinux We may have a missing test case because I didn't patch Asciidoctor.js and the build is still green...

mojavelinux:
Keep in mind this is only hit when registering a custom converter class or using the template converter. And Asciidoctor.js may already hit the fallback. But that's just a guess.

Mogztter:
I removed Asciidoctor template.js from the test suite a while back I think that's why. I will add a test with a dummy custom converter.

PR #2823, opened 2018-07-27
Improve files section in gemspec file.

junaruga:
I think we can simplify how to create `s.files` in `asciidoctor.gemspec`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,9 @@ env:
   global:
   # use system libraries to speed up installation of nokogiri
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
-script: bundle exec rake coverage test:all
+script:
+  - bundle exec rake coverage test:all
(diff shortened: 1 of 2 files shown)


mojavelinux:
The files grep is done the way it is to be compatible with all versions of JRuby on all platforms. This is critical for integration with AsciidoctorJ. I put in many iterations on these aggregations to get exactly the result I wanted, so I don't want to introduce regressions by reverting them.

junaruga:
All right. I see that the `asciidoctor` is integrated on windows and non git command environment now.
Thank you for the explanation.

mojavelinux:
I've documented the reasoning behind the logic in the gemspec. It was a good exercise since some of the reasoning was escaping me.

junaruga:
Thank! for the documentation.
It helps people like me who want to know the reason of the implementation in the future.

PR #2826, opened 2018-07-31
Update README.adoc

jwehmschulte:
Added Arch Linux and pacman to the list of Linux package managers.


mojavelinux:
Very cool! Thanks!

PR #2838, opened 2018-08-11
Update sr and sr_Latn locale attributes

BojanStipic:
Update missing attributes for Serbian language.


mojavelinux:
Thanks @BojanStipic!

PR #2841, opened 2018-08-24
resolves #2840 transmit id and role assigned to list item in AST to output

### LINKED ISSUE #2840: HTML converter should transmit id and role on list items
Even though there isn't a syntax yet for adding an ID or role to a list item, these attributes might still be assigned by an extension. Therefore, the built-in HTML converter should look for and transmit the id and role attributes on a list item, if assigned, to the output.

mojavelinux:
resolves #2840


mojavelinux:
I believe this is now ready to go.

PR #2845, opened 2018-08-28
Trivial fix in abstract_node.rb comment

owenh000:
Fix an incorrect word. This was noticed on rubydoc.org while working
with the Asciidoctor API.


mojavelinux:
Thanks @owenh000!

Issue #2764, opened 2018-05-07
Drop support for unmaintained Ruby versions


mojavelinux:
Going forward, Asciidoctor will only support versions of CRuby which are still actively maintained (i.e., receiving updates) according to https://www.ruby-lang.org/en/downloads/. That also means Asciidoctor will only support versions of JRuby that are aligned with these versions of CRuby.

Support equates to which versions of Ruby we test against.

At the time of filing this issue, this includes CRuby 2.2 and up, which means JRuby 9.0.x or JRuby 9.1.x.

What we won't do is set a minimum version of Ruby in the gemspec. The reason for this decision is that it's not a very friendly way to enforce a version, in my experience. It can block installation of gems which depend on asciidoctor that are designed to run on older versions of Ruby, which is an irritating situation and one which cannot be worked around. It's better to clearly state that using an unsupported version of Ruby is an "at your own risk" sort of thing. Enforcement through policy is the route I prefer to take. There's no guarantee Asciidoctor will actually work in those scenarios, but there's no guarantee it won't either.

mojavelinux:
Since we'll be dropping Ruby < 2 (I know...those versions are ancient, but we've been very nice), here are things that can now be changed or refactored:

* remove core_ext patches (for Ruby < 2)
* switch to modern Hash syntax (e.g., `key: value`)
* remove places where magic regexp vars are backed up (look for `m = $~`)
* remove unicode escape sequence hacks (e.g., drop `150.chr`)
* remove `require 'rubygems'` from bin script
* use require_relative instead of modifying the $LOAD_PATH
* use inherit flag of const_get to look for constant in current module only
* use truncate for number precision calculations
* don't enclose method call in brackets when assigning multiple variables
* mark private methods as private (and perhaps use protected as well)
* replace replace .lines.entries with .lines in tests
* consider reading files using r:UTF-8 mode
* drop legacy regexp constants
* remove force unicode line length hack
* remove qualified name lookup patch in extensions
* use gsub w/ hash replacement

mojavelinux:
We will have to be conscious that Opal still relies on some of the patches for old versions of Ruby. We'll just need to tease those apart.

mojavelinux:
The new minimum versions are C Ruby 2.3 and up and JRuby 9.1 and up. The policy is to follow the lifecycle of the runtimes. (After all, why should we support what they don't support?)

mojavelinux:
[...]

PR #2847, opened 2018-08-30
Remove LOAD_PATH manipulation

jhawthorn:
We should rely on rubygems or bundler to add lib/ to $LOAD_PATH.


mojavelinux:
We plan to shift to require_relative in Asciidoctor 2.0 (the next release). I'm delaying this change until then, which coincides with the dropping of unsupported Ruby versions.

mojavelinux:
I'm addressing this as part of #2764. (See #2946 for details)

PR #2849, opened 2018-09-01
resolves #2848 add DSL method to unshift extension onto chain

### LINKED ISSUE #2848: Add DSL method to register document extensions at front of chain
It's currently possible to register a process extension (preprocess, postprocess, tree process, etc) at the front of the chain using `option :position, :>>`. However, there's no DSL to express this option.

Add a DSL method to register the current extension at the front of the chain of processors. For example:

```
  tree_processor do
    unshift
    process do |doc|
      // cal me first
      nil
    end
  end
end
```

Another idea is to allow unshift to be called before tree_processor, though that would only work with the DSL.

mojavelinux:


mojavelinux:
@Mogztter I tried a lot of different names and ended up settling on `prefer`. It communicates the same message as `prepend` but doesn't conflict with the built-in Ruby method. Since I liked it both inside and outside of the extension block, I decided to implement it in both places. See the test for examples.

PR #2851, opened 2018-09-02
resolves #2842 allow reserved IDs to be specified via the API

### LINKED ISSUE #2842: Add option to pass a list of reserved IDs to processor
The processor will not generate the same ID twice. However, it's possible that the document into which the output is embedding has it's own set of IDs. It should be possible to pass a list of reserved IDs to the processor so it avoids generating IDs that collide with them. In fact, the processor itself might start with a default list to avoid generating IDs with those used in the default HTML (header, footer, etc).

mojavelinux:
- allow reserved IDs to be specified using the API option :ids
- register built-in reserved IDs if the API option :ids is not specified


mojavelinux:
We are still missing some of the built-in IDs. Though, that makes me wonder whether we should change some of those to a CSS class instead (the document header comes to mind).

PR #2865, opened 2018-09-08
Escape quotation marks in image URI HTML attribute.

### LINKED ISSUE #2862: html5 backend does not escape image URI value for use in an HTML attribute.
If you run asciidoctor on the following input:

```asciidoc
image::"foo.png"[Foo]
```

it emits the following HTML:

```html
<img src=""foo.png"" alt="Foo">
```

The rendered HTML is invalid, or at least unintuitive, because the quotation marks in the image URI are not getting escaped.

Now, I understand that the input AsciiDoc markup is invalid, but it would not be unusual for a novice to AsciiDoc to make this kind of mistake, especially if they have a programming background and think that a "string" can go there. I think in this case, we should make sure that image URIs are escaped with the necessary escaping rules for HTML attributes.

richardxia:
Fixes #2862.

This changes the HTML backend to escape quotes in image URI, since it is valid, albeit unusual, for an image URI to contain quote characters. This also adds a short test demonstrating this case.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -551,11 +551,11 @@ Your browser does not support the audio tag.
         if svg
           img = (read_svg_contents node, target) || %(<span class="alt">#{node.alt}</span>)
         elsif obj
-          fallback = (node.attr? 'fallback') ? %(<img src="#{node.image_uri(node.attr 'fallback')}" alt="#{encode_quotes node.alt}"#{width_attr}#{height_attr}#{@void_element_slash}>) : %(<span class="alt">#{node.alt}</span>)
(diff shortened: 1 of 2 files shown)


mojavelinux:
I almost pulled the trigger on this for 1.5.8, but for some reason it's really bothering me that we have to look for a double quote in a filename. That's a very abnormal case that would have to be checked on every image processed. I'll keep thinking about it, but it really strikes me as something a validator should have caught (which we can say once we have said validator).

elextr:
Aren't there other characters that need escaping in URIs besides quotes?

mojavelinux:
@elextr Yes, that's true. XML special characters are not escaped either. I suppose that could be the compromise. We should escape XML special characters, but also include the double quote since this is an attribute value. And the character I'm thinking of specifically is & which is not currently escaped.

PR #2877, opened 2018-09-20
resolves #2872 correct list parsing

### LINKED ISSUE #2872: Level-related list parsing code uses superfluous comparisons
The parsing code that consults the list level is invalid. As of right now, the `level` property on a block is the section level, so it doesn't play a factor in list parsing.

We should reconsider the meaning of the `level` property. If necessary, we should add a new property to track the list depth. Then, revisit the logic in `parse_list` to determine when the list level needs to be compared. If there are no such cases, this logic should be removed from the parser.

mojavelinux:
- don't set or check level
- consolidate logic for moving between ancestor and child lists


mojavelinux:
merged as d817fa46adb71038439d59b43aff58ae9bf746d9

Issue #2880, opened 2018-09-20
Crash on attribute and line continuation with no data


zelivans:
As promised another bug found with the fuzzer.

## Raw crash
```
/home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/parser.rb:2131:in `process_attribute_entry': asciidoctor: FAILED: <stdin>: Failed to load AsciiDoc document - undefined method `lstrip' for nil:NilClass (NoMethodError)
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/parser.rb:2107:in `parse_block_metadata_line'
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/parser.rb:2038:in `parse_block_metadata_lines'
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/parser.rb:123:in `parse_document_header'
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/parser.rb:93:in `parse'
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/document.rb:565:in `parse'
	from /home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor.rb:1360:in `load'
	from tst.rb:4:in `<main>'
```

## Explanation
Asciidoctor crashes with an uncaught `NoMethodError` in the `process_attribute_entry`. This happens when the it finds `LINE_CONTINUATION` or `LINE_CONTINUATION_LEGACY` in the end of a value but there is no further data so `reader.peek_line` returns nil, so the `lstrip` call fails.

Reproduce with the following input:

```
:t: t +
```

This is a minimized test case but it can of course work with any attribute and any line continuation symbol. In my PR I made a simple change to prevent the `lstrip` call on nil. Perhaps add a test case to check against this issue if you want.

mojavelinux:
This is an incredible catch. I can't believe we've gotten this far and never noticed this before.

mojavelinux:
Merged as 6e95520ed12617a13fb8d67db74daebaf653e3f5.

PR #2881, opened 2018-09-20
Fix crash caused by empty data after line continuation symbol

zelivans:
Fixes https://github.com/asciidoctor/asciidoctor/issues/2880

This is a simple fix that checks that `next_line` is not nil before attempting to call `lstrip` on it. It can be tested by checking that the following example `:t: t +` does not throw an Exception.

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -2128,7 +2128,7 @@ class Parser
         value = ''
       elsif value.end_with? LINE_CONTINUATION, LINE_CONTINUATION_LEGACY
         con, value = value.slice(-2, 2), (value.slice 0, value.length - 2).rstrip
-        while reader.advance && !(next_line = reader.peek_line.lstrip).empty?
+        while reader.advance && (next_line = reader.peek_line) && !(next_line = next_line.lstrip).empty?
           if (keep_open = next_line.end_with? con)
             next_line = (next_line.slice 0, next_line.length - 2).rstrip
           end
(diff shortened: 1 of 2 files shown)


mojavelinux:
Could you add a test? It doesn't have to be perfect as I'll likely run it myself and make adjustments if necessary. It's just important to demonstrate the fix so that I can get right to testing it.

zelivans:
I wrote a simple test for this. I only tested with `LINE_CONTINUATION` but it follows the same logic as `LINE_CONTINUATION_LEGACY` in the code.

@mojavelinux by the way, in my fix I didn't log any warning when there is no following line after the line continuation symbol, because the bug only happens when it is the last line if I understand correctly. I don't think it really matters because no one is expected to do that anyway, but let me know if you want to also log an error.

mojavelinux:
To avoid an unnecessary call to `lstrip`, I think we can use this logic.

```ruby
while reader.advance && !(next_line = reader.peek_line || '').empty?
  next_line = next_line.lstrip
  if (keep_open = next_line.end_with? con)
```

The redundant call to `empty?` on an empty string is acceptable since at most it can happen once per document.

zelivans:
Thanks, this also feels more like the Ruby way to handle this. Should I commit on top? Or can you edit and own the previous commit?

mojavelinux:
You can just commit on top if you want. When all the commits come in from one person, I tend to just squash them when I merge.

mojavelinux:
Updated and merged as 6e95520ed12617a13fb8d67db74daebaf653e3f5.

zelivans:
Thanks.

mojavelinux:
Thank you!

PR #2893, opened 2018-10-01
resolves #1360 preserve comment guard on callout when icons not enabled

### LINKED ISSUE #1360: Callouts in HTML 5 code listings break syntax when not using ':icons: font'
Callouts in HTML 5 code listings eat comment characters when not using `:icons: font`, thereby breaking copy & paste for code examples with callouts.

```
[source,ruby]
----
puts 'Hello world' # <1>
----
<1> I'd love to be a comment.
```

Without `:icons: font`, this results in a snippet similar to the following (note the missing `#` in front of `(1)`):

``` ruby
puts 'Hello world' (1)
```

The expected behaviour would look more like what `a2x` does:

``` ruby
puts 'Hello world' # (1)
```

Since GitHub uses asciidoctor internally, this effectively breaks code callouts on GitHub-rendered asciidoc files. See this gist for an example:

https://gist.github.com/sometimesfood/d826d5cd364e0ee5e881

mojavelinux:
- when icons are not enabled, preserve the comment guard in front of the callout number in verbatim blocks

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -27,6 +27,7 @@ Enhancements::
   * substitute attribute references in target of custom block macro (honoring attribute-missing setting) (#2839)
   * interpret `<.>` as an auto-numbered callout in verbatim blocks and callout lists (#2871)
   * require marker for items in callout list to have circumfix brackets (e.g., `<1>` instead of `1>`) (#2871)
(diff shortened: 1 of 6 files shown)


cup:
https://github.com/asciidoctor/asciidoctor/issues/1360

cup:
https://github.com/github/markup/issues/1258

PR #2898, opened 2018-10-06
resolves #2798 colspecs can be separated by semi-colon instead of comma

### LINKED ISSUE #2798: Allow semi-colon to be used as column spec separator
To avoid having to quote the value of the cols attribute on a table, allow the semi-colon to be used as an alternate separator character. For example:

```
[cols=1h;2;4]
|===
| Column 1
| Column 2
| Column 3
|===
```

This is consistent with the syntax used for attribute values in the include directive (e.g., tags and lines).

mojavelinux:


elextr:
Appear to be no changes to documentation though?

mojavelinux:
Could you clarify what you mean by documentation in this context? Do you mean the API docs?

elextr:
Isn't this a change to the Asciidoc language, so shouldn't it be documented in the user guide?

mojavelinux:
Yes, for sure. But for now, that's a separate repository (asciidoctor.org). The change has to happen here first, then there. (As soon as the migration to Antora is complete, the docs will be with the code).

elextr:
Oh, didn't realise it was a separate repo, I could never manage a project where they are separate, they would never be aligned.  

+1 to migrating to any solution that makes it one PR including code and documentation.

Issue #2903, opened 2018-10-08
footnote on heading numbered incorrectly


jaredh159:
Given asciidoc like this:

```ADOC
== H2

Para1.footnote:[fn #1]

=== H3footnote:[fn #2]

Para2.footnote:[fn #3]
```

Incorrect markup is generated, with _two_ footnote number "1", as shown here (from `asciidocLive`):

![2018-10-08_13-53-03](https://user-images.githubusercontent.com/7050938/46625346-a648e100-cb01-11e8-9263-1249142ff292.png)

The generated html is:

```HTML
<div class="sect1">
<h2 id="_h2">H2</h2>
<div class="sectionbody">
<div class="paragraph">
<p>Para1.<sup class="footnote">[<a id="_footnoteref_1" class="footnote" href="#_footnotedef_1" title="View footnote.">1</a>]</sup></p>
</div>
<div class="sect2">
<h3 id="_h31">H3<sup class="footnote">[<a id="_footnoteref_1" class="footnote" href="#_footnotedef_1" title="View footnote.">1</a>]</sup></h3>
<div class="paragraph">
<p>Para2.<sup class="footnote">[<a id="_footnoteref_2" class="footnote" href="#_footnotedef_2" title="View footnote.">2</a>]</sup></p>
</div>
</div>
</div>
</div>
<div id="footnotes">
<hr>
<div class="footnote" id="_footnotedef_1">
<a href="#_footnoteref_1">1</a>. fn #2
</div>
<div class="footnote" id="_footnotedef_1">
<a href="#_footnoteref_1">1</a>. fn #1
</div>
<div class="footnote" id="_footnotedef_2">
<a href="#_footnoteref_2">2</a>. fn #3
</div>
</div>
```

jaredh159:
Added a PR with a failing test: https://github.com/asciidoctor/asciidoctor/pull/2904.

jaredh159:
@mojavelinux any thoughts on this?  Known issue? Did you see that I [opened a PR](https://github.com/asciidoctor/asciidoctor/pull/2904) with a failing test?

If I was comfortable in Ruby, or had a few hints where the problem might lie, I'd be willing to try to tackle a solution, but I'm a bit at a loss and needing input. 

Seems like footnotes should work on a heading, and I could find nothing in any of the documentation that would lead me to believe you can't put a footnote on an h3 or other heading element.

jaredh159:
[...]

PR #2904, opened 2018-10-08
add failing test for #2903 -- bad footnote numbering for heading note

jaredh159:
See #2903.

Hopefully the test is sufficient here, never written a line of Ruby before. 🤔


mojavelinux:
The test is off to a good start. We'll probably need some more assertions to verify everything is in the proper place. But before we get to that, we'll need to actually solve the problem. :thinking:

mojavelinux:
Cherry-picked and merged as 156d0108982075ae52888705356a1dab4ad0b4fd.

PR #2908, opened 2018-10-10
resolves #1042 add API to retrieve authors

### LINKED ISSUE #1042: Add API to retrieve collection of authors
Currently the list of authors is stored in the authors attribute as a comma-separated value. Expose this list as a collection.

We should also consider adding an Author class that holds the data for an individual author. That way, we can easily extract all the information for the authors.

This feature is needed in Asciidoctor EPUB3, among other places.

Mogztter:


mojavelinux:
Thanks! At first glance, this looks good. Since it introduces a new API, I'm going to have a close look, but I'm confident that it will make it for 1.5.8.

mojavelinux:
The changes to the converters look good. However, for the API, I'd like to compute the array dynamically based on the document attributes since we want to keep this information in sync. Essentially, this is an attribute-backed API, not a separate data container. I'll submit an update to the PR.

mojavelinux:
> As far as I understand author is an alias for author_1

Essentially, yes. Once parsed, they aren't kept in sync, but initially they have the same value.

> but what if both are defined ? Which one wins ?

This is undefined behavior. With this implementation, the non-indexed version would win (and that's true of the previous implementation too).

> The Document#authors API does not have to deal with that because it's already resolved ?

The API is mimicking the previously behavior for consistently.

> And we can't call Document#authors API if it has been resolved yet ? 

Correct, though this API should only be used after the header has been parsed (which is pretty logically, I think).

> I'm just making sure that we won't have two "first" author smile

I'll admit, the way the authors are processed is rather convoluted. The current behavior attempts to match AsciiDoc Py as closely as possible. But post-2.0.0, it may be time to rethink this whole thing because it's getting out of hand. We may want to move away from being attribute backed and define a delineation between the header and the body (meaning these values can't be changed using attributes after the header is parsed...but that would be a breaking change).

mojavelinux:
Shall we proceed?

Mogztter:
Thanks for the clarification.
Yes we can proceed :+1:

mojavelinux:
Thanks!

PR #2929, opened 2018-10-21
resolves #2928 prevent warning about invalid `:asciidoc` option

### LINKED ISSUE #2928: Slim::Engine emits warning about invalid option when using custom slim templates
Since Slim 4, Slim now validates the names of all options passed to the template. Slim issues a warning for any option that isn't registered with Slim::Engine. Since the `:asciidoc` option is not registered, Slim::Engine emits a warning when using custom slim templates with Asciidoctor.

The fix is to register the `:asciidoc` option so Slim knows to expect it.

mojavelinux:
prevent warning about invalid `:asciidoc` option when using custom templates with Slim 4

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -57,6 +57,7 @@ Fixes::
   * fix crash when restoring passthroughs if passthrough role is enclosed in single quotes (#2882, #2883)
   * don't eagerly apply subs to inline attributes in general
   * make sure encoding of output file is UTF-8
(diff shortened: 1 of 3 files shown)


mojavelinux:
merged as 58b57b55b25bcbffcf74354f02b312fe291321ce

mojavelinux:
merged as 58b57b55b25bcbffcf74354f02b312fe291321ce

mojavelinux:
merged as 58b57b55b25bcbffcf74354f02b312fe291321ce

PR #2933, opened 2018-10-24
resolves #2783 define any character match to be compatible with JavaScript

### LINKED ISSUE #2783: Define a multiline "any" character class for Asciidoctor.js compatibility
Ruby and JavaScript don't have the same idea about the `.` wildcard in regular expressions. That we've already established. But currently, the character classes defined in the Ruby code (`CC_ALL`) is not enough to account for the discrepancy. The dot character in JavaScript also does not match line and paragraph separator characters (see https://github.com/asciidoctor/asciidoctor.js/issues/367). So it's just not reliable. As a consequence, the regular expressions in the transpiled Asciidoctor.js code have to be patched.

```js
data = data.replace(/Opal\.const_set\([^']+'[^']+Rx', *\/.+\/\);$/gm, function (m) {
  return m.replace(/\.([*+])/g, '[^\\n]$1');
});
```

This patching can be avoided by always using a constant in place of a dot wildcard in the regular expressions. However, there isn't just one case. There are, in fact, two. It depends on whether multiline mode is enabled on the expression. Here are the two cases.

* `[^\n]`
* `[\s\S]` (multiline mode)

Currently, all instances of the second case are already represented using the constant `C_ALL`. I propose we split this into two (more accurately named) constants.

* `C_ANY = '[^\n]'`
* `C_MANY = '[\s\S]'` (where MANY is short for "multiline ANY")

This change will require coordination with the Asciidoctor.js build. We may need to temporarily define the constants for Asciidoctor.js in the Ruby code until they are supplied by Asciidoctor.js.

cc: @Mogztter

mojavelinux:


mojavelinux:
@Mogztter See if this branch works with the Asciidoctor.js test suite without this patch:  https://github.com/asciidoctor/asciidoctor.js/blob/master/npm/module/builder.js#L38-L41

I'm trying to make the most minimal change possible so we can remove the patch and not introduce too many unknowns.

mojavelinux:
I tested it and it seems to work. Any it doesn't require a change to Asciidoctor.js (except to remove the patch).

mojavelinux:
@Mogztter I thought about only replacing the `.` where it affected the list parsing (which addresses the original problem). But that seemed kind of arbitrary and would still cause Asciidoctor and Asciidoctor.js to parse documents differently. So I went ahead and just replaced all instances of `.` in regular expressions with `CC_ALL` or `CC_ANY`.

I decided to keep the `CC_ALL` constant (instead of renaming it to `CC_MANY`) and just add `CC_ANY` to match any non-endline character. The names aren't perfect, but the best I could come up in the time I had. The problem with `CC_MANY` is that one might assume it matches multiple characters (when read as "many" instead of "multiline any").

PR #2936, opened 2018-10-27
resolves #2553 allow ampersand in implicit email address

### LINKED ISSUE #2553: Ampersand breaks email links
Ampersand breaks email links

According to [RFC 5322](https://tools.ietf.org/html/rfc5322#page-12 ) ampersand is an allowed character.

I've made several attempts at using escape characters and quotes; so far nothing has worked.

I get the same output using the javascript renderer in atom and the html renderer, however, the pdf renderer ends up with something slightly different.

Asciidoctor Code:
```
me&you@myemail.com

me{amp}you@myemail.com

me&amp;you@myemail.com

http://www.test.com/?='123'&test=34
```

Atom javascript renderer:
ampersand_in_email_javascript

html renderer:
ampersand_in_email_html

pdf renderer:
ampersand_in_email_pdf

mojavelinux:


test/substitutions_test.rb
```diff
@@ -629,6 +629,8 @@
       assert_equal %q{<a href="mailto:author+website@4fs.no">author+website@4fs.no</a>}, para.sub_macros(para.source)
       para = block_from_string('john@domain.uk.co')
       assert_equal %q{<a href="mailto:john@domain.uk.co">john@domain.uk.co</a>}, para.sub_macros(para.source)
+      para = block_from_string('bert&ernie@sesamestreet.com')
+      assert_equal %q{<a href="mailto:bert&amp;ernie@sesamestreet.com">bert&amp;ernie@sesamestreet.com</a>}, para.apply_subs(para.source)
```
Mogztter: For a second I read Bernie! Feel the Bern 😆

mojavelinux: :heart_eyes: :fire:

PR #2946, opened 2018-10-29
resolves #2764 remove code used to support older versions of Ruby

### LINKED ISSUE #2764: Drop support for unmaintained Ruby versions

mojavelinux:
resolves #2764

* remove workarounds for Ruby < 2.3
* use require_relative instead of modifying $LOAD_PATH
* use shorthand for character references
* remove brackets around method call when assigning multiple variables
* scope constant lookups
* modernize UTF-8 encoding operations
* read input File as string
* rename Helpers.normalize_lines* methods
* switch to modern hash syntax
* remove unused code; remove trailing blank line in class body
* add visibility modifiers to methods
* Reader#push_include should not fail if data is nil
* use .adoc as AsciiDoc file extension in test suite


mojavelinux:
@Mogztter Now would be a good time to shake the tree and see if this causes Asciidoctor.js and problems.

Mogztter:
Alright I will run the tests suite against your branch

mojavelinux:
Buh-bye Ruby 1.8 and 1.9! :bomb:

PR #2948, opened 2018-10-31
New attributes for Ruby constants (ruby-version, ruby-revision, etc).

PauloFrancaLacerda:
Firstly, I'm embarrassingly newbie about SVN and such. Never used it in my whole life, so, please, forgive for any stupid thing I might have done here and, obviously, lead me in the right direction.

I've added Ruby constants ("ruby-version", "ruby-revision", etc) for informational purposes.
One of them I personally use in my doc's footer, like this:

ADOC:
`:last-update-label: Made with Asciidoctor {asciidoctor-version} (on Ruby {ruby-version} for Windows) in`

RESULT:
`Made with Asciidoctor 1.5.8 (on Ruby 2.5.3 for Windows) in 2018-10-22 13:17:18 -0300`
footer

### DIFF
--- a/lib/asciidoctor/document.rb
+++ b/lib/asciidoctor/document.rb
@@ -385,6 +385,14 @@ class Document < AbstractBlock
     attr_overrides['asciidoctor'] = ''
     attr_overrides['asciidoctor-version'] = VERSION
 
+    # Ruby constants added by Paulo França Lacerda (BlaisePoint Informática)
+    attr_overrides['ruby-engine']       = RUBY_ENGINE
+    attr_overrides['ruby-version']      = RUBY_VERSION
+    attr_overrides['ruby-revision']     = RUBY_REVISION
+    attr_overrides['ruby-patchlevel']   = RUBY_PATCHLEVEL
+    attr_overrides['ruby-platform']     = RUBY_PLATFORM
+    attr_overrides['ruby-release-date'] = RUBY_RELEASE_DATE
+
     attr_overrides['safe-mode-name'] = (safe_mode_name = SafeMode.name_for_value @safe)
     attr_overrides["safe-mode-#{safe_mode_name}"] = ''
     attr_overrides['safe-mode-level'] = @safe


PauloFrancaLacerda:
In my personal Asciidoctor "proof" html:
ruby-attrs

mojavelinux:
I like this idea. And you did a fine job making the PR.

I think we should put this behind a feature flag. It should also be only allowed in safe or unsafe mode since this can be a security vector (it has the potential to give attackers information about a remote system).

I'm thinking for the feature flag, something like `system-info` or `runtime-info`. So if you set the following in your document:

```
= Document Title
:system-info:
```

Then the various system-related attributes will get set. And it's agnostic, so other implementation can set other information. For example, AsciidoctorJ can set the Java version and Asciidoctor.js can set the JavaScript version.

cc: @Mogztter

PauloFrancaLacerda:
> 
> 
> I like this idea. And you did a fine job making the PR.

Thank you.
> 
> I think we should put this behind a feature flag. It should also be only allowed in safe or unsafe mode since this can be a security vector (it has the potential to give attackers information about a remote system).

I didn't think about the secure issue. Good catch!

> 
> I'm thinking for the feature flag, something like `system-info` or `runtime-info`. So if you set the following in your document:
> 
> ```
> = Document Title
> :system-info:
> ```

Well, I have no idea on how to implement this in Ruby.
I'd better leave it for real Ruby programmers.  \\\\:D

> 
> Then the various system-related attributes will get set. And it's agnostic, so other implementation can set other information. For example, AsciidoctorJ can set the Java version and Asciidoctor.js can set the JavaScript version.

Looks good. It'd open an useful door to users of all implementations.

mojavelinux:
> Well, I have no idea on how to implement this in Ruby.
I'd better leave it for real Ruby programmers. \\:D

I can certainly help with that part. What's most important is arriving at a decision about requirements. Let's see what @Mogztter and @robertpanzer have to say about the naming.

elextr:
Agree it should be translator agnostic, but maybe it could still be a few separate attributes:

```
:systeminfo: Windows 10 or Ubuntu 18.04 or macos <some animal> 10.7 etc
:translator: Asciidoctor 1.5.8 or Asciidoc Python 9.0.0 etc
:platform: Ruby 2.0.0 or Python 3.6 etc
```

PauloFrancaLacerda:
> Agree it should be translator agnostic, but maybe it could still be a few separate attributes:
> 
> :systeminfo: Windows 10 or Ubuntu 18.04 or macos <some animal> 10.7 etc

I believe Dan suggested "system-info" as a flag, not as a string compound attribute.

elextr:
> I believe Dan suggested "system-info" as a flag, not as a string compound attribute.

Yes, thats why I didn't use a hyphen in the attribute name, or you could call it `OSinfo` or some other name.  If `system-info` is set and its safe to do so, the attributes I suggested would have values, otherwise they would be defined but empty so a document with a line like your original example:

```
Made {translator} {platform} {systeminfo} {datetime}
```

 would still say "Made 2018-10-22 13:17:18 -0300" if the `system-info` attributes were not allowed.

PauloFrancaLacerda:
> ...If `system-info` is set and its safe to do so, the attributes I suggested would have values, otherwise they would be defined but empty so a document with a line like your original example:
> 
> ```
> Made {translator} {platform} {systeminfo} {datetime}
> ```
> 
> would still say "Made 2018-10-22 13:17:18 -0300" if the `system-info` attributes were not allowed.

Setting them to empty strings is indeed a simple and efficient way to garantee convenient results (as in your example), despite the value of `attribute-missing` and `attribute-undefined`.

PauloFrancaLacerda:
> 
> 
> Interesting, I like this idea.
> If the attributes are related to the runtime (Ruby, JavaScript, Java), I think the flag should be `runtime-info`. But we could also add a flag named `system-info` that will define system attributes:
> 
>     * `system-platform`
>     * `system-version` or `system-release`
>     * `system-arch`
>     * ...

Do the other implementations  have the same system details like those exposed by Ruby? I mean... version, release number/date, patch level, etc...
I'm asking this because it would be nice if the agnostic attributes could be the more detailed as possible but without losing consistency.

PauloFrancaLacerda:
> More or less wink
> ...patchlevel and revision are really specific.

So, we could provide those specific attributes for the Ruby platform and let other implementations leave them as empty strings?

mojavelinux:
I've decided that this is best done as an extension and I have added a sample extension to the extensions lab for demonstration. See https://github.com/asciidoctor/asciidoctor-extensions-lab/blob/main/lib/ruby-attributes-preprocessor/extension.rb

Adding these attributes couples AsciiDoc too closely to the runtime for them to be part of core. It also has the risk of exposing too much information from the server, which could be used as an attack vector...so I'm sure that GitHub and GitLab would frown on it. Since it's easy enough to use an extension, I'm going to leave it at that.

PR #2954, opened 2018-11-03
Remove Ruby 1.8 support

Mogztter:
Asciidoctor.js is working fine (no regression) :tada: 

**Note:** I'm using this branch to check that Asciidoctor.js is running fine but we don't necessarily want to merge it :wink:


mojavelinux:
I already have a PR started for this change. See https://github.com/asciidoctor/asciidoctor/pull/2946. Could you switch to that one instead?

Mogztter:
Oops I didn't check :flushed: 

> Could you switch to that one instead?

Yep :+1:

mojavelinux:
Thanks!

PR #2964, opened 2018-11-11
Fixes #2963. Implement Table::Cell.block? and Table::Column.block?

### LINKED ISSUE #2963: Table::Cell and Table::Column don't implement block?
`Table::Cell` and `Table::Column` extend `AbstractNode` which defines the method `block?`.
However these classes don't provide an own implementation so that calling `block?` raises an exception.
Returning `false` seems to be a reasonable result to me.

robertpanzer:
This PR simply implements the method `block?` for `Table::Cell` and `Table::Column` so that they no longer raise an exception when being called.

### DIFF
--- a/lib/asciidoctor/table.rb
+++ b/lib/asciidoctor/table.rb
@@ -215,6 +215,10 @@ class Table::Column < AbstractNode
     end
     col_pcwidth
   end
+
+  def block?
+    false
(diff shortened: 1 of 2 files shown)


mojavelinux:
This was resolved by #3103. Thanks!

PR #2968, opened 2018-11-23
Pass the environment variables to popen3 using the builtin parameter

aerostitch:
This is just a nitpick.
To provide environment variables to popen3, it's best to use a hash as parameter to the popen3 method as shown in https://docs.ruby-lang.org/en/2.5.0/Open3.html#method-i-popen3
This way you also don't need a mechanism to set the environment variables that you want to change.

### DIFF
--- a/test/invoker_test.rb
+++ b/test/invoker_test.rb
@@ -632,21 +632,15 @@ eve, islifeform - analyzes an image to determine if it's a picture of a life for
     ruby = File.join RbConfig::CONFIG['bindir'], RbConfig::CONFIG['ruby_install_name']
     executable = File.join ASCIIDOCTOR_PROJECT_DIR, 'bin', 'asciidoctor'
     input_path = fixture_path 'encoding.asciidoc'
-    old_lang = ENV['LANG']
-    ENV['LANG'] = 'US-ASCII'
-    begin
-      # using open3 to work around a bug in JRuby process_manager.rb,
-      # which tries to run a gsub on stdout prematurely breaking the test
-      cmd = %(#{ruby} #{executable} -o - --trace #{input_path})
-      # warnings may be issued, so don't assert on stderr
-      stdout_lines = Open3.popen3(cmd) {|_, out| out.readlines }
-      refute_empty stdout_lines
-      stdout_lines.each {|l| l.force_encoding Encoding::UTF_8 } if Asciidoctor::FORCE_ENCODING
-      stdout_str = stdout_lines.join
-      assert_includes stdout_str, 'Codierungen sind verrückt auf älteren Versionen von Ruby'
-    ensure
-      ENV['LANG'] = old_lang
-    end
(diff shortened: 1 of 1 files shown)


aerostitch:
Note that I'm working on that specific one because when packaging it in Debian, the test about the non-UTC timezone is failing in a chroot environment with:
```
  1) Failure:                                                                             
Minitest::Test::TestInvoker#test_should_show_timezone_as_offset_if_system_TZ_is_not_set_to_UTC [/<<PKGBUILDDIR>>/test/invoker_test.rb:699]:
Expected false to be truthy. 
```
And the dates in the generated file are actually UTC:
```
23:57:32 UTC
23:57:32 UTC
```
This change doesn't fix it, but I thought it would be cleaner to change that at least.

aerostitch:
@mojavelinux do you know if asciidoctor is started in a separate subprocess in jruby?
That would explain why passing them without an export makes the test fail maybe?
I don't understand why this is not working for jruby even if the syntax to provide environment variables is correct: https://github.com/jruby/jruby/blob/master/test/mri/test_open3.rb#L78
And also why, even if I changed 3 tests, only 1 of them fails. Are the other env variables set to their default, making the tests pointless?

aerostitch:
The fix for the test failure in the Debian env I was mentioning is: #2969

aerostitch:
If I'm correct, the failure I see should be fixed in JRuby 9.2.1 thanks to this patch: https://github.com/jruby/jruby/pull/5380/

mojavelinux:
> To provide environment variables to popen3, it's best to use a hash as parameter to the popen3 method 

I believe the reason I did it the way I did it is because passing a hash was not portable across the versions of Ruby we were supporting. But that was when we supported Ruby < 2.3. Now that master has dropped old versions of Ruby, we may be able to revisit.

mojavelinux:
> do you know if asciidoctor is started in a separate subprocess in jruby?

When we invoke the cli directly, it should be.

> I don't understand why this is not working for jruby even if the syntax to provide environment variables is correct

Join the club. I see you mentioned that it might now be fixed in JRuby. But these are the tests I bang my head against all the time. There is so much variation across the different versions of Ruby when it comes to subprocesses and timezones, and the combination of the two.

If we can make the build green on master, that's really all that matters.

aerostitch:
Sorry for the late reply, it's been crazy those last months and your answers got lost in my mailbox.
Which versions of JRuby you want the project to support?

mojavelinux:
Asciidoctor 2.0.0 (which is master) will support JRuby 9.1 and 9.2. It will also follow the CRuby schedule from now on (so Ruby 2.3 and up).

aerostitch:
Ok, so I'm guessing we can't merge this one before we drop the support for JRuby < 9.2.1 so I'm going to close this one.
Sorry for the noise.

PR #2969, opened 2018-11-23
Take in account SOURCE_DATE_EPOCH when testing against timezones

aerostitch:
Just lost 3 hours trying to figure out why, when running in a reproducible build environment, this timezone test was failing. Silly me! :laughing: 
Anyway, this patch allows you to run the tests properly in a reproducible builds environment (aka with `SOURCE_DATE_EPOCH` set).

### DIFF
--- a/test/invoker_test.rb
+++ b/test/invoker_test.rb
@@ -686,8 +686,6 @@ Sample *AsciiDoc*
       end
       if old_source_date_epoch
           ENV['SOURCE_DATE_EPOCH'] = old_source_date_epoch
-      else
-          ENV.delete 'SOURCE_DATE_EPOCH'
       end
     end
   end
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks! I had not thought about the tests running in the reproducible build environment. Thanks for catching this.

test/invoker_test.rb
```diff
@@ -682,6 +684,11 @@ def readlines
       else
         ENV.delete 'TZ'
       end
+      if old_source_date_epoch
+          ENV['SOURCE_DATE_EPOCH'] = old_source_date_epoch
+      else
+          ENV.delete 'SOURCE_DATE_EPOCH'
```
mojavelinux: I don't think it's necessary to delete the SOURCE_DATE_EPOCH environment variable after the test since Asciidoctor doesn't set this value. Can you remove it and retest?

aerostitch:
I'm so sorry, your answers totally slipped through the cracks.
I just updated the PR with your remarks. Feel free to squash when merging if you feel it's now acceptable.
Thanks!

mojavelinux:
Thanks!

mojavelinux:
Really nice catch, btw!

PR #2970, opened 2018-11-23
Refactor the timezone-related tests to use the usual invoke_cli_to_buffer instead of popen3

aerostitch:
The idea is to try to keep the same way of running the tests.
It is not required to fork a new process here as the call to `::Time.now` calls glibc under the hood which will take in account the environment variable dynamically.

This PR embeds #2969. Keeping both open in case you want to handle the 2 issues separately.
If this PR is ok with you we can close #2968.

### DIFF
--- a/test/fixtures/doctime-localtime.adoc
+++ b/test/fixtures/doctime-localtime.adoc
@@ -1,2 +1,0 @@
-{doctime}
-{localtime}
--- a/test/invoker_test.rb
+++ b/test/invoker_test.rb
@@ -665,43 +665,51 @@ Sample *AsciiDoc*
   end
 
   test 'should show timezone as UTC if system TZ is set to UTC' do
-    ruby = File.join RbConfig::CONFIG['bindir'], RbConfig::CONFIG['ruby_install_name']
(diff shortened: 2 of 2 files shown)


aerostitch:
I'm wondering if the JRuby failure is due to the fact that the JVM only loads the timezone data when it starts?
I don't know JRuby really well, so maybe there's another way to fix that.

mojavelinux:
Invoking the CLI in tests is really slow and something I'd rather avoid. Is it really necessary to convert these to CLI tests?

I'll admit that I wasted a lot of time getting the timezone tests working across all the versions of Ruby, which is why I'm a bit reluctant to shake the tree.

To answer your question, yes, every version of Ruby seems to do something different, which also varies by platform. It's kind of huge headache to test timezones.

mojavelinux:
> I'm a bit surprised that the cli tests are slower than calling the asciidoctor executable via popen3.

It's not popen3 vs the cli that's the problem. Most of the invoker tests use the API directly. Only a handful of tests actually drop down to an external Ruby process. And that's slow because it has to cold boot the Ruby runtime, which for JRuby can take a half second or more.

aerostitch:
right! makes sense, sorry! :)

PR #2978, opened 2018-11-30
resolves #2983 remove conditional comments for IE

### LINKED ISSUE #2983: Side navigation will be broken in IE11
There is a checkbox that "display intranet sites in compatibility view" in IE11 [Tools -> Compatibility View setting]. This is enable by default. So, If we put documents on an intranet,  It will be Compatibility Mode 5 or 7. Because, "IE=edge" has been commented by the conditional comment. Conditional Comments is just a comment on IE10 and more. For that reason, Style will be broken.

Gitter
https://gitter.im/asciidoctor/asciidoctorj?at=5c066dcb6c3c565096d0bd93

My pull request
https://github.com/asciidoctor/asciidoctor/pull/2978

e.g.
image

Hentaro1989:
Conditional Comments is unavailable on IE10 and more. So, I think we should remove the Conditional Comments.

e.g.
If we browse documents on intranet with IE, The Conditional Comments will be just a comment. Then, IE will work as older Document mode.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -44,7 +44,7 @@ module Asciidoctor
       result << %(<html#{@xml_mode ? ' xmlns="http://www.w3.org/1999/xhtml"' : ''}#{lang_attribute}>)
       result << %(<head>
 <meta charset="#{node.attr 'encoding', 'UTF-8'}"#{slash}>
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks @Hentaro1989!

Hentaro1989:
@mojavelinux My pleasure.

PR #2984, opened 2018-12-06
migrate fontawesome5 and render svg

### LINKED ISSUE #2836: Font Awesome CDN uses old version
I noticed the Font Awesome CDN defaults to 4.7. The latest can be found here, which is 5.2 right now: https://fontawesome.com/how-to-use/on-the-web/setup/getting-started?using=web-fonts-with-css

CDN: https://use.fontawesome.com/releases/v5.2.0/css/all.css

Manually overriding with :iconfont-cdn: works, but it would be great if the default were updated.

ghost:
It is huge changing.
I define svg file in yaml.

So,we try to migrate newest fontawesome 5.

of cource ,this PR fail test.

sry,my poor English. 

fix #2836


asdfg0280:
write new mode test

lib/asciidoctor/converter/html5.rb
```diff
@@ -1,3 +1,5 @@
+require 'yaml'
```
Mogztter: I don't think it's a good idea to use the `yaml` module because I doubt it will work in Asciidoctor.js.

lib/asciidoctor/converter/html5.rb
```diff
@@ -1117,12 +1119,23 @@ def inline_footnote node
 
     def inline_image node
       if (type = node.type) == 'icon' && (node.document.attr? 'icons', 'font')
-        class_attr_val = %(fa fa-#{node.target})
-        {'size' => 'fa-', 'rotate' => 'fa-rotate-', 'flip' => 'fa-flip-'}.each do |key, prefix|
-          class_attr_val = %(#{class_attr_val} #{prefix}#{node.attr key}) if node.attr? key
+        # load font awesome5 list
+        font_awesome5 = YAML.load_file(File.dirname(__FILE__)+ '/const/font_awesome5.yml')
```
Mogztter: We shouldn't load the file every time `inline_image` is called

Mogztter:
@asdfg0280 I'm curious, how do you generate the file `font_awesome5.yml` ?

asdfg0280:
I generated that yml by `https://use.fontawesome.com/releases/v5.5.0/css/all.css`

asdfg0280:
I generated that yml by `https://use.fontawesome.com/releases/v5.5.0/css/all.css`

lib/asciidoctor/converter/html5.rb
asdfg0280: Okay, I try to another plan.

I intend to define `svg` .And write `svg` .

asdfg0280:
I define FontAwesome5 in `fontawesome.rb`

.gitignore
```diff
@@ -9,3 +9,4 @@ pkg/
 /.yardoc/
 /rdoc/
 /shippable/
+vendor/*
```
mojavelinux: The preferred approach in the Asciidoctor ecosystem is to store your local gems in .bundle/gems. The .bundle folder is already excluded. I don't want to add another exclusion.

data/stylesheets/asciidoctor-default.css
```diff
@@ -361,8 +361,13 @@ div.unbreakable{page-break-inside:avoid}
 .white-background{background-color:#fafafa}
 .yellow{color:#bfbf00}
 .yellow-background{background-color:#fafa00}
-span.icon>.fa{cursor:default}
-a span.icon>.fa{cursor:inherit}
+svg{
```
mojavelinux: Changes to the stylesheet need to be made in the stylesheet factory, then rebuilt as a minimized file. See https://github.com/asciidoctor/asciidoctor-stylesheet-factory

mojavelinux:
We're going to need to think of a way to do this without importing all of FontAwesome into Asciidoctor. That's not sustainable. We either need to find an existing gem that provides FontAwesome, or make one.

PR #2985, opened 2018-12-09
Remove the warning when the runtime is Opal/JavaScript

Mogztter:
When registering a custom converter with Asciidoctor.js the following warning message is printed:

> asciidoctor: WARNING: gem 'concurrent-ruby' is not installed. This gem is recommended when registering custom converters.

This warning message does not make sense in Asciidoctor.js and shouldn't be printed to avoid confusion.

Note: I've added a test case (on my local workspace) but I don't know how to simulate the absence of `concurrent-ruby` and/or to fake an `opal` Ruby engine...

### DIFF
--- a/lib/asciidoctor/converter/factory.rb
+++ b/lib/asciidoctor/converter/factory.rb
@@ -44,17 +44,20 @@ module Asciidoctor
         # Returns the default [Factory] singleton instance
         def default initialize_singleton = true
           return @__default__ || new unless initialize_singleton
-          # FIXME this assignment itself may not be thread safe; may need to use a helper here
-          @__default__ ||= begin
-            unless defined? ::Concurrent::Hash
-              # NOTE dynamic require is ignored by Opal
-              require ::RUBY_MIN_VERSION_1_9 ? 'concurrent/hash' : 'asciidoctor/core_ext/1.8.7/concurrent/hash'
+          if RUBY_ENGINE == 'opal'
+            @__default__ ||= new
+          else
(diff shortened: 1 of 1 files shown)


lib/asciidoctor/converter/factory.rb
```diff
@@ -53,7 +53,9 @@ def default initialize_singleton = true
             new ::Concurrent::Hash.new
```
mojavelinux: But isn't Asciidoctor.js providing Concurrent::Hash so that this LoadError isn't hit in the first place?

Mogztter:
I've updated my pull request.
I think it's better to not execute code that does not make sense in an Opal environment. With this special condition, you will be able to remove the dynamic require (when Ruby 1.9 support will be dropped).

mojavelinux:
> you will be able to remove the dynamic require (when Ruby 1.9 support will be dropped).

FYI, master has already dropped support for Ruby < 2.3.

Mogztter:
@mojavelinux But why do we still have the following line:

```rb
require ::RUBY_MIN_VERSION_1_9 ? 'concurrent/hash' : 'asciidoctor/core_ext/1.8.7/concurrent/hash'
```

mojavelinux:
Oh yeah, I just haven't merged it yet. Antora 2.0 release was holding me up. Asciidoctor 2.0 release is my next focus...call it my holiday project.

mojavelinux:
Yep, I agree. Thanks!

PR #2999, opened 2019-01-07
resolves #2995 use methods on File class instead of IO class

### LINKED ISSUE #2995: Use methods on File instead of IO
The read, binread, and write methods on the IO class add extra functionality that we don't need. In particular, they look for a magic symbol at the start of the filename for invoking a subprocess. Although Asciidoctor already protects against this behavior (by expanding the filename to a system path), we still don't need Ruby to be looking for it.

mojavelinux:
resolves #2995


Mogztter:
I should probably run the Asciiodoctor.js tests suite against this branch :wink:

mojavelinux:
@Mogztter Are we ready to proceed?

Mogztter:
@mojavelinux Yes, the build is working and I will make some adjustments in https://github.com/asciidoctor/asciidoctor.js/pull/634

PR #3010, opened 2019-01-10
resolves #1121 introduce 4+ tilde fences for open block

### LINKED ISSUE #1121: Allow open blocks to be nested

mojavelinux:


elextr:
Since the standardisation of Asciidoc is about to start, this is not the time to be making major syntax changes.

mojavelinux:
Sorry, this has to happen. I cannot complete certain tasks without it. No one is forcing anyone to use this feature, but I will be making use of it.

elextr:
>  but I will be making use of it.

That is irresponsible and premature, operating within a community means not taking a unilateral action that has impact on others.  You have now made a potentially incompatible version of Asciidoc, the thing you profess to be against.

mojavelinux:
I disagree. We have talked about this for a very long time and have always said it would be included in the next major version of Asciidoctor. It's absolutely critical for doing any sort of advanced block extension where you need a nested structure. Many people are blocked on it because of this.

mojavelinux:
I'll also point out that the four tilde syntax was supported in AsciiDoc Python for a very long time. It was only in the 8.x releases that it disappeared. So it's not entirely new to AsciiDoc.

mojavelinux:
In this area of delimited blocks, Asciidoctor already differs drastically from AsciiDoc Python in that Asciidoctor allows nested blocks. So again, this is an optional feature that Asciidoctor has always brought to the table which will not affect any existing documents.

mojavelinux:
(I do believe that the four tilde syntax should be standardized, as well as nested blocks, but that will be a proposal I will put forth to the working group when the time comes).

mojavelinux:
I key part of proposing a code-first spec is that we have to demonstrate a concept working in the real world. And that's exactly why it makes total sense to get this into Asciidoctor now on that account (beyond my point about needing it for developing extensions).

mojavelinux:
I took a walk to get a fresh perspective and think about what you said.

Here's the compromise I'm open to making. I'll label this syntax in the CHANGELOG and documentation as EXPERIMENTAL, with the explanation that we're putting it out there to get feedback on whether this is the direction we want to go. This is very typical practice in the specification process (which we're not yet in, but nonetheless). This allows early adopters to write complex block extensions and see whether it addresses the issue I alluded to. I will also explain that we plan to submit this as an enhancement to AsciiDoc, but that until that happens, it could change at any time. That at least allows people to go into it with the right expectations.

wdyt?

elextr:
I'm sorry if I was abrupt, and I am not necessarily against this, but if you start using a new syntax it leaves little choice for the standards process, either it accepts it or it invalidates any documents that have used it.  

Also I don't see why you are deprecating the double dash version, there is no reason to remove it, it does not clash with any other syntax, and removing it will invalidate the majority of existing documents that have used it for list element extension.  And if it isn't going to be removed it doesn't need to be deprecated, documenting that the alternative has advantages is important though.

I wasn't aware of Asciidoc Python using tilde, the only reference I can find is in the changelog to add a word followed by four tildes, eg `code~~~~`, no reference to when it was removed though.  (But glad it was, that was a ugly idea :)

Yes nested open blocks are important and yes open blocks fixed syntax does not allow nesting.  It needs to be fixed, but unilaterally picking a syntax just as the standards process starts sends the wrong message.  

I am not so sure about some of the other block nestings though, what exactly are the semantics of a sidebar inside an example, not to mention a sidebar inside a sidebar?  But thats something the standards process can examine.

Yes the standards process needs to demonstrate that it is implementable, but implementing it now as part of the next public release of Asciidoctor is different to implementing it in an experimental standards compliant version.  I didn't think you would want to experiment on the loyal Asciidoctor users.

You posted your suggestion of making it experimental while I was writing this and it is exactly what I was about to suggest, (though I had dinner not a healthy walk :grin:).  I'm glad Github made the post visible in time.  

If Asciidoctor has a process for marking things experimental and its documented as that, then it sounds like the best solution.  I am aware that you are about to release a major version Asciidoctor 2.0.0 that may have incompatibilities, and this way will allow you to add it in time and start using it.

mojavelinux:
> If Asciidoctor has a process for marking things experimental and its documented as that, then it sounds like the best solution

:+1: 

Keep in mind, this is very unlikely to be used on GitHub since it's for extensions and GitHub doesn't support extensions. If you nested open blocks on GitHub, it would just look like plain paragraphs. So this really is about tailoring documents to a specific environment.

I'll give you an example for reference: https://docs.couchbase.com/home/contribute/tabs.html (we want to use an open block for that instead of an example block, but then it forbids the use of open blocks in the content).

mojavelinux:
> I'm sorry if I was abrupt

No need to apologize. Challenging decisions and assumptions is an important part of the process. I'm abrupt too sometimes because I'm trying to move forward. When the matter is important, it needs to be voiced with equal conviction. As long as we keep the debate focused on the topic (as I believe we have here), we'll be productive.

elextr:
> > I don't see why you are deprecating the double dash version

> I never said I was.

Sorry, I read #1121 as saying you would, so I assumed you did.

A very simple use-case I see for nested open blocks is list extensions on lists inside list extensions, (also testing my understanding of the proposal).

```
* blah 1
+
~~~~
some stuff in an open block extending a list item

some more stuff introducing a list inside the extended list item

. an item on a list inside an extended list item
+
~~~~~~
an open block extending a list item inside an open block list item extension

some more
~~~~~~

. another item on the list inside the extended list item
~~~~

* blah 2
+
etc
```

Hopefully I got that right and it made sense.

mojavelinux:
> I read #1121 as saying you would, so I assumed you did.

I just wanted to make sure there wasn't any confusion about my current position. Removing core syntax would be painful at this point. Whenever this new syntax becomes generally available in AsciiDoc, I will certainly discourage the use of two-character syntax because it's simply not consistent. But that's very different from deprecating or removing it.

mojavelinux:
> Hopefully I got that right and it made sense.

Yep, that covers it.

mojavelinux:
I've decided to defer this until the spec process is underway (which would be 2.1.0 or later). See my statement here: https://github.com/asciidoctor/asciidoctor/issues/1121#issuecomment-460037338

mojavelinux:
The spec process is underway and this was the first accept language change. See https://gitlab.eclipse.org/eclipse/asciidoc-lang/asciidoc-lang/-/blob/main/spec/sdrs/sdr-001-open-block-delimiter.adoc?ref_type=heads

PR #3015, opened 2019-01-12
rename Helpers.uri_encode to Helpers.encode_uri_component and use native function

mojavelinux:
- rename Helpers.uri_encode to Helpers.encode_uri_component (to borrow terminology from JavaScript)
- implement Helpers.encode_uri_component using CGI.escape in *Ruby and encodeURIComponent in Opal
- which characters are encoded now complies with RFC-3986
- move Helpers tests to helpers_test.rb


mojavelinux:
Technically, this gives us a more accurate result. But don't search about the correct way to encode a component for a URI (especially spaces). It's a rabbit hole.

The reason for using CGI.escape is because it's accelerated (implemented in C). But since JavaScript already has a native function, we can just use that. However, it might be nice if this could be wired up in Opal. A slight patch is needed to get the two implementations to agree.

mojavelinux:
I also think the name encode_uri_component is much more clear than uri_encode, so I made the switch.

lib/asciidoctor/helpers.rb
```diff
@@ -126,13 +126,23 @@ def self.uri_prefix str
...
-  #
-  # Returns the String with all URI reserved characters encoded.
-  def self.uri_encode str
-    str.gsub(UriEncodeCharsRx) { $&.each_byte.map {|c| sprintf '%%%02X', c }.join }
+  # Internal: Encode a URI component String for safe inclusion in a URI.
+  #
+  # str - the URI component String to encode
+  #
+  # Returns the String with all reserved URI characters encoded (e.g., /, &, =, space, etc).
+  if RUBY_ENGINE == 'opal'
```
Mogztter: If you want I can move this code in Asciidoctor.js

mojavelinux: Better would be to have CGI.escape implemented in Opal.

lib/asciidoctor/helpers.rb
```diff
@@ -126,13 +126,23 @@ def self.uri_prefix str
...
+    def self.encode_uri_component str
+      %x(
+        return encodeURIComponent(str).replace(/%20|[!'()*]/g, function (m) {
+          return m === '%20' ? '+' : '%' + m.charCodeAt(0).toString(16)
```
Mogztter: At first I didn't understand why this was needed and then I read: https://unixpapa.com/js/querystring.html
I think we should explain why `encodeURIComponent` is not enough.

mojavelinux:
Looks like the behavior of CGI.escape changed after Ruby 2.3. I'll need to update the assertion.

lib/asciidoctor/helpers.rb
Mogztter: That's the next next step :+1:

lib/asciidoctor/helpers.rb
mojavelinux: Best would be to link to the following explanation, which is where I got this from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#Description (see last paragraph of section)

lib/asciidoctor/helpers.rb
mojavelinux: :+1: 

No rush. I'm just playing around with some options here. Since we're changing the characters being encoded, the goal will be for it to make it into 2.0.0.

mojavelinux:
What's absolutely hilarious is that there is virtually no agreement between the methods in Ruby and JavaScript about how to encode a URI component string. There are differences between `CGI.escape` and `URI.encode_www_form_component`, there are differences across versions of Ruby, and they all differ from `encodeURIComponent` from JavaScript. But I think what have now provides the most consistency...whatever that is.

PR #3016, opened 2019-01-13
resolves #2106 make syntax highlighter pluggable

### LINKED ISSUE #2106: Add extension point for integrating an alternative source highlighter
The source (aka syntax) highlighter should be an extension point. This will allow users to integrate substitute source highlighters such as [Highlight](http://www.andre-simon.de/doku/highlight/en/highlight.php) or [Highlights](https://www.npmjs.com/package/highlights), which are not supported in core.

The extension should provide callbacks at strategic points when the syntax highlighter needs to operate on the source.

For a server-side highlighter, this includes:

* extracting and restoring callouts
* highlighting the source
* setting attributes on the pre and code elements in the HTML
* generating CSS

For a client-side highlighter, this includes:

* setting attributes on the pre and code elements in the HTML
* linking to CSS
* linking to the JavaScript library
* inserting JavaScript to load the library

Since these requirements are so different, we might actually be looking at two-different extension points. The other option is to only focus on the server-side scenario since the client-side highlighter is so trivial to integrate.

mojavelinux:
resolves #2106

- define the Asciidoctor::SyntaxHighlighter module to handle all aspects of syntax highlighting
- define built-in SyntaxHighlighter adapters for coderay, pygments, highlight.js, html-pipeline, and prettify
- move all embedded syntax highlighting logic into SyntaxHighlighter adapters
- allow SyntaxHighlighter adapter class to self register with a source highlighter name using register_for
- load SyntaxHighlighter adapters lazily (except for highlight.js and html-pipeline)
- add :syntax_highlighters and :syntax_highlighter_factory options to API to allow syntax highlighter resolution to be customized
- instantiate SyntaxHighlighter adapter once header is finalized if source-highlighter attribute is set on document and basebackend is html
- provide SyntaxHighlighter factory modules for registering syntax highlighter adapters (default, custom, and default proxy)
- allow SyntaxHighlighter adapter to contribute docinfo at :head and :footer slots in HTML output document
- move retrieval and writing of stylesheets for syntax highlighters to SyntaxHighlighter adapters
- wire methods in Stylesheets class that pertain to syntax highlighting to SyntaxHighlighter adapters
- remove hard-coded styles in Pygments output
- update default stylesheets to support Pygments styles and drop CSS workarounds
- don't fail if value of pygments-style attribute is not recognized; instead fallback to default style
- don't remove nested pre tags in HTML produced by Pygments when linenums are enabled
- update default stylesheet to avoid styling nested pre tags
- rename pygments table class to pygments-table
- change pygments base CSS selector from .listingblock .pygments to pre.pygments
- allow color for generic text, line numbers, and line number border to inherit from Pygments style
- align stylesheet for default CodeRay style with other syntax highlighters
- set default background color for prettify using a selector with the lowest precedence
- update and add tests
- add tests for the Pygments integration
- add pygments.rb dependency and enable Pygments integration tests if the environment variable PYGMENTS=true is set
- move syntax highlighter tests to dedicated test file

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -40,8 +40,6 @@ module Substitutors
     inline: SUB_GROUPS.keys + NORMAL_SUBS,
   }
 
-  SUB_HIGHLIGHT = ['coderay', 'pygments']
-
   CAN = ?\u0018
   DEL = ?\u007f
 
@@ -67,11 +65,6 @@ module Substitutors
   ESC_R_SB = '\]'
 
   PLUS = '+'
-
-  PygmentsWrapperDivRx = %r(<div class="pyhl">(.*)</div>)m
-  # NOTE handles all permutations of <pre> wrapper
-  # NOTE trailing whitespace appears when pygments-linenums-mode=table; <pre> has style attribute when pygments-css=inline
-  PygmentsWrapperPreRx = %r(<pre\b[^>]*?>(.*?)</pre>\s*)m
 
   # Internal: A String Array of passthough (unprocessed) text captured from this block
   attr_reader :passthroughs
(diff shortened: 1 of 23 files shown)


lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,80 @@
+module Asciidoctor
+  module SyntaxHighlighter
+    PROVIDED = RUBY_ENGINE_OPAL ? {} : {
```
Mogztter: So we have to provide/implement a syntax highlighter in Asciidoctor.js (ie. Highlight.js) ?

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def highlight?; end
+
+    def format node, source, language, opts
+      node.content
+    end
+
+    def write_stylesheet? doc; end
+
+    @lock = ::Mutex.new
```
Mogztter: I think we should "hide" this implementation detail.
Opal does not provide a `Mutex` so the following exception will be thrown `Mutex: uninitialized constant Mutex`.

lib/asciidoctor/substitutors.rb
```diff
@@ -1345,158 +1338,89 @@ def resolve_pass_subs subs
...
-        result = sub_specialchars source
-      end
+    doc_attrs = @document.attributes
+    syntax_hl_name = syntax_hl.name
+    if (linenums_mode = (attr? 'linenums', nil, false) ? (doc_attrs[%(#{syntax_hl_name}-linenums-mode)] || :table).to_sym : nil)
+      start_line_number = 1 if (start_line_number = (attr 'start', nil, 1).to_i) < 1
     end
+    highlight_lines = resolve_lines_to_highlight source, (attr 'highlight') if attr? 'highlight', nil, false
+
+    highlighted, source_offset = syntax_hl.highlight self, source, (attr 'language', 'text', false),
```
Mogztter: `highlight` method is optional ?

Mogztter:
The code is much better and it will also reduce the size of `asciidoctor-core.js` :heart_eyes_cat: :package:

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,80 @@
+module Asciidoctor
+  module SyntaxHighlighter
+    PROVIDED = RUBY_ENGINE_OPAL ? {} : {
+      'coderay' => 'syntax_highlighter/coderay',
+      'highlight.js' => 'syntax_highlighter/highlightjs',
+      'highlightjs' => 'syntax_highlighter/highlightjs',
```
Mogztter: I don't think we should do it in this pull request but we should deprecate `highlightjs` ?

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: You just need to require one of them in the Asciidoctor.js build (since Opal can't require dynamically). But I think I'm going to require highlight.js statically since that's the most common choice and it works across all the platforms. So look for that change.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: Asciidoctor.js will need to provide a basic implementation of Mutex. Here's what would be required.

```
class Mutex
  def synchronize 
    yield
  end
  def owned?
    true
  end
end
```

I may also use this in the converter factory as well (and remove concurrent/hash since it's overkill)

lib/asciidoctor/substitutors.rb
mojavelinux: Correct. Asciidoctor doesn't call `highlight` for client-side syntax highlighters. Instead, it prepares the assets (docinfo) and the real work happens when the page loads. `highlight` is only used when the highlighting is part of the Asciidoctor conversion. The `highlight?` method tells Asciidoctor whether it will be using this method or not.

What's missing on this class are some API docs to explain this.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: Yeah, let's address that separately. It's tricky. I wish the library had an unambiguous name :/

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: I have an idea to avoid the lock on read. (Not that this matters to Asciidoctor.js, which would never need the lock semantics).

lib/asciidoctor/syntax_highlighter.rb
Mogztter: Oh right, I'm silly :grimacing: 

>  But I think I'm going to require highlight.js statically since that's the most common choice and it works across all the platforms. So look for that change.

:+1:

lib/asciidoctor/substitutors.rb
Mogztter: Right I was missing that part `@subs[idx] = :highlight` and then `highlight_source` is only called `@subs` contains `:highlight`.
All good :wink:

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,82 @@
...
+      'pygments' => 'syntax_highlighter/pygments',
+    }
+
+    attr_reader :name
+
+    def initialize doc, opts = {}; end
+
+    def docinfo? location; end
+
+    def highlight?; end
```
Mogztter: Maybe we can avoid this method and just check for `highlight` ?
In other words, I don't think it makes sense to have `highlight? true` without an `highlight` method or `highlight? false` with an `highlight` method, right ?

Pushing this idea a bit further, I don't think we shouldn't allow the user to build "invalid object". I like the flexibility of Ruby but in this case we should maybe define a "trait" (replace with the correct Ruby terminology) for server-side highlighter since the following methods don't make sense in a client-side highlighter:

 *  `highlight`
 * `write_stylesheet?`
 * `library_available?`
 * `write_stylesheet`

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,82 @@
+module Asciidoctor
+  module SyntaxHighlighter
```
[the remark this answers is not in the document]
Mogztter: I'm wondering where we should put the documentation of `highlight` since we don't have an Interface or an Abstract class. Should we need to define something like:

```rb
module ServerSideSyntaxHighlighter 
  # documentation goes here
  def highlight node, source, language, opts
    fail NotImplementedError, "A server-side syntax highlighter must implement #highlight"
  end
end
```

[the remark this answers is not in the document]
mojavelinux: I considered having a separate base class for a build-time highlighter. It's a superset of a client-side highlighter.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: The Ruby way is to make "like a duck" objects. I tend to prefer that informal style over formal interfaces. This is consistent with how we did the extensions too.

What we're really lacking is documentation about how to work with the extension points in Asciidoctor. Once we have that, I think it becomes much more clear.

lib/asciidoctor/syntax_highlighter.rb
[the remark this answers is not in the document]
mojavelinux: > Maybe it's too early and the API is still evolving but at some point we should definitely add documentation on this class.

Of course. It's just taken me all the time I've had up to this point just to get this working. So API documentation is a natural next step, after the tests.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: I can see combining `highlight?` and `library_available?`. There's really no reason to claim to be able to highlight if the library isn't available, and it seems like it exposes too much detail. Either it can highlight or it can't.

mojavelinux: Actually, write_stylesheet? / write_stylesheet could be used by a client-side highlighter. Just because the highlighting happens on the client doesn't mean the stylesheet will automatically be there. That's up to the highlighter implementation.

So I think if we address highlight? / library_available? we end up with a unified API (except for the `highlight` function, which is understandable).

mojavelinux:
I think we're getting pretty close. I'd like to look into whether we can avoid the lazy load of pygments / coderay in Stylesheets (perhaps with a guard). I'm also going to prototype a rouge and prism impl just to make sure we didn't miss anything we'll need.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: :+1:

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: Avoided. So the idea is that only a single attempt is to made to lazy load a syntax highlighter. All other entrants defer.

mojavelinux: @Mogztter Are you willing to put in a stub for Mutex. I'm happy to add a fallback in core if you'd rather it be there.

lib/asciidoctor/substitutors.rb
mojavelinux: Now it should be even clearer since highlight? is determined whether to add the :highlight sub and whether the underlying library is available.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: Since we load it eagerly, we actually defer this problem a bit. I'd still like to align on a single name, though, because it's a pain to have to deal with the exception.

lib/asciidoctor/syntax_highlighter.rb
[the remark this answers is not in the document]
mojavelinux: API docs added!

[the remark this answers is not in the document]
mojavelinux: I also added a NotImplementedError as you suggested in case it is called out of bounds.

lib/asciidoctor/syntax_highlighter.rb
Mogztter: I can put a stub for Mutex in Asciidoctor.js (or Opal) but I just feel bad because the implementation will effectively do nothing. So we just adding a layer of abstraction.
But at the same time the other alternative is to add a special case in Asciidoctor core and that's not good either.

lib/asciidoctor/syntax_highlighter.rb
Mogztter: > Actually, write_stylesheet? / write_stylesheet could be used by a client-side highlighter. Just because the highlighting happens on the client doesn't mean the stylesheet will automatically be there. That's up to the highlighter implementation.

Indeed you're right.

> So I think if we consolidate highlight? / library_available? we end up with a unified API (except for the highlight function, which is understandable).

:+1:

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: There's nothing to feel bad about because JavaScript essentially has a global mutex. As long as the code in JavaScript doesn't yield to IO, it's synchronous.

Having a no-op implementation is just part of having a language that runs in so many different environments. In fact, the concurrent gem does exactly this for Ruby platforms that don't have threads. So providing an implementation for Mutex is an important part of Opal being a complete Ruby.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: :+1:

mojavelinux:
> But we might be able to wire those calls back into the syntax highlighter.

After a lot of research, trial, and error, I finally figured out how to do this. Here's how the implementation looks:

```rb
def coderay_stylesheet_data
  (SyntaxHighlighter.for 'coderay').stylesheet
end
```

lib/asciidoctor/syntax_highlighter/coderay.rb
```diff
@@ -0,0 +1,90 @@
...
+
+      def load_coderay
+        (defined? ::CodeRay::Duo) ? :loaded : (Helpers.require_library 'coderay', true, :warn).nil? ? :unavailable : :loaded
+      end
+    end
+
+    module Styles
+      include Loader
+
+      def stylesheet
```
mojavelinux: @Mogztter Do you like the name `stylesheet` or `stylesheet_data` better? (or something else?)

lib/asciidoctor/syntax_highlighter/coderay.rb
```diff
@@ -0,0 +1,90 @@
...
+        @@stylesheet_cache ||= (::File.read ::File.join DATA_DIR, 'stylesheets', stylesheet_filename).rstrip
+      end
+
+      def stylesheet_filename
```
mojavelinux: @Mogztter Do you like `stylesheet_filename` or `stylesheet_path` better? (or something else)

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    # Public: Format the highlighted source for inclusion in an HTML document.
+    #
+    # node   - The source block being processed.
+    # lang   - The source language for this block (e.g., ruby).
+    # opts   - A Hash of options that control syntax highlighting:
+    #          :nowrap - a Boolean that indicates that wrapping should be disabled
+    #
+    # Returns the highlighted source String wrapped in preformatted tags (e.g., pre and code)
+    def format node, lang, opts
```
mojavelinux: @Mogztter Do you think we should provide a more complete default implementation for format? If not in the SyntaxHighlighter module, perhaps in the Base class? The reason I hadn't done it is because it couples us rather tightly to HTML, but then again so do all the implementations so far. I'm thinking the Base class is the right place for it. The idea would be you could skip this method unless you really needed specific HTML...and we could encourage a standard format.

lib/asciidoctor/syntax_highlighter/coderay.rb
mojavelinux: I like these suggestions, especially stylesheet_basename.

I'm still torn between read_stylesheet and load_stylesheet. However, I'm leading more towards "read" because "load" implies it's either being parsed into an object model or loaded into another object. So I'll start with read_stylesheet and we'll see how it feels.

lib/asciidoctor/syntax_highlighter/coderay.rb
mojavelinux: See other discussion thread.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: I'm going to proceed with putting this default impl in the Base class.

lib/asciidoctor/syntax_highlighter/html_pipeline.rb
```diff
@@ -0,0 +1,10 @@
+module Asciidoctor
+  class SyntaxHighlighter::HtmlPipeline < SyntaxHighlighter::Base
+    register_for 'html-pipeline'
```
Mogztter: I'm not familiar with this "highlighter", is it related to https://github.com/jch/html-pipeline ?

lib/asciidoctor/syntax_highlighter.rb
```diff
@@ -0,0 +1,200 @@
...
+          syntax_hl
+        end
+      end
+
+      private def registry
+        @registry ||= {}
+      end
+    end
+
+    class CustomFactory
```
Mogztter: We should probably add a test with a `CustomFactory` (or a `Factory`)

lib/asciidoctor/syntax_highlighter/html_pipeline.rb
mojavelinux: It sets up the HTML so syntax highlighting will work on GitHub. (It's still a bit of a black box, but that's what it's designed for).

You can verify by making an AsciiDoc file in a GitHub repository that contains the following contents:

```
++++
<pre lang="ruby"><code>puts 'Hello, World!'</code></pre>
++++
```

GitHub only seems to consider the `lang` attribute on the `<pre>` tag. Technically, the `<code>` tag is optional, but I include it for semantic completeness.

lib/asciidoctor/syntax_highlighter.rb
mojavelinux: Good point. I'll add one.

As a quick question, I'm thinking the CustomFactory would be more useful if defined as:

```rb
class CustomFactory
  include Factory

  def initialize registry = nil 
    @registry = registry
  end
end
```

So now you can do:

```rb
factory = Asciidoctor::SyntaxHighlighter::CustomFactory.new name => clazz
```

Then we'd define the DefaultFactoryProxy (which allows the custom factory to fall back to the built-in defaults), as follows:

```rb
class DefaultFactoryProxy < CustomFactory
  include DefaultFactory

  def for name
    (@registry.key? name) ? @registry[name] : super
  end
end
```

wdyt?

mojavelinux:
We are sooooo close!

lib/asciidoctor/syntax_highlighter.rb
Mogztter: > As a quick question, I'm thinking the CustomFactory would be more useful if defined as

Why not, as far as I understand it's a shortcut:

```rb
syntax_hl_factory = Asciidoctor::SyntaxHighlighter::CustomFactory.new
syntax_hl_factory.register clazz, name

# shortcut
syntax_hl_factory = Asciidoctor::SyntaxHighlighter::CustomFactory.new name => clazz
```

> Then we'd define the DefaultFactoryProxy (which allows the custom factory to fall back to the built-in defaults), as follows:

But we could define a custom strategy to resolve the syntax highlighter in the CustomFactory ?
So the `DefaultFactoryProxy` should first check that the `CustomFactory.for name` does not return something and then fallback to the built-in no ? Or maybe I'm reading backward...

mojavelinux: > So the DefaultFactoryProxy should first check that the CustomFactory.for name does not return something and then fallback to the built-in no ?

That's exactly what it does. But it also allows the entry to be nil (which is why we check the key).

Mogztter: `DefaultFactory` defines a `for` method and `DefaultFactoryProxy` includes `DefaultFactory` but overrides the method so `super` will call `DefaultFactory.for` right ?
But now you also extend `CustomFactory` which can define a `for` method too ? And now I'm lost :laughing:

mojavelinux: The include supersedes the extend. So `super` is DefaultFactory.for. All the extend does is give us the constructor (and the type). (In the case of Opal, `super` will jump to Factory.for, so everything still works).

Mogztter: If I execute the code it will make things easier to understand but on screen I'm like 🤯

mojavelinux:
Here's an earlly prototype of the Prism integration: https://gist.github.com/mojavelinux/850373bb1b5a8e0334b4038e90617509

There are plenty of options still to enable, but that should give you a picture of how it works and what's possible.

Mogztter:
Asciidoctor.js build against this branch :tada:  but one test is failing :disappointed: 
When using a missing syntax highlighter a warning message was printed. For instance:

```
:source-highlighter: coderay

[source, bash]
----
ls -la
----
```

And the warning message was:

```
optional gem 'coderay' is not installed. Functionality disabled.
```

The above warning message does not make any sense in a JavaScript environment but at least you know that `coderay` is not available (and quickly understand why your code is not highlighted).
So I think we should add a generic message:

```
syntax highlighter named 'coderay' is not available, please use: 'highlight.js' or 'prism.js'.
```

mojavelinux:
If you want that behavior, you simply need to register an implementation that emits that message when the `highlight?` method is invoked.

I don't think it's really necessary, though. If the source highlighter is not recognized, then it should just silently ignore it. That's true for any unrecognized name right now. We just had a special case with coderay and pygments because it was hard-wired into the core before.

Good to hear the code works with Asciidoctor.js! Thanks for verifying.

mojavelinux:
I see your point. But that's what I think the extension mechanism affords you. You now have the ability to provide your own factory, which means you can decide how to handle an unresolved highlighter. So you can tailor it to your needs. But for core, I see the source highlighter as a suggestion. It's not a warning if it's not there. Otherwise, a lot of documents would all of a sudden start producing warnings and we would have a lot of people complaining (since there's little overlap in which source highlighters different environments support).

mojavelinux:
And we're merged!

Mogztter:
> And we're merged!

Yahoooo :tada: :tada: :tada: 

> I see your point. But that's what I think the extension mechanism affords you. You now have the ability to provide your own factory, which means you can decide how to handle an unresolved highlighter. 

Alright I will :+1:

mojavelinux:
> You now have the ability to provide your own factory

By you, I meant a developer integrating Asciidoctor into their application.

PR #3025, opened 2019-01-17
resolves #3020 don't hard code shaded color in source block

### LINKED ISSUE #3020: Don't override color of generic text in source block as specified by syntax highlighter theme
There are several places in the default stylesheet where a hard-coded color (namely `#999`) is specified inside a source block. Typically, these assignments are for shading line numbers and borders. Instead of a hard-coded color, the stylesheet should assign an opacity so the color gets blended with the theme background. This can be accomplished using opacity and currentColor.

mojavelinux:
- don't hard code shaded color in source block so it works with different themes/styles


mojavelinux:
This needs to be rebased against the changes from #3016 before being merged.

PR #3033, opened 2019-01-26
resolves #1040 add syntax highlighter adapter for Rouge

### LINKED ISSUE #1040: Integrate Rouge for source highlighting

mojavelinux:
resolves #1040

- add SyntaxHighlighter adapter implementation for Rouge
- change CSS class of line numbering table for Pygments and Rouge to linenotable
- reuse styles from default stylesheet to layout the line number column
- remove nested `<pre>` wrapper in Pygments output when line numbering is enabled and linenums mode is inline
- ensure syntax highlighter library is loaded when looking up base style
- add the rouge gem as a development dependency

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -20,6 +20,7 @@ Enhancements::
   * drop support for Ruby < 2.3 and JRuby < 9.1 and remove workarounds (#2764)
   * drop support for Slim < 3 (#2998)
   * make syntax highlighter pluggable; extract all logic into adapter classes (#2106)
+  * add syntax highlighter adapter for Rouge (#1040)
   * add support for start attribute when using prettify to highlight source blocks with line numbering enabled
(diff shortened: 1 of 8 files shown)


cup:
https://github.com/github/markup/issues/1258

cup:
https://github.com/asciidoctor/asciidoctor.org/issues/838

PR #3035, opened 2019-01-27
Switch to squiggly heredoc in test suite & CLI

mojavelinux:


mojavelinux:
This change will be broken up into multiple commits. This PR was to verify that this change can be safely made.

Issue #3005, opened 2019-01-09
Drop DocBook 4.5 converter


mojavelinux:
I'd like to drop the built-in DocBook 4.5 converter going into Asciidoctor 2.0.

Keep in mind, this is just about dropping the converter as a **built-in** feature of core. Asciidoctor allows converters to be developed and loaded independently. So if there's still a strong need to generate DocBook 4.5 using Asciidoctor 2.0, it's possible to take the code from 1.5.8 and create a dedicated converter out of it (perhaps which still extends the DocBook 5 converter).

We just won't be shipping the DocBook 4.5 converter with the core gem any longer.

mojavelinux:
Keep in mind the DocBook 4.5 specification is over 12 years old, and DocBook 5 became a specification over 10 years ago.

PR #3038, opened 2019-02-01
resolves #2891 redesign converter API

### LINKED ISSUE #2891: Converter::Factory should support resolving a built-in converter
Currently, the converter factory only resolves custom converters. This presents a problem for integrations that need to access or extend a built-in converter. Add a flag to the resolve method (or introduce a dedicated method) that will resolve a built-in converter (honoring the order of precedence).

mojavelinux:
resolves #2891

- redesign converter API based on design of syntax highlighter API
- fold default converter factory (static / global) into Converter
- make registration and resolution of global converters thread-safe
- allow converter factory to be passed using :converter_factory API option
- provide forgiving default implementation for convert method in Base converter
- change handles? method on Converter module to always return true
- move current implementation of handles? method from Converter module to Base converter
- add _content_only and _skip helper methods to Base converter
- warn if a handler method for a node is not found in a subclass of the Base converter
- rename BackendInfo module to BackendTraits; map backend_info method to backend_traits
- allow computation of backend traits to be disabled using init_backend_traits
- introduce non-static CustomFactory
- map old (deprecated) public API to new API
- reconfigure built-in (aka provided) converters to use Base converter; set backend traits
- remove BuiltIn converter
- add backend_traits_source keyword argument to CompositeConverter constructor
- honor htmlsyntax set by converter
- move Writer to dedicated source file
- remove unused convert_with_options method on Converter
- add and update tests
- prefix internal helper methods in converters with underscore
- make write_alternate_pages a static method of the manpage converter
- delineate private sections of the manpage converter


lib/asciidoctor/converter.rb
```diff
@@ -1,225 +1,394 @@
...
+    # node as its sole argument. Implementations are free to override this method to provide different behavior.
     #
-    # Returns the [String] result
+    # See {Converter#convert} for details about the arguments and return value.
     def convert node, transform = nil, opts = nil
-      raise ::NotImplementedError
+      opts ? (send transform || node.node_name, node, opts) : (send transform || node.node_name, node)
+    rescue
+      if ::NoMethodError === (ex = $!) && ex.receiver == self && ex.name.to_s == (transform || node.node_name)
+        raise ::NotImplementedError, %(#{self.class} (backend: #{@backend}) must implement the ##{ex.name} method)
```
mojavelinux: @Mogztter Do you think a missing convert delegate method should be an exception or a warning? (essentially the behavior of _skip_with_warning).

Mogztter: It depends :wink: 

In my opinion it's easy to miss a warning message and my personal taste is to "fail fast".
But if I don't expect someone to use a given node (for instance a `listing` node because we agree that we won't use them) then I'd probably prefer a warning. In this case we decide that the converter won't implement the `listing` method so the person who use a `listing` block will (hopefully) notice that the block is missing, read the warning message and use another syntax.
In other words, severity can be relative depending on your context. 

A warning message seems more flexible but we "lower" the severity from error (you just can't continue unless you fix it) to warning. But since we can configure `--failure-level=WARN`, I guess it's better if we issue a warning... :thinking:

lib/asciidoctor/converter.rb
```diff
@@ -1,225 +1,394 @@
...
+            @@registry[backend]
+          end : catch_all
         end
       end
 
-      def supports_templates
-        backend_info[:supports_templates] = true
-      end
+      PROVIDED = {
+        'docbook45' => %(#{__dir__}/converter/docbook45),
```
Mogztter: DocBook 4.5 is still a thing ? https://github.com/asciidoctor/asciidoctor/issues/3005
I believe it's the next step but maybe we can add a notice to express deprecation ?

lib/asciidoctor/converter.rb
```diff
@@ -1,225 +1,394 @@
...
-        if value
-          backend_info[:basebackend] = value
+    private
+
+    attr_reader :registry
+  end
+
+  # Mixed into the {Converter} module to provide the global registry of converters that are registered statically.
+  #
+  # This registery includes built-in converters for {Html5Converter HTML 5}, {DocBook5Converter DocBook 5} and
```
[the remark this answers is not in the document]
mojavelinux: Nice catch!

lib/asciidoctor/converter.rb
mojavelinux: On the way. That will happen after this gets merged.

lib/asciidoctor/converter.rb
mojavelinux: I'm thinking perhaps an error message is most appropriate. The reason I don't think we should fail is that the purpose of this method is to help simplify the development of a converter. But it's not very development friendly if it keeps failing for every unimplemented node. Once a converter is done, this should never happen. So I think that tells us what we should do here.

lib/asciidoctor/converter/factory.rb
```diff
@@ -1,231 +0,0 @@
...
-      # or nil if no converter is resolved.
-      #
-      # backend - the String backend name
-      # opts    - an optional Hash of options that get passed on to the converter's
-      #           constructor. If the :template_dirs key is found in the options
-      #           Hash, this method returns a {CompositeConverter} that delegates
-      #           to a {TemplateConverter}. (optional, default: {})
-      #
-      # Returns the [Converter] object
-      def create backend, opts = {}
```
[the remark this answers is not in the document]
mojavelinux: Yep. There were too many abstractions and it was breaking my mind. I think it's much easier to understand now without all the delegation.

lib/asciidoctor/converter.rb
Mogztter: I agree 👍

mojavelinux: Just pushed that change. I'm going to start with a warning since that's what we were using before in other converters.

PR #3045, opened 2019-02-05
resolves #3029 delegate assignment of datetime attributes to a method

### LINKED ISSUE #3029: Delegate date/time assignments to a method
The goal is to abstract the document date resolution and attributes assignments when a `Document` is instantiated. And most specifically the following code:

https://github.com/asciidoctor/asciidoctor/blob/e6ff47641a0e4bcc5d9e5271eaaf5b6389f5cacc/lib/asciidoctor/document.rb#L490-L506

It will then be easier for Asciidoctor.js to provide the most accurate implementation.

Mogztter:


mojavelinux:
I made some updates. I think it's good.

The only open question at this point is what to do with the assignments in Asciidoctor.load.

mojavelinux:
Oh! I see what we can do, and this will actually help with other uses of the API too. We instead of computing the doc* datetime attributes in load, we should pass the input_mtime as an option to the Document constructor. That way, we can wait and compute those when the Document is created, when we are computing the local datetime attributes.

mojavelinux:
PR updated. Let me know what you think.

Mogztter:
Oh right, I didn't saw the code duplication when input is a File.

> Instead of computing the doc* datetime attributes in load, we should pass the input_mtime as an option to the Document constructor. That way, we can wait and compute those when the Document is created, when we are computing the local datetime attributes.

I like it :+1:

mojavelinux:
Excellent. Let's march forward then!

Issue #2876, opened 2018-09-20
Split --quiet flag into --quiet and --silent


mojavelinux:
Currently, the `--quiet` CLI flag silences all warnings **and errors**. However, this is really what most commands call `--silent`. Change the `--quiet` flag to only suppress info and warnings. Then introduce a `--silent` flag that is absolutely silent, giving no indication of success or failure other than the exit code.

PR #3046, opened 2019-02-05
resolves #3030 separate the -v and -w CLI flags

### LINKED ISSUE #3030: Add -w option flag to CLI
The `-w` option flag in Ruby enables language warnings. It's customary for Ruby-based CLIs, like the one in Asciidoctor, to pass this option flag through to the language.

mojavelinux:
- use -v to set logger level to debug
- use -w to enable pedantic Ruby warnings (aka $VERBOSE = true)
- log possible invalid reference warnings at debug level

Note that -q still sets $VERBOSE = nil and is thus mutually exclusive with -w (in fact it overrides it).

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -43,6 +43,7 @@ Enhancements::
   * when using a server-side syntax highlighter, highlight content of source block even if source language is not set (#3027)
   * remove the 2-character (i.e., `""`) quote block syntax
   * don't allow block role to inherit from document attribute; only look for role in block attributes (#1944)
+  * split out functionality of -w CLI flag (script warnings) from -v CLI flag (verbose logging) (#3030)
 
 Improvements::
 
@@ -57,8 +58,12 @@ Improvements::
   * ensure linenos class is added to linenos column when source highlighter is pygments and pygments-css=style
   * rename CSS class of Pygments line numbering table to linenotable (to align with Rouge) (#1040)
   * remove unused Converter#convert_with_options method (#2891)
+<<<<<<< HEAD
   * don't store the options attribute on the block once the options are parsed (#3051)
   * add an options method on AbstractNode to retrieve the set of option names (#3051)
+=======
+  * log possible invalid references at debug level (#3030)
+>>>>>>> d2a3e87d... resolves #3030 separate the -v and -w CLI flags
 
 Bug Fixes::
 
--- a/man/asciidoctor.1
+++ b/man/asciidoctor.1
@@ -2,12 +2,12 @@
 .\"     Title: asciidoctor
 .\"    Author: Dan Allen, Sarah White, Ryan Waldron
 .\" Generator: Asciidoctor 2.0.0.dev
-.\"      Date: 2019-02-04
+.\"      Date: 2019-02-09
 .\"    Manual: Asciidoctor Manual
 .\"    Source: Asciidoctor 2.0.0.dev
 .\"  Language: English
 .\"
-.TH "ASCIIDOCTOR" "1" "2019-02-04" "Asciidoctor 2.0.0.dev" "Asciidoctor Manual"
+.TH "ASCIIDOCTOR" "1" "2019-02-09" "Asciidoctor 2.0.0.dev" "Asciidoctor Manual"
 .ie \n(.g .ds Aq \(aq
 .el       .ds Aq '
 .ss \n[.ss] 0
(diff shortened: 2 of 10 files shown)


mojavelinux:
@Mogztter wdyt? it feels right to separate `-w` and `-v`. Imagine turning on `-v` and seeing the pedantic warnings from all third-party libraries, such as rouge. Thata's the case today. To me, `-v` turns on info and debug messages. `-w` turns on pedantic warnings. And these should be separate. Am I on track?

Mogztter:
@mojavelinux I'm not familiar with the `-w` flag in Ruby but as far as I understand it will enable Ruby warnings like:

```
/path/to/rouge/lexers/shell.rb:20: warning: already initialized constant Rouge::Lexers::Shell::KEYWORDS
```

If that's the case then yes I think we should have two separated flags:

* `-v`: turns on verbose mode (ie. "application" messages)
* `-w`: turns on pedantic warnings (ie. "system" messages)

Maybe we should mention that `-w` enables system/runtime/language/Ruby warnings so it's clear that it won't conflict with application messages (and vice-versa).

man/asciidoctor.1
```diff
@@ -178,23 +178,27 @@ If this option is not set (default: FATAL), the program exits with a status code
 .sp
 \fB\-q, \-\-quiet\fP
 .RS 4
-Silence warnings.
+Silence all messages.
```
Mogztter: all application messages ? or all messages ? even if I enable the pedantic warnings ?

mojavelinux: Perhaps we should just say "no output".

Though, there's another open issue that proposed to break this down even further. See https://github.com/asciidoctor/asciidoctor/issues/2876.

mojavelinux:
> as far as I understand it will enable Ruby warnings like:

That's it exactly.

> Maybe we should mention that `-w` enables system/runtime/language/Ruby warnings so it's clear that it won't conflict with application messages (and vice-versa).

That's what "pedantic warnings" means. It's a reference to gcc. I go into more detail in the man page, where I say "warnings in executed code".

man/asciidoctor.1
Mogztter: Indeed it's a `silent` mode. Until we split the `--quiet` option, we could add a "no output":

> Silence all messages. (no output)

Mogztter:
> That's what "pedantic warnings" means. It's a reference to gcc. 

I didn't know that... Also in French "pedantic" (pédant) is pejorative. It means explaining something in a pretentious way. 

> I go into more detail in the man page, where I say "warnings in executed code".

Alright, I think we could be a bit more specific in the `--help` too because not everyone will get the gcc reference. What do you think ?

man/asciidoctor.1
mojavelinux: :+1: 

I'm going to go with "Don't output messages" since technically you can still output the document to STDOUT.

mojavelinux:
> in French "pedantic" (pédant) is pejorative. It means explaining something in a pretentious way.

It's kind of that way in English too. It means to harp on or to nag. That's essentially what the Ruby interpreter is doing. In this mode, it complains about stuff that may not matter, but it's going to bug you anyway.

I'll keep it simple and say either "code warnings" or "script warnings".

mojavelinux:
I think I came up with something that's going to work.

man/asciidoctor.1
```diff
@@ -173,43 +173,47 @@ Matching templates found in subsequent directories override ones previously disc
...
+.sp
+\fB\-w, \-\-warnings\fP
+.RS 4
+Turn on script warnings (applies to executed code).
 .RE
 .sp
 \fB\-t, \-\-timings\fP
 .RS 4
-Display timings information (time to read, parse and convert).
+Print timings report to stderr (time to read, parse and convert).
```
Mogztter: You mean `stdout` ? https://github.com/asciidoctor/asciidoctor/blob/eb251b6e704396219f3cb184396cc2ec1a641e6f/lib/asciidoctor/timings.rb#L49

mojavelinux: Nope, the CLI sends the report to stderr (so it doesn't get mixed up with the converted document going to stdout).

Mogztter: Just checked, it's `stderr`

CHANGELOG.adoc
```diff
@@ -57,8 +58,12 @@ Improvements::
   * ensure linenos class is added to linenos column when source highlighter is pygments and pygments-css=style
   * rename CSS class of Pygments line numbering table to linenotable (to align with Rouge) (#1040)
   * remove unused Converter#convert_with_options method (#2891)
+<<<<<<< HEAD
```
Mogztter: :cactus:

mojavelinux: Thanks! Fixed.

PR #3063, opened 2019-02-10
resolves #3059 change default value of fallback attribute name on attr/attr? to nil

### LINKED ISSUE #3059: AbstractNode#attr and AbstractNode#attr? should not inherit by default
Falling back to an attribute on the document is more of an exception than the rule, and when it does happen, the fallback name usually needs to be specified (otherwise, the name it looks for is very generic). Therefore, the default value for the third argument to AbstractNode#attr and AbstractNode#attr? (fallback_name) should be nil. This should simplify a lot of calls into these methods where this behavior must be explicitly disabled (and only a few where it needs to be reenabled).

mojavelinux:
resolves #3059

- change default value of fallback attribute name on attr/attr? to nil
- look for table-frame, table-grid, and table-stripes attributes on document as fallback fro frame, grid, and stripes on table

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -46,6 +46,8 @@ Enhancements::
   * split out functionality of -w CLI flag (script warnings) from -v CLI flag (verbose logging) (#3030)
   * log possible invalid references at debug level (#3030)
   * use the third argument of AbstractNode#attr / AbstractNode#attr? to set the name of a fallback attribute to look for on the document (#1934)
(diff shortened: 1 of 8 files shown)


mojavelinux:
I wonder if we should look for `source-linenums` on the document if `linenums` is not set on the block. That would make it easy to turn on linenums globally. This would be similar to `source-language`.

mojavelinux:
Actually, we already have `source-linenums-option`, so I think we're covered.

PR #3066, opened 2019-02-10
Resolves #2840, assign id and role on list items

### LINKED ISSUE #2840: HTML converter should transmit id and role on list items

Mogztter:
resolves #2840

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -650,7 +650,9 @@ Your browser does not support the audio tag.
     result << %(<ol class="#{node.style}"#{type_attribute}#{start_attribute}#{reversed_attribute}>)
 
     node.items.each do |item|
-      result << '<li>'
+      id_attr = item.id ? %( id="#{item.id}") : ''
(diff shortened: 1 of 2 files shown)


mojavelinux:
I've actually been sitting on a PR to fix this for awhile. See #2841. The reason I haven't merged it yet is that I need to understand how it impacts performance (and optimize if possible). I'd be interested in the extensions test you wrote. Could you submit it to that other branch?

Mogztter:
> I've actually been sitting on a PR to fix this for awhile. See #2841. The reason I haven't merged it yet is that I need to understand how it impacts performance (and optimize if possible).

@mojavelinux Silly me I missed that 😬 

> I'd be interested in the extensions test you wrote. Could you submit it to that other branch?

Sure 👍

Mogztter:
Closed in favor of #2841

PR #3075, opened 2019-02-15
enforce and report relative max include depth properly

mojavelinux:
- enforce and report relative max include depth properly
- depth=0 on include directive disables nested includes (instead of depth=1)
- rename PreprocessorReader#exceeded_max_depth? to PreprocessorReader#exceeds_max_depth?
- change PreprocessorReader#exceeds_max_depth? to return nil if includes are disabled
- add api doc to PreprocessorReader#exceeds_max_depth?

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -73,6 +73,7 @@ Improvements::
   * pass :input_mtime option to Document constructor; let Document constructor assign docdate/time/year attributes (#3029)
   * never mutate strings; add a `frozen_string_literal: true` magic comment to top of all Ruby source files (#3054)
   * always use docdate and doctime to compute docyear and docdatetime (#3064)
+  * rename PreprocessorReader#exceeded_max_depth? to PreprocessorReader#exceeds_max_depth? and return nil if includes are disabled
 
 Bug Fixes::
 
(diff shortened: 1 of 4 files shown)


mojavelinux:
@Mogztter This one's a little weird. Either due to a programming error or a misguided attempt to reproduce behavior in AsciiDoc Python, the following snippet would disable nested includes:

```
include::include-me.adoc[depth=1]
```

And if you had instead set the depth to 0, you'd get the message:

```
maximum include depth of 1 exceeded
```

If the depth on the include directive is 1, you would expect that it would allow one level of includes. If the depth is 0, you would expect that it would allow no includes. So that's what I've changed the behavior to.

I also discovered it was possible to break out of the max-include-depth limit by setting the depth attribute on the include directive. The max-include-depth should always be enforced as an absolute upper limit. So now it is.

Mogztter:
Sorry I missed this notification.

> If the depth on the include directive is 1, you would expect that it would allow one level of includes. If the depth is 0, you would expect that it would allow no includes. So that's what I've changed the behavior to.

Indeed this behavior makes much more sense.

> I also discovered it was possible to break out of the max-include-depth limit by setting the depth attribute on the include directive. The max-include-depth should always be enforced as an absolute upper limit. So now it is.

Oh! 
I just crashed my browser with:

**root.adoc**
```
include::child.adoc[depth=2]
```

**file.adoc**
```
include::root.adoc[depth=2]
```

mojavelinux:
Yep. It's a sneaky one!

PR #3090, opened 2019-02-20
resolves #3089 fix a nil exception when using two consecutive inline macros

### LINKED ISSUE #3089: `block in restore_passthroughs': undefined method `[]' for nil:NilClass (NoMethodError)
#### Reproduction case

```
echo "pass:m[a] stem:[b]" |  ./bin/asciidoctor - --trace
```
```
/path/to/asciidoctor/lib/asciidoctor/substitutors.rb:319:in `block in restore_passthroughs': undefined method `[]' for nil:NilClass (NoMethodError)
	from /path/to/asciidoctor/lib/asciidoctor/substitutors.rb:316:in `gsub'
	from /path/to/asciidoctor/lib/asciidoctor/substitutors.rb:316:in `restore_passthroughs'
	from /path/to/asciidoctor/lib/asciidoctor/substitutors.rb:112:in `apply_subs'
	from /path/to/asciidoctor/lib/asciidoctor/block.rb:112:in `content'
	from /path/to/asciidoctor/lib/asciidoctor/converter/html5.rb:716:in `paragraph'
	from /path/to/asciidoctor/lib/asciidoctor/converter.rb:376:in `convert'
	from /path/to/asciidoctor/lib/asciidoctor/abstract_block.rb:75:in `convert'
	from /path/to/asciidoctor/lib/asciidoctor/abstract_block.rb:84:in `block in content'
	from /path/to/asciidoctor/lib/asciidoctor/abstract_block.rb:84:in `map'
	from /path/to/asciidoctor/lib/asciidoctor/abstract_block.rb:84:in `content'
	from /path/to/asciidoctor/lib/asciidoctor/document.rb:988:in `content'
	from /path/to/asciidoctor/lib/asciidoctor/converter/html5.rb:176:in `document'
	from /path/to/asciidoctor/lib/asciidoctor/converter.rb:376:in `convert'
	from /path/to/asciidoctor/lib/asciidoctor/document.rb:919:in `convert'
	from /path/to/asciidoctor/lib/asciidoctor.rb:1463:in `convert'
	from /path/to/asciidoctor/lib/asciidoctor/cli/invoker.rb:106:in `invoke!'
	from ./bin/asciidoctor:15:in `<main>'
```

Mogztter:
resolves #3089

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -322,9 +322,6 @@ module Substitutors
       end
       subbed_text.include?(PASS_START) ? restore_passthroughs(subbed_text, false) : subbed_text
     end
-  ensure
-    # free memory if in outer call...we don't need these anymore
-    passes.clear if outer
   end
 
   # Public: Substitute quoted text (includes emphasis, strong, monospaced, etc.)
--- a/test/substitutions_test.rb
+++ b/test/substitutions_test.rb
@@ -34,6 +34,12 @@ context 'Substitutions' do
       assert_equal %(this<br>\nis<br>\n&#8594; Asciidoctor<br>\n<br>\n), result
     end
 
+    test 'apply_subs should evaluate two inline macros' do
+      para = block_from_string %(pass:m[a] stem:[b])
+      result = para.apply_subs(para.lines)
+      assert_equal [%(a \\$b\\$)], result
+    end
+
     test 'should expand subs passe to expand_subs' do
       para = block_from_string %({program}\n*bold*\n2 > 1)
       para.document.attributes['program'] = 'Asciidoctor'


lib/asciidoctor/substitutors.rb
```diff
@@ -322,9 +322,6 @@ def restore_passthroughs text, outer = true
       end
       subbed_text.include?(PASS_START) ? restore_passthroughs(subbed_text, false) : subbed_text
     end
-  ensure
-    # free memory if in outer call...we don't need these anymore
-    passes.clear if outer
```
Mogztter: Since the comment say that we don't need to clear `passes` anymore I've removed it to fix this issue.

mojavelinux: I did a lot more digging into this issue to really understand it. While this seems like the straightforward fix, the problem is it allows the passthroughs Hash continue to grow throughout the lifetime of the document (each time substitutions are applied). This isn't significant for a one-time conversion, but it does become a problem if the document is being reused.

The problem we're observing stems from the nested call to apply_subs. It doesn't get the memo that it's nested. After trying out several approaches, I'm going to propose moving the outer/nested flag to apply_subs where we can have greater control over tracking where we are.

mojavelinux:
Superseded by #3091.

PR #3091, opened 2019-02-20
resolves #3089 fix crash caused by an inline passthrough with the macros sub

### LINKED ISSUE #3089: `block in restore_passthroughs': undefined method `[]' for nil:NilClass (NoMethodError)

mojavelinux:
resolves #3089

- fix crash caused by inline passthrough macro with the macros sub clearing the remaining passthrough placeholders

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -52,6 +52,8 @@ module Substitutors
   # EPA, end of guarded protected area (\u0097)
   PASS_END = ?\u0097
 
+  PASS_UNRESOLVED = { text: '??pass??' }
+
   # match passthrough slot
   PassSlotRx = /#{PASS_START}(\d+)#{PASS_END}/
 
@@ -71,11 +73,12 @@ module Substitutors
 
   # Public: Apply the specified substitutions to the text.
   #
-  # text  - The String or String Array of text to process; must not be nil.
-  # subs  - The substitutions to perform; must be a Symbol Array or nil (default: NORMAL_SUBS).
+  # text   - The String or String Array of text to process; must not be nil.
+  # subs   - The substitutions to perform; must be a Symbol Array or nil (default: NORMAL_SUBS).
+  # nested - A Boolean indicating whether we're in a nested substitution (internal use only) (default: nil)
   #
   # Returns a String or String Array to match the type of the text argument with substitutions applied.
-  def apply_subs text, subs = NORMAL_SUBS
+  def apply_subs text, subs = NORMAL_SUBS, nested = false
     return text if text.empty? || !subs
 
     if (is_multiline = ::Array === text)
@@ -109,7 +112,7 @@ module Substitutors
         logger.warn %(unknown substitution type #{type})
       end
     end
-    text = restore_passthroughs text if has_passthroughs
+    text = restore_passthroughs text, nested if has_passthroughs
 
     is_multiline ? (text.split LF, -1) : text
   end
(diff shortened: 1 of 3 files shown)


Mogztter:
resolves #3089 (to link the issue and the pull request)

mojavelinux:
I like this new strategy. It's less API invasive...and it works with Opal (as far as I can tell).

mojavelinux:
> (to link the issue and the pull request)

it drives me nuts that GitHub requires that you put the issue number in the comment instead of the title.

lib/asciidoctor/substitutors.rb
```diff
@@ -303,28 +321,19 @@ def extract_inner_passthrough text, pre, attributes = nil
...
     passes = @passthroughs
-    # passthroughs may have been eagerly restored (e.g., footnotes)
-    #if outer && (passes.empty? || !text.include?(PASS_START))
-    #  return text
-    #end
-
     text.gsub PassSlotRx do
       # NOTE we can't remove entry from map because placeholder may have been duplicated by other substitutions
-      pass = passes[$1.to_i]
+      pass = passes[$1.to_i] || PASS_UNRESOLVED
```
Mogztter: I don't really like this placeholder value. Is it to avoid code duplication ? 
I guess we could replace it with a condition ? if `passes` is defined then do X else to do Z ?

mojavelinux: This should never happen. But in the extreme case that it does (perhaps a misbehaving extension), we just want to avoid a crash. And because it's such an exception, I don't want to waste any time checking if the value is there. (We don't have to use a constant...we could just inline the value).

Mogztter:
> I like this new strategy. It's less API invasive...and it works with Opal (as far as I can tell).

Indeed and the state is "encapsulated" (not just a dangling boolean param)

mojavelinux:
Precisely.

lib/asciidoctor/substitutors.rb
mojavelinux: Is your issue with the constant or the fallback? If it's just a constant, I can inline it.

I prefer the fallback right now because we need to stop being bitten by this crash and this solves it with the least amount of overhead.

Mogztter: It's both. 
As you've mentioned it's unlikely that `passes[$1.to_i]` will equal `{ text: '??pass??' }` but if it happens we won't know for sure if it's the fallback value or not (but that's not my biggest concern)

> And because it's such an exception, I don't want to waste any time checking if the value is there.

Behind the scene Ruby will check if the value is there right ? When you say "I don't want to waste any time checking" you mean at runtime ?
I think it should maybe be a bit faster since we don't need to check `if (type = pass[:type])` if `passes[$1.to_i]` is nil.
Also since we know that `pass[:subs]` will be undefined maybe we can skip this call `subbed_text = apply_subs(pass[:text], pass[:subs])`.

My point is that we execute the same code path but it fells like we should not.

```rb
if (pass = passes[$1.to_i])
  # do X
else
  # do Z 
end
```

mojavelinux:
The last implementation had a measurable impact on performance. I came up with another way by moving the lock variable to the current node. I can confirm this solution has negligible or no impact on performance. (This is a very touchy part of the code).

lib/asciidoctor/substitutors.rb
mojavelinux: Let me text perf. As I mentioned, this is tough because this code is just hammered to death.

mojavelinux: Done. And you'll like that we even log an error now ;)

PR #3095, opened 2019-02-22
update extension DSL method aliases to use infinitive form of verbs

mojavelinux:


mojavelinux:
The current method is not parallel with the other DSL methods.

Consider this group:

* named
* bound_to
* positional_attrs
* parses_content_as

One is not like the others. `positional_attrs` should be an action. Hence `maps_positional_attributes_to`. For example:

```
maps_positional_attributes_to 'attribution', 'citetitle'
```

We might also consider:

* maps_unnamed_attributes
* maps_unnamed_attributes_to
* maps_positional_attrs
* maps_positional_attributes_as

What feels most natural?

mojavelinux:
We really need to pick a direction here so there aren't a half-dozen ways to write each configuration parameter.

Mogztter:
My block processor is named yell, bound to paragraph, with positional attributes named attribution and citetitle and its content is parsed as simple.

> Though perhaps it reads better as:

It may be less syntactically correct but infinitive verbs might be easier for non native speaker.

> We really need to pick a direction here so there aren't a half-dozen ways to write each configuration parameter.

💯
Definitely we should use a single form.

mojavelinux:
I mostly decided to keep them the same, but deprecate all the non-infinitive forms. I also added `match_format` (to replace the deprecated `using_format`), dropped `matches` and `matching`, and dropped `name_attributes`.

Mogztter:
Sounds reasonable 👌

PR #3098, opened 2019-02-24
introduce helper method in parser to detect uniform line

mojavelinux:


mojavelinux:
cc: @Mogztter 

If this ends up improving the performance in Opal, I'll also introduce a method to count newlines, which is the other place we're using count in core.

Mogztter:
> If this ends up improving the performance in Opal, I'll also introduce a method to count newlines, which is the other place we're using count in core.

@mojavelinux confirmed! https://github.com/asciidoctor/asciidoctor.js/pull/669#issuecomment-467052586

mojavelinux:
@Mogztter Thanks for testing!

I really like this method because it makes the intent much more clear.

mojavelinux:
Oh, before I merge, I have a question. Can you implement the Opal-specific code into Asciidoctor.js (since it's just an optimization)? If so, I'll remove it from this PR.

mojavelinux:
Btw, I thought about making the signature more lazy, like:

```
def self.uniform? str, chr = str.chr, len = string.length
  ...
end
```

However, since this is a low-level method, I decided against it since it has a small performance price and yet never used.

mojavelinux:
Here's the code that needs to go over to Asciidoctor.js for this:

```rb
module Asciidoctor
module Parser
  if `String.prototype.repeat`
    def self.uniform? str, chr, len
      `chr.repeat(len) === str`
    end
  else
    def self.uniform? str, chr, len
      `Array.apply(null, { length: len }).map(function () { return chr }).join('') === str`
    end
  end
end
end
```

It's up to you whether to support IE (which is the else block).

PR #3099, opened 2019-02-25
resolves #2884 add clearer exception message when input data is binary or has invalid encoding

### LINKED ISSUE #2884: Encoding normalization failure?
One of the first things I immediately saw when starting to test asciidoctor is that it fails to deal with malformed binary data. I don't know if asciidoctor intends to deal with such input at all, but it throws an unhandled `Encoding::InvalidByteSequenceError` Exception when it fails to deal with the input encoding. I think it would be correct to either prevent this from happening or properly handle the exception.

I only examined the first crashes I got in which the exception is raised from `normalize_lines_array` in `helpers.rb`. I did not see any crashes in `normalize_lines_from_string` but my hunch says it also needs some care. 

Since the reference crash files are binary [I uploaded them to gist](https://gist.github.com/zelivans/74bc642add4829975e2e1f77025379e4#file-id-000001-sig-10-src-000076-op-havoc-rep-128). The two top crash lines (full in the gist):

1. 
```
/home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/helpers.rb:78:in `encode': asciidoctor: FAILED: /home/ariel/afl-kisaten/private/sandbox/asciidoctor/encoding_crash/id:000000,sig:10,src:000048,op:havoc,rep:128: Failed to load AsciiDoc document - "\xDDc" on UTF-16BE (Encoding::InvalidByteSequenceError)
```

2. 
```
/home/ariel/afl-kisaten/private/sandbox/asciidoctor/asciidoctor/lib/asciidoctor/helpers.rb:75:in `encode': asciidoctor: FAILED: /home/ariel/afl-kisaten/private/sandbox/asciidoctor/encoding_crash/id:000001,sig:10,src:000076,op:havoc,rep:128: Failed to load AsciiDoc document - incomplete "t" on UTF-16LE (Encoding::InvalidByteSequenceError)
```

mojavelinux:
resolves #2884

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -1296,7 +1296,7 @@ module Asciidoctor
       source = input
     elsif ::Array === input
       source = input.drop 0
-    else
+    elsif input
       raise ::ArgumentError, %(unsupported input type: #{input.class})
     end
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -85,6 +85,7 @@ Improvements::
   * don't store fallback reference text on :bibref node (#3085)
   * call AbstractNode#reftext instead of AbstractNode#text to retrieve reference text for bibref node (#3085)
   * only map unparsed attrlist of inline macro to target when format is short
+  * add clearer exception message when source data is binary or has invalid encoding (#2884)
 
 Bug Fixes::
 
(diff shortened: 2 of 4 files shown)


lib/asciidoctor.rb
```diff
@@ -1296,7 +1296,7 @@ def load input, options = {}
       source = input
     elsif ::Array === input
       source = input.drop 0
-    else
+    elsif input
```
mojavelinux: This allows input to be nil. The processor will treat it as an empty document.

Issue #2524, opened 2018-01-02
AbstractBlock#find_by should be able to find nodes in tables


mojavelinux:
The Abstract#find_by method currently skips over the contents of tables. This method should support finding nodes in tables, particularly AsciiDoc table cells (which each contain a nested AsciiDoc document). To avoid complicating this method, it may be necessary to introduce a separate method or a dedicated flag.

This method is useful for extensions which may need to process blocks in AsciiDoc table cells.

mojavelinux:
This issue raises a lot of questions, such as should we allow rows, cells, and the nested document of an AsciiDoc table cell to be returned? I think we should spin off a separate issue that just deals with looking for blocks inside an AsciiDoc table cell, then deal with the rest later. See #3101.

mojavelinux:
One problem, in particular, is that Cell is not an AbstractBlock, so it doesn't have the find_by method. Arguably, Cell should be an AbstractBlock, as raise in #2963.

mojavelinux:
Let's focus this issue on finding cells. If we decide find_by should find columns (which are really more of an internal construct), we can open a separate issue.

mojavelinux:
resolved by #3097.

PR #3103, opened 2019-02-25
resolves #2963 modify Cell class to extend from AbstractBlock

### LINKED ISSUE #2963: Table::Cell and Table::Column don't implement block?

mojavelinux:
resolves #2963

- modify Cell class to extend AbstractBlock
- don't split paragraphs in table cell at line that resolves to blank and adjacent to other non-blank lines
- set level property on Cell and Column
- set content_model property on Cell
- define block? and inline? methods on Column that both return false


mojavelinux:
@robertpanzer Before I merge, we need to make a decision about whether to define the `block?` (and possibly `inline?`) methods on Column or not. Otherwise, I think this fits the requirements for #2963 and #2524.

PR #3125, opened 2019-03-02
fix(asciidoc-vs-markdown): use `ul` consistent with user manual

myii:
* https://asciidoctor.org/docs/user-manual/#nested
  - No leading space before the asterisks


mojavelinux:
Technically, the leading spaces are optional. But for the purpose of comparison with Markdown, I think you're right that we should not include them.

myii:
@mojavelinux Yes, I understood that they were optional.  My angle was consistency only.  Thanks for the merge.

PR #3130, opened 2019-03-06
use terminology for find_by that better aligns with NodeFilter

mojavelinux:
- assume that true implies "accept"
- assume that false implies "skip"
- prefer :reject rather than :skip
- prefer :terminate instead of :skip_children

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -70,6 +70,7 @@ Enhancements / Compliance::
   * AbstractBlock#find_by finds table cells, which can be selected using the :table_cell context in the selector (#2524)
   * allow ampersand to be used in e-mail address (#2553)
   * propogate ID assigned to inline passthrough (#2912)
+  * rename control keywords in find_by to better align with the standard NodeFilter terminology
 
 Improvements::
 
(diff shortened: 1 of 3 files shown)


mojavelinux:
@Mogztter What do you think about this change? Specifically, how do you feel about the name "terminate" to "reject children"?

I managed to add the new keywords without breaking backwards compatibility. Technically, `:skip` is the wrong directive for ignoring children (according to the NodeFilter from DOM), but since `false` implies `:skip` already, we can simply add `:reject` and allow `:skip` to be used as an alias for `:reject`.

Mogztter:
I like it :+1:

mojavelinux:
Actually, terminate would be misleading because it gives the impression that the search stops entirely. What we're trying to express is that the recursive descent does not continue past the current node. With that in mind, let's choose from these words:

* truncate
* crop
* clip
* cut
* pare
* prune

I kind of like prune the best. It gives you the picture of cutting off branches from a trunk, which is precisely what we're doing.

Mogztter:
> Actually, terminate would be misleading because it gives the impression that the search stops entirely. What we're trying to express is that the recursive descent does not continue past the current node. With that in mind, let's choose from these words:

It's true but you can also read it as "terminate" the current "branch" (and not terminate the whole search).

> I kind of like prune the best. It gives you the picture of cutting off branches from a trunk, which is precisely what we're doing.

It might be because I'm using git a lot but I also like prune :+1:

mojavelinux:
Let's go with prune.

PR #3134, opened 2019-03-08
resolves #3131 truncate with precision (instead of rounding) when computing abswidth for columns in DocBook output

### LINKED ISSUE #3131: Use precision when computing absolute column width (colabswidth)
For some reason, the absolute column width is computed using a different rounding strategy than the percentage column width. The same strategy should be used in both cases, namely to truncate to the precision factor.

mojavelinux:
resolves #3131

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -98,6 +98,7 @@ Improvements::
   * add clearer exception message when source data is binary or has invalid encoding (#2884)
   * rename context for table cell and table column to :table_cell and :table_column, respectively
   * rename hardbreaks document attribute to hardbreaks-option; retain hardbreaks as a deprecated alias (#3123)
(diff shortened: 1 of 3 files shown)


Mogztter:
I think `truncate` is implemented as part of Number (and does not rely on Ruby Math) so we should be able to remove https://github.com/asciidoctor/asciidoctor.js/blob/master/lib/asciidoctor/js/opal_ext/number.rb

mojavelinux:
@Mogztter That's correct. Where this implementation is still missing is Ruby 2.3. So we have a polyfill in core for it.

mojavelinux:
What's funny about this pull request is that it's really something I'd rather just remove entirely. It's very specific to DocBook / PDF output and probably rarely used, if ever.

PR #3143, opened 2019-03-11
revise the RDoc on the Asciidoctor module [skip ci]

mojavelinux:


Mogztter:
I will update the documentation in Asciidoctor.js: https://github.com/asciidoctor/asciidoctor.js/blob/35232011574afb13fbdff10d2532ad38ab644425/src/asciidoctor-core-api.js#L120

PR #3144, opened 2019-03-12
Always store section numeral as string

vmj:
The first commit tries to make sure `numeral` is always a string, as it states in the RDoc.

The second commit changes the `numeral` of book parts from arabic to roman numerals.  The code used to make the translation only in `sectnum` method, leaving the actual attribute to arabic numeral.  Now the translation is done earlier.

These changes are part of the asciidoctor/asciidoctorj#549

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -76,6 +76,7 @@ Enhancements / Compliance::
 Improvements::
 
   * propagate document ID to DocBook output (#3011)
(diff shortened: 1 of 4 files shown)


mojavelinux:
Let's just drop `number=` since it really just doesn't make sense.

vmj:
Awesome!  Thanks, @mojavelinux, for cleaning this up :)

mojavelinux:
Thanks again!

PR #3146, opened 2019-03-13
resolves #1444 rename header_footer option to standalone

### LINKED ISSUE #1444: Rename header_footer option to clarify its intent
The `header_footer` option controls whether the converter outputs a standalone document (`document` converter method) or an embeddable document (`embedded` converter method). Not only is this distinction not clear from the name `header_footer`, the terms "header" and "footer" are overloaded, so it's really not clear what this option does for someone who doesn't read lots of mailinglist posts.

We should consider changing the name of this option to `embeddable` or similar so the real intent is clear. Of course, can can continue to support the old option for backwards compatibility reasons, though it will be considered deprecated (and may issue a deprecation warning). If both are specified, the new option will win out.

mojavelinux:
* still honor header_footer option as fallback for backwards compatibility

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -71,6 +71,7 @@ Enhancements / Compliance::
   * allow ampersand to be used in e-mail address (#2553)
   * propogate ID assigned to inline passthrough (#2912)
   * rename control keywords in find_by to better align with the standard NodeFilter terminology
+  * rename header_footer option to standalone (while still honoring header_footer for backwards compatibility) (#1444)
 
 Improvements::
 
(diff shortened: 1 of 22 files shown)


Mogztter:
We might need to reconsider the command line option `-s`.
Now that we have an option named "standalone" it can be confusing to add `-s` to convert to an embeddable document.

`-e` is already taken but we could only allow the long format to specify the eRuby implementation as, in my opinion, it's not used often.
If we do that we could use `-e, --embeddable` to convert to an embeddable document.

"standalone" is fine but since it's the "default" mode, I'm in favor of reversing the boolean logic and named it `embeddable` to be consistent with the CLI:

```
$ asciidoctor README.adoc --embeddable
```

```rb
Asciidoctor.convert_file 'README.adoc', embeddable: true
```

mojavelinux:
I looked into this while I was implementing the change and I determined it would be prohibitively complicated to reverse the logic and not severely break backwards compatibility. The reason is the code relies on the option not being set in several circumstances, and there would be no way for us to know whether not setting it meant to enable or disable it if we reversed the logic. Therefore, we need to keep the boolean logic the same.

Choosing the option name `standalone` is further justified by the fact that pandoc has an option with the same name, hence establishing a precedent.

I'm not super concerned about the overloading of "s" between the option name and the CLI since these are different contexts and thus not a direct mapping. I have always believed that `-s` is a strange choice for disabling the header and footer, but it's critical to maintain for backwards compatibility. (Despite the fact that `-s` is the option name in pandoc). We could add a more friendly option as an alias, though.

I have decided that even though it's the wrong grammar, embedded is the more widely adopted term than embeddable. Rather than try to swim upstream, let's just go with what's natural.

Mogztter:
> We could add a more friendly option as an alias, though

👍 

> I have decided that even though it's the wrong grammar, embedded is the more widely adopted term than embeddable. Rather than try to swim upstream, let's just go with what's natural.

I agree, embedded sounds familiar, embeddable not so much.

PR #3150, opened 2019-03-16
when resolving a convert method for a node, only consider methods prefixed with convert_

mojavelinux:
- prefix convert methods on built-in converters with convert_
- drop leading underscore from helper methods on built-in converters
- add adapter for older converters that rely on unprefixed convert methods
- use optimize convert handler dispatcher in HTML 5 converter
- only pass options to convert handler method if value is non-nil
- update RDoc

### DIFF
--- a/lib/asciidoctor/converter.rb
+++ b/lib/asciidoctor/converter.rb
@@ -36,7 +36,7 @@ module Asciidoctor
 #
 #   class Html5Converter < (Asciidoctor::Converter.for 'html5')
 #     register_for 'html5'
-#     def paragraph node
+#     def convert_paragraph node
 #       %(<p>#{node.content}</p>)
 #     end
 #   end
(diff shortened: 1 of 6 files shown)


mojavelinux:
@Mogztter I feel like this is just the right thing to do. It has negligible to no impact on performance. And it helps keep convert methods from colliding with helper methods and built-in Ruby methods. wdyt?

I managed to figure out how to avoid breaking existing converters that use the built-in HTML 5 converter by leaning on method_missing.

Mogztter:
I didn't mind the underscore prefix for helper methods but arguably it's less hacky to just prefix convert methods with `convert_`.

> It has negligible to no impact on performance. And it helps keep convert methods from colliding with helper methods and built-in Ruby methods.

I'm fine with that change 👍 

> I managed to figure out how to avoid breaking existing converters that use the built-in HTML 5 converter by leaning on method_missing.

Cool, nicely done 😉

PR #3157, opened 2019-03-18
resolves #1699 create details/summary tag set from example block if collapsible option is set

### LINKED ISSUE #1699: New block: details
The new [`<details>` and `<summary>`](https://developer.mozilla.org/it/docs/Web/HTML/Element/details) elements are getting support on [firefox too](https://bugzilla.mozilla.org/show_bug.cgi?id=591737) as of latest nightly (47) – they already are on chrome, safari and opera.

Currently we can use an open block and a custom template, but I think it would be great implementing out-of-the-box support for them, as they can be very useful (they even got their official html element, so...)

Maybe it could be merged with the [Q&A](http://asciidoctor.org/docs/user-manual/#question-and-answer-style-list)?

mojavelinux:


lib/asciidoctor/converter/html5.rb
```diff
@@ -571,13 +571,23 @@ def convert_dlist node
 
   def convert_example node
     id_attribute = node.id ? %( id="#{node.id}") : ''
-    title_element = node.title? ? %(<div class="title">#{node.captioned_title}</div>\n) : ''
-
-    %(<div#{id_attribute} class="exampleblock#{(role = node.role) ? " #{role}" : ''}">
+    if node.option? 'collapsible'
+      class_attribute = node.role ? %( class="#{node.role}") : ''
+      summary_element = node.title? ? %(<summary>#{node.title}</summary>) : '<summary>Details</summary>'
```
Mogztter: I think we should i18n "Details" right ? (controlled by an attribute)

mojavelinux: In truth, the summary should never not be set. The only reason we place it here is so that the styles will work if you forget to set it. (If we figure out how to do this purely with CSS, even better).

mojavelinux: This is one of those really weird things where the browser adds an element, but it's not really in the DOM. That makes it extremely difficult to style. So while I can think of some hacks to try to inject CSS in its absence, this is really a fallback scenario and we should just encourage the use of a title.

mojavelinux: The alternative is to use a blank value, which would also work.

Mogztter: If we use a default value then we should allow the user to override it.
Arguably a user could do the following:

```adoc
:details-title: Détails

.{details-title}
[%collapsible]
====
This content is revealed when the user clicks the words "Toggle Me".
====
```

> The alternative is to use a blank value, which would also work.

That could work.
In my opinion, the only advantage of using an overridable default value is that you can use a more concise syntax (ie. an example block without title):

```adoc
[%collapsible]
====
This content is revealed when the user clicks the words "Toggle Me".
====
```

mojavelinux: Since there are still decisions to make about the i18n title (such as the name of the attribute), and I don't want it to hold up the 2.0 release, let's pursue it in a follow-up issue. We can move forward with the hard-coded title for now.

PR #3160, opened 2019-03-18
resolves #450 generate index in HTML 5 converter

### LINKED ISSUE #450: Output index when using HTML5 backend (like a2x)
Emulate the behavior of a2x to output an index of index terms in the same way as a2x -f xhtml (via the docbook toolchain). Output in a section with the style `index`, if that section is present.

Example terms:

```
Lions and tigers (Bengal and Siberian) are big cats.
(((Big cats,Lions)))
(((Big cats,Tigers,Bengal Tiger)))
(((Big cats,Tigers,Siberian Tiger)))
Multi-entry terms generate separate index entries.
```

Example index section:

```
[index]
== Index terms
```

Organize the index entries as definition lists and nested index entries as nested definition lists.

mojavelinux:
- add new model for storing index terms
- populate index in HTML 5 output if index special section is present
- update tests to use new model


mojavelinux:
This is missing the stylesheet changes.

mojavelinux:
@Mogztter I think this needs more time to bake. However, the problem is, it introduces a break in the catalog API, which would prevent us from adding it as part of 2.1. However, since we know the API is going to change, we can break it in 2.0 by simply not registering the index terms in the catalog anymore. Then, when we add it back in with the new model, it's a new API and not a breaking one.

Mogztter:
> I think this needs more time to bake. However, the problem is, it introduces a break in the catalog API, which would prevent us from adding it as part of 2.1

This feature feels like 2.1.

>  However, since we know the API is going to change, we can break it in 2.0 by simply not registering the index terms in the catalog anymore.

Yes I think that's the right approach.

PR #3168, opened 2019-03-19
resolves #3161 log warning if conditional expression in ifeval directive is invalid

### LINKED ISSUE #3161: Warn if ifeval doesn't match
Right [now](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/reader.rb#L955) if `ifeval` contains an invalid expression it is considered `false`. I think it'd be a little more friendly to emit a warning here.

If this is something that you'd like in Asciidoctor I'd be happy to open up a pull request!

mojavelinux:
resolves #3161

- log warning if conditional expression in ifeval directive is invalid (#3161)
- drop lines that contain an invalid preprocessor directive (#3161)


nik9000:
Awesome!

PR #3171, opened 2019-03-19
resolves #2868 log warning when include directive is not resolved due to missing attribute

### LINKED ISSUE #2868: Warn about dropped include directive when attribute is not resolved
When an attribute in the target of an include cannot be resolved, the processor warns about the line being dropped, but gives no indication that it's an include directive. The warning should mention the include directive.

This is related to #2855.

mojavelinux:


Mogztter:
:+1:

PR #3177, opened 2019-03-20
resolves #3176 log message at info level if inline macro processor returns a String value

### LINKED ISSUE #3176: Warn if inline macro processor returns a String value
We want to encourage extensions to remain within the AST model abstraction. Therefore, if an inline macro returns a String, the processor should issue a warning.

If the extension wants to return raw text, it should be wrapped in an Inline node of type `:quoted` with no type:

```rb
create_inline parent, :quoted, 'raw text'
```

Which will implicitly set the type to `:unquoted`.

This is not yet a perfect abstraction, but it gets us one step closer to a pure model. Ideally, the extension would return content which can be run through any converter. This may mean applying subs when creating the Inline node. (As a follow-up step, we may want to check if the Inline defines a subs attribute and apply those to the text).

mojavelinux:


mojavelinux:
@Mogztter I wonder if we should add a `create_pass` (or `create_inline_pass`) helper. wdyt? Then it would be as simple as:

```
create_pass parent, 'text'
```

mojavelinux:
(Of course, that makes me wonder if the helper methods should follow the template names, so it would be `create_pass` for block and `create_inline_pass` for an inline node).

mojavelinux:
See #3178 for addition of create_inline_pass helper method.

PR #3180, opened 2019-03-21
use .merge to duplicate Hash instead of .dup

mojavelinux:
- use .merge to duplicate Hash instead of .dup because it's much faster
- include patch for .merge to accept variable length argument list


mojavelinux:
@Mogztter this has a stunning impact on performance for CRuby.

PR #3197, opened 2019-03-23
resolves #3196 link --trace option to $TRACE global; log backtrace if $TRACE is set

### LINKED ISSUE #3196: Include backtrace in log messages where appropriate if --trace option is set
If the --trace option is set by the CLI, or the $TRACE global variable is true, include the backtrace in any log messages where the backtrace would be relevant (such as a load error). This will help with debugging.

The key to solving this issue is hooking the --trace CLI option to the $TRACE global variable. When using the API, this variable should be set directly.

mojavelinux:


mojavelinux:
After further review, we have decided not to proceed with this change. See the issue for details. I have, however, brought over the test into the main branch.

PR #3198, opened 2019-03-23
resolves #2676 export static functions on Extensions as module functions

### LINKED ISSUE #2676: Make Asciidoctor::Extensions.register a module_function
Because `register` is defined inside `class << self`, it is currently not possible to write an extension like this:

````ruby
require 'asciidoctor/extensions'

include Asciidoctor::Extensions

register do
  inline_macro :git do
    process do |parent, target, attrs|
      command = "git-#{ target }"
      create_anchor parent, command, :type => :link, :target => "https://git-scm.com/docs/#{ command }"
    end
  end
end
````

Much of the stdlib uses `module_function`, so we get used to it working like that. I think this seems more intuitive for anyone. Can we consider turning `register` and friends into a `module_function`?

mojavelinux:


mojavelinux:
Experience has taught me that module functions are a dreadful idea because they put functions into the global space where they can conflict with other libraries. There's no reason a library should be doing this (but a user is still free to introduce a shim that provides this functionality). I won't be moving forward with this idea.

PR #3204, opened 2019-03-25
test: adjust message when asciimath is not available

### LINKED ISSUE #1884: Reasons for LoadErrors when loading Ruby libs are swallowed.
As observed in robertpanzer/asciidoctorj-pdf#3 requiring the rouge library fails on Windows when loading the gem from an exploded directory.
As https://github.com/asciidoctor/asciidoctor/blob/93063e2500369612e923c04695fbf6e3054b31ba/lib/asciidoctor/helpers.rb#L30 completely swallows the LoadError it seems like rouge was not available.
In fact the error comes deeper from within the rouge code when it loads polyglot.

It would be nice if the original error backtrace and message would be passed on.

tmzullinger:
The asciidoctor/helpers text was changed in a667772a ("resolves #1884
add detail to load error message if path is not gem name (PR #3195)",
2019-03-25).

This breaks test/blocks_test.rb and test/substitutions_test.rb,
resulting in a failure like:

--- expected
+++ actual
@@ -1 +1 @@
-"optional gem 'asciimath' is not installed. Functionality disabled."
+"optional gem 'asciimath' is not available. Functionality disabled."

Ensure the text in the asciimath tests matches what we output from
asciidoctor/helpers.

### DIFF
--- a/test/blocks_test.rb
+++ b/test/blocks_test.rb
@@ -1859,7 +1859,7 @@ context 'Blocks' do
           assert_equal expect, actual.strip
           assert_equal :loaded, doc.converter.instance_variable_get(:@asciimath_status)
         else
-          assert_message logger, :WARN, 'optional gem \'asciimath\' is not installed. Functionality disabled.'
+          assert_message logger, :WARN, 'optional gem \'asciimath\' is not available. Functionality disabled.'
           assert_equal :unavailable, doc.converter.instance_variable_get(:@asciimath_status)
         end
       end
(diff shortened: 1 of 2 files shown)


tmzullinger:
I ran into this while working on updating the Fedora packages from 2.0.0 to 2.0.1, where we don't have asciimath.

tmzullinger:
The inclusion of the diff causes grief when trying to apply this patch manually.  I thought indenting it would resolve that, but it doesn't.  I can drop that from the commit message if we want to avoid causing anyone trouble if they try to extract this with `git format-patch` and apply it manually.

mojavelinux:
Thanks @tmzullinger!

PR #3219, opened 2019-03-29
switch CI build from trusty to xenial

mojavelinux:


mojavelinux:
JRuby is failing when using subprocesses (both in the invoker tests and to interface with Pygments). I have no idea why this is happening.

mojavelinux:
I verified the tests run on xenial using a local Docker container. I'm not going to fight with Travis over this one right now.

PR #3236, opened 2019-04-03
resolves #3231 assume implicit AsciiDoc extension on interdoc xref macro target with no extension

### LINKED ISSUE #3231: xref:name#[] is not interpreted as source-to-source reference since 2.x
Asciidoctor 1.5.x converts `xref:name#[]` to href `name.html` (when out `outfilesuffix` is `.html`). Now it’s converted to `name`. This is a very serious **breaking change** for me and I don’t even see any reason for it. I have thousands of documents that will be converted incorrectly in Asciidoctor 2.0+.

Related to #3021

mojavelinux:


mojavelinux:
I believe this change is benign. The only thing it blocks is being able to use the xref macro for linking to an extensionless file. However, we don't recommend the use of extensionless files anyway. And there's still room in the future for a dedicated attribute to provide a hint to *not* assume the extensionless target is an AsciiDoc file.

PR #3240, opened 2019-04-04
replace calls to File.extname with Helpers.extname

mojavelinux:


lib/asciidoctor/helpers.rb
```diff
@@ -207,6 +207,24 @@ def self.extname? path
     (last_dot_idx = path.rindex '.') && !(path.index '/', last_dot_idx)
   end
 
+  # Public: Retrives the file extension of the specified path. The file extension is the portion of the path in the last
```
Mogztter: Retrieves ?

mojavelinux: Indeed. Nice catch.

PR #3244, opened 2019-04-06
move source for main API entry points (load, load_file, convert, convert_file) to separate files

mojavelinux:


mojavelinux:
@Mogztter This shouldn't affect any existing behavior or integration with the Asciidoctor.js build. However, I think it's a step towards making the source code easier to page through. I'm also thinking of moving the regular expressions to a separate file as well.

PR #3259, opened 2019-04-13
CI: Use jruby-9.2.7.0, 2.4.6

olleolleolle:
- Also: Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


mojavelinux:
Thanks!

PR #3264, opened 2019-04-15
move regular expression definitions to separate source file

mojavelinux:


lib/asciidoctor/rx.rb
```diff
@@ -0,0 +1,728 @@
...
+  #   file:///path
+  #   data:info
+  #
+  #   not c:/sample.adoc or c:\sample.adoc
+  #
+  UriSniffRx = %r(^#{CG_ALPHA}[#{CC_ALNUM}.+-]+:/{0,2})
+
+  # Detects XML tags
+  XmlSanitizeRx = /<[^>]+>/
+#end
```
Mogztter: ?

mojavelinux: It's the end of the pseudo module Rx, though at this point I think it's clear we decided not to use a submodule and thus we can remove the whole pseudo block.

PR #3277, opened 2019-04-25
fixed markup

0xflotus:


mojavelinux:
Thanks for catching and fixing this!

PR #3291, opened 2019-05-02
resolves #3289 fix style of title on excerpt (quoteblock variation) (PR #3291)

### LINKED ISSUE #3289: Title on excerpt quote block renders on top of quote bar
When I use a title on a quote block with an excerpt role

```
.Comments
[.excerpt]
____
This is a comment.
____
```

the title is rendered in HTML partially on top of the quote bar

title-quote-excerpt

This is with Asciidoctor 2.0.9 and Google Chrome 73.0.3683.103.

The excerpt role was added in commit 0672524.

mojavelinux:


mojavelinux:
The fix is to apply the left border to the blockquote tag instead of the enclosing container. That way, the title appears above decorated portion.

Issue #571, opened 2013-08-19
Introduce an IncludeResolver extension


mojavelinux:
Building on the extension work in #100, introduce an extension point that is responsible for resolving the include target to qualified path or IO object. This differs from an IncludeProcessor because the resolver merely locates the resource and, optionally, prepares it for reading. The built-in processor then takes over from there.

```
class MyIncludeResolver < IncludeResolver
  def resolve target, attributes
    "includes/#{target}"
  end
end

Extensions.register do |document|
  include_resolver MyIncludeResolver
end
```

Instead of `IncludeResolver`, we may want to name this `ResourceResolver` so that it can be used in other places where files need to be looked up. One example is if we pass in a configuration file at some point, the ResourceResolver could be used to locate it.

mojavelinux:
We'll definitely need support for an IO return value so that the resolver can resolve a file in a virtual file system.

mojavelinux:
The key benefit of this extension is that the extension writer can resolve the include without having to worry about providing the line and tag filter functionality. This differs from the include processor, which takes over all processing.

mojavelinux:
Another possibility is to expose the methods that Asciidoctor uses to handle tag filtering, line filtering, and indentation stripping. That would allow these features to be composed in ways we can't anticipate, such as in extensions.

Mogztter:
> Another possibility is to expose the methods that Asciidoctor uses to handle tag filtering, line filtering, and indentation stripping. That would allow these features to be composed in ways we can't anticipate, such as in extensions.

I think it's a good idea to expose the methods. And if in practice the same pattern keeps repeating, we could then provide an extension point. But first I think we should make it easy to compose these features:

- resolving the target
- reading the content
- filtering by tags
- filtering by line numbers

mojavelinux:
:+1: to composability. But, like you said, it should not be a reason not to solve the real issue at hand.

cjbrooks12:
I've run into [a problem](https://github.com/asciidoctor/asciidoctorj/issues/903) where I need this issue resolved. I see there's a PR that's been open for a while addressing this issue, is there anything I can do to help get it merged soon?

mblythe86:
> Another possibility is to expose the methods that Asciidoctor uses to handle tag filtering, line filtering, and indentation stripping. That would allow these features to be composed in ways we can't anticipate, such as in extensions.

[...]

PR #3296, opened 2019-05-05
Extract the unresolved directive and return block

Mogztter:
My goal is to create a proper API to filter lines by tags and/or line numbers.
I've introduced minimal changes to extract the return block from the lines filtering processing.

For now the method takes the following arguments:

- a `File` (we could pass a list of `lines` using `readlines` but maybe it's less performant ?)
- a list of tags or a list a line numbers
- expanded target (used to create an "include cursor" shown in warnings log during the processing)
- parsed attributes (used to define `partial-option`)
- include path (used in warning logs during the processing)
- relative path to the outermost document (used in the `push_include` method)
- target type (used in warning logs during the processing)

From a user perspective, I think we should be able to filter lines given a list of tags or line numbers:

```rb
lines = filter_lines(lines, {tags: {"**"=>true, "bark"=>false}})
```

```rb
lines = filter_lines(lines, {linenos: [2, 3]})
```

And the global flow:

```rb
class MyIncludeProcessor < Asciidoctor::Extensions::IncludeProcessor
  def process document, reader, target, attrs
    content = reader.read target
    linenos, tags = reader.parse_include_filtering_attributes attrs
    content, inc_offset = reader.filter_include_lines(content, {linenos: linenos, tags: tags})
    if inc_offset
      reader.push_include content, target, target, inc_offset, attrs
    end
  end
end
```

If we don't want to rework this part of the code, we could introduce a new extension point that could be notified just before the `push_include`.
That would allow a developer to work with the content to be included (ie. already filtered).

Ref #571

### DIFF
--- a/lib/asciidoctor/reader.rb
+++ b/lib/asciidoctor/reader.rb
@@ -1090,116 +1090,114 @@ class PreprocessorReader < Reader
         end
       end
 
-      if inc_linenos
-        inc_lines, inc_offset, inc_lineno = [], nil, 0
-        begin
+      begin
+        if inc_linenos
           reader.call inc_path, read_mode do |f|
-            select_remaining = nil
-            f.each_line do |l|
-              inc_lineno += 1
-              if select_remaining || (::Float === (select = inc_linenos[0]) && (select_remaining = select.infinite?))
-                # NOTE record line where we started selecting
-                inc_offset ||= inc_lineno
-                inc_lines << l
-              else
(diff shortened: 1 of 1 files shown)


Mogztter:
While working on https://gitlab.com/gitlab-org/gitlab-ce/merge_requests/28417 I stumbled across https://github.com/jirutka/asciidoctor-include-ext

Maybe we could take some ideas from it.

PR #3305, opened 2019-05-18
upgrade truffleruby to 19.2.0

Mogztter:
* [x] https://github.com/oracle/truffleruby/issues/1692

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
     env: JRUBY_OPTS='--dev'
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
-  - rvm: truffleruby-19.1.0
+  - rvm: truffleruby-19.2.0
     if: env(TRAVIS_EVENT_TYPE) = cron
     env:
     # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
   - rvm: truffleruby-19.2.0
-    if: env(TRAVIS_EVENT_TYPE) = cron
+    #if: type = cron
     env:
     # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
     - TRUFFLERUBYOPT="--vm.Xmx3g"
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ matrix:
     env: JRUBY_OPTS='--dev'
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
+  - rvm: truffleruby-19.1.0
+    if: env(TRAVIS_EVENT_TYPE) = cron
+    env:
+    # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
+    - TRUFFLERUBYOPT="--vm.Xmx3g"
 env:
   global:
   # use system libraries to speed up installation of nokogiri
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
   - rvm: truffleruby-19.2.0
+    # run against TruffleRuby on a daily basis
     #if: type = cron
     env:
     # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
   - rvm: truffleruby-19.2.0
-    # run against TruffleRuby on a daily basis
+    # run against TruffleRuby on a daily basis.
     #if: type = cron
     env:
     # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ rvm:
 - 2.5.5
 - 2.4.6
 - &oldest_ruby 2.3.8
-# the test suite currently crashes on truffleruby
-#- truffleruby-rc13
 matrix:
   include:
   - rvm: jruby-9.2.7.0
@@ -19,8 +17,8 @@ matrix:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
   - rvm: truffleruby-19.2.0
-    # run against TruffleRuby on a daily basis.
-    #if: type = cron
+    # run against TruffleRuby nightly
+    if: type = cron
     env:
     # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
     - TRUFFLERUBYOPT="--vm.Xmx3g"
--- a/Gemfile
+++ b/Gemfile
@@ -25,3 +25,7 @@ end
 group :ci do
   gem 'simplecov', '~> 0.16.0'
 end
+
+if RUBY_ENGINE == 'truffleruby'
+  gem 'haml', '~> 4.0'
+end
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -71,7 +71,7 @@ context 'API' do
       ensure
         tmp_input.close!
       end
-    end unless RUBY_ENGINE == 'truffleruby' # https://github.com/oracle/truffleruby/issues/1714
+    end
 
     test 'should not load invalid file' do
       sample_input_path = fixture_path('hello-asciidoctor.pdf')
--- a/test/api_test.rb
+++ b/test/api_test.rb
@@ -71,7 +71,7 @@ context 'API' do
       ensure
         tmp_input.close!
       end
-    end
+    end unless RUBY_ENGINE == 'truffleruby' # https://github.com/oracle/truffleruby/issues/1714
 
     test 'should not load invalid file' do
       sample_input_path = fixture_path('hello-asciidoctor.pdf')
--- a/test/invoker_test.rb
+++ b/test/invoker_test.rb
@@ -171,7 +171,7 @@ context 'Invoker' do
         warnings = err.string
       end
       assert_equal false, $VERBOSE
-      refute_empty warnings
+      refute_empty warnings unless RUBY_ENGINE == 'truffleruby'
     rescue
       $VERBOSE = old_verbose
     end
--- a/test/invoker_test.rb
+++ b/test/invoker_test.rb
@@ -167,11 +167,14 @@ context 'Invoker' do
     begin
       warnings = nil
       redirect_streams do |out, err|
-        invoke_cli_to_buffer(%w(-w -o /dev/null), '-') { $NO_SUCH_VARIABLE || 'text' }
+        invoke_cli_to_buffer(%w(-w -o /dev/null), '-') {
+          A_CONST = 10  
+          A_CONST = 20
+        }
         warnings = err.string
       end
       assert_equal false, $VERBOSE
-      refute_empty warnings unless RUBY_ENGINE == 'truffleruby'
+      refute_empty warnings
     rescue
       $VERBOSE = old_verbose
     end


Mogztter:
`bundle install` does not throw a segmentation fault anymore but took 285 seconds.
For comparison, JRuby took approximatively 25 seconds and MRI 2.6 took 15 seconds.

`bundle exec rake` throws an `OutOfMemoryError` exception on Travis:

```console
Fatal error: java.lang.OutOfMemoryError: Could not allocate an aligned heap chunk
```

The full stack trace can be found on Travis: https://travis-ci.org/asciidoctor/asciidoctor/jobs/534193527

For reference, on my laptop with 16gb of memory, the above command does not throw an `OutOfMemoryError` exception.

mojavelinux:
I'm fine with updating the version string, but it's very unlikely we'll enable this in CI anytime soon because it's so unbelievably slow. If you can figure out how to only have it run when the build is triggered, such as nightly, then I'd consider that.

Mogztter:
No worries, I've just enabled it to see how it goes on Travis.

mojavelinux:
I'm glad that you gave it another shot. I am very curious to know when/if it will run to completion, and I'd like to continuously validate that it does...if we can find a good way.

test/api_test.rb
```diff
@@ -71,7 +71,7 @@
       ensure
         tmp_input.close!
       end
-    end
+    end unless RUBY_ENGINE == 'truffleruby' # can't modify frozen String in truffleruby-19.0.0/lib/mri/delegate.rb:349:in `encode!'
```
mojavelinux: We don't force encoding in Asciidoctor anymore, so I'm curious where this call is coming from.

Mogztter:
Following @eregon suggestion in https://github.com/oracle/truffleruby/issues/1696#issuecomment-496158478 I've added `TRUFFLERUBYOPT="--vm.Xmx3g"` and now the build is running fine.

test/api_test.rb
eregon: Maybe it's a bug of `Tempfile.new` with the `:encoding` keyword, could you report it to https://github.com/oracle/truffleruby/issues ?

mojavelinux:
It looks like there are still test failures.

Would you be able to configure the build so it only run on truffleruby during a nightly build?

Mogztter:
> It looks like there are still test failures.

Yes, once https://github.com/oracle/truffleruby/pull/1694 is merged, it will be :green_heart: 

> Would you be able to configure the build so it only run on truffleruby during a nightly build?

Sure.
It think the easiest way is to configure a [cron job on Travis](https://docs.travis-ci.com/user/cron-jobs/), I'm trying to include TruffleRuby when `TRAVIS_EVENT_TYPE` equals `cron`.

Mogztter:
It may be working... @mojavelinux could you please try to trigger a cron job ?

mojavelinux:
I'll have to merge it in order to test (since the branch is not in the upstream). But that's fine, because you've given me the starting point. I'll circle back if it fails.

mojavelinux:
We want to start running a nightly build anyway, so this is a perfect fit for that.

@Mogztter I figured out the right syntax to get it to work during the cron job. The syntax is:

```
if: type = cron
```

However, it seems like the tests still fail. See https://travis-ci.org/asciidoctor/asciidoctor/jobs/539079822

Mogztter:
> However, it seems like the tests still fail. 

Yes, hopefully my fix for https://github.com/oracle/truffleruby/issues/1692 will be available in the next TruffleRuby release.

test/api_test.rb
mojavelinux: @Mogztter?

Mogztter:
https://github.com/oracle/truffleruby/releases/tag/vm-19.1.0 has been released 🎉 
I will update to Truffle Ruby 19.1.0 and check the encoding/Tempfile.new bug.

mojavelinux:
Sweet! Thanks!

test/api_test.rb
Mogztter: The full stack trace:

```console
Minitest::Test::TestAPI::TestLoad#test_should_not_load_file_with_unrecognized_encoding:
FrozenError: can't modify frozen String
    /home/guillaume/.rvm/rubies/truffleruby-19.1.0/lib/mri/delegate.rb:349:in `encode!'
    /home/guillaume/.rvm/rubies/truffleruby-19.1.0/lib/mri/delegate.rb:349:in `write'
    /home/guillaume/.rvm/rubies/truffleruby-19.1.0/lib/mri/delegate.rb:349:in `write'
    /home/guillaume/workspace/opensource/asciidoctor/asciidoctor/test/api_test.rb:64:in `test_should_not_load_file_with_unrecognized_encoding'
```

Mogztter:
Tests are running fine but I had to disable a few of them.
The main reason is that Haml depends on Ripper but this library is not supported by TruffleRuby (it has a no-op implementation).

So when `Ripper#sexp` or `Ripper#lex` is called, I had to disable the test.

Mogztter:
#### TruffleRuby 19.1.0

```
$ time bundle exec rake test
```
```console
Finished in 112.499464s, 21.4223 runs/s, 76.9693 assertions/s.

2410 runs, 8659 assertions, 0 failures, 0 errors, 0 skips
bundle exec rake test  273,66s user 20,54s system 225% cpu 2:10,50 total
```

#### MRI 2.6.3

```
$ time bundle exec rake test
```
```console
Finished in 2.366769s, 1023.7586 runs/s, 3681.3903 assertions/s.

2423 runs, 8713 assertions, 0 failures, 0 errors, 0 skips
bundle exec rake test  2,87s user 0,16s system 100% cpu 3,027 total
```

mojavelinux:
Is it possible to downgrade Haml to a version that doesn't use Ripper? (such as 4.0.7). We can set it in the Gemfile. Then we don't have to disable tests...since we aren't really testing Haml itself.

Mogztter:
That's a good idea, I didn't know that Ripper was introduced in version 5.0.0: 

> Analyze attribute values by Ripper and render static attributes beforehand. (Takashi Kokubun)

https://github.com/haml/haml/blob/9bd79f26e38b1507cd439e5ee8f5eb1023c98701/CHANGELOG.md


I will give it a try.

test/api_test.rb
eregon: @Mogztter  Could you report on TruffleRuby's bug tracker please?

eregon:
Another solution would be to fix Haml to not try to use Ripper if it's not implemented, since it sounds like an optional optimization. We might want to do that anyway so latest Haml works on TruffleRuby.

mojavelinux:
@erebor I agree, but Haml isn't our concern. I recommend Slim for our users anyway.

eregon:
@mojavelinux Yes, I meant this is something the TruffleRuby team should do, as mentioned in https://github.com/oracle/truffleruby/issues/1712

Mogztter:
The build failure is not related, `rvm` had some issues: https://travis-ci.org/asciidoctor/asciidoctor/jobs/554296249

Mogztter:
@mojavelinux downgraded to Haml 4 and it's working fine.

test/api_test.rb
Mogztter: @eregon Yes, I will try to create a minimal reproduction case.

mojavelinux:
Great progress!

.travis.yml
```diff
@@ -10,18 +10,21 @@ rvm:
 - &oldest_ruby 2.3.8
 - jruby-9.2.7.0
 - jruby-9.1.17.0
-# the test suite currently crashes on truffleruby
-#- truffleruby-rc13
 matrix:
   include:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
+  - name: "TruffleRuby"
```
mojavelinux: Is there a reason you're defining the name property here?

.travis.yml
```diff
@@ -10,18 +10,21 @@ rvm:
...
+    rvm: truffleruby-19.1.0
+    if: env(TRAVIS_EVENT_TYPE) = cron
 env:
   global:
   # use system libraries to speed up installation of nokogiri
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   - PYGMENTS_VERSION='~> 1.2.0'
   - SOURCE_DATE_EPOCH=1521504000
+  # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
+  - TRUFFLERUBYOPT="--vm.Xmx3g"
```
mojavelinux: Perhaps just define this environment variable in the matrix entry?

asciidoctor.gemspec
```diff
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'cucumber', '~> 3.1.0'
   # erubis is needed for testing alternate eRuby impls
   s.add_development_dependency 'erubis', '~> 2.7.0'
-  s.add_development_dependency 'haml', '~> 5.0.0'
+  s.add_development_dependency 'haml', '>= 4.0', '< 5.0'
```
mojavelinux: Can you do this in the Gemfile instead? I still want to test on Haml 5 for other Rubies.

It would look something like this:

```rb
if RUBY_ENGINE == 'truffleruby'
  gem 'haml', '~> 4.0'
end
```

test/invoker_test.rb
```diff
@@ -171,7 +171,7 @@
         warnings = err.string
       end
       assert_equal false, $VERBOSE
-      refute_empty warnings
+      refute_empty warnings unless RUBY_ENGINE == 'truffleruby'
```
mojavelinux: Could we assert what the warnings are when running on TruffleRuby?

Mogztter: `warnings` is empty. I guess that `-w` has no effect when running on TruffleRuby.
You want to add the following assertion:

```rb
if RUBY_ENGINE == 'truffleruby'
  assert_empty warnings # no script warnings when running on TruffleRuby
else
  refute_empty warnings
end
```

asciidoctor.gemspec
Mogztter: done

.travis.yml
Mogztter: I didn't know it was possible, fixing...

.travis.yml
Mogztter: No, I've removed it.

.travis.yml
```diff
@@ -10,12 +10,15 @@ rvm:
...
 - jruby-9.1.17.0
-# the test suite currently crashes on truffleruby
-#- truffleruby-rc13
 matrix:
   include:
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
+  - rvm: truffleruby-19.1.0
+    if: env(TRAVIS_EVENT_TYPE) = cron
+    env:
```
Mogztter: should we use a single line ?

```yml
# use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
env: TRUFFLERUBYOPT="--vm.Xmx3g"
```

we could even use an inline comment but I'm always a bit scared of how it will get interpreted by the YAML parser:

```yml
env: TRUFFLERUBYOPT="--vm.Xmx3g" # use 3g to avoid OutOfMemoryError when running the tests suite against TruffleRuby
```

test/invoker_test.rb
mojavelinux: Yes please. Let's assert the current behavior so we know when it changes.

.travis.yml
mojavelinux: Either way works for me.

test/invoker_test.rb
eregon: `-w` enables warnings on TruffleRuby like on MRI, but maybe the particular warning you're expecting here is not emitted. Although in this case I guess `-w` is passed to AsciiDoctor, not as a Ruby interpreter option?

mojavelinux: Asciidoctor passes the option through.

mojavelinux: We turn on script warnings by setting `$VERBOSE = true`. Perhaps this doesn't work on TruffleRuby.

mojavelinux: Aha! Here's the problem. The following code doesn't generate a warning in TruffleRuby.

```rb
puts $NO_SUCH_VARIABLE
```

However, the following does:

```rb
def fn *args
end

vals = %w(a b c)
fn *vals
```

So we just need a different example.

mojavelinux: I'd like the find the most concise code that can produce a warning. There are lots of other things that could work.

mojavelinux: Wow, TruffleRuby really doesn't complain about much. Most of the warnings I know of in Ruby don't show up in TruffleRuby.

mojavelinux: Got it:

```rb
String /foo/
```

Mogztter: What does this code do ? 🤔

mojavelinux: We are validating that using `-w` turns on Ruby script warnings. So we need to make a warning happen.

Mogztter: Yes but I'm curious since this code is "valid" Ruby, what does it do ?

mojavelinux: It converts the regular expression to a String. But it throws a warning since the `/` is ambiguous.

Mogztter: I see... thanks for the explanation. I will go ahead and replace the code to make sure that a warning is thrown when running on TruffleRuby, MRI or JRuby.

mojavelinux: Thanks!

If you want, you can leave a comment saying that the point of the code is to force a warning. Otherwise, it's pointless logic.

Mogztter: It's not working, when I'm running the tests suite I can see the warning in the console (before the tests suite starts):

```
/home/guillaume/workspace/opensource/asciidoctor/asciidoctor/test/invoker_test.rb:172: warning: Ambiguous first argument; make sure.
```

And the test fails:

```
Minitest::Test::TestInvoker#test_should_enable_script_warnings_if_w_flag_is_specified [/home/guillaume/workspace/opensource/asciidoctor/asciidoctor/test/invoker_test.rb:181]:
Expected "" to not be empty.
```


```rb
  test 'should enable script warnings if -w flag is specified' do
    old_verbose, $VERBOSE = $VERBOSE, false
    begin
      warnings = nil
      redirect_streams do |out, err|
        invoke_cli_to_buffer(%w(-w -o /dev/null), '-') { String /foo/ }
        warnings = err.string
      end
      assert_equal false, $VERBOSE
      refute_empty warnings
    rescue
      $VERBOSE = old_verbose
    end
  end
```

Mogztter: Never mind I get the same error with MRI:

```
/home/travis/build/asciidoctor/asciidoctor/test/invoker_test.rb:170: warning: ambiguous first argument; put parentheses or a space even after `/' operator
 ```

Not sure why but `String /foo/` is not working.

Comment (Mogztter) [reply test/invoker_test.rb]: I think it's because Ruby warns when the code is interpreted not executed.
Ruby knows that `String /foo/` is ambiguous right away but it cannot determine that `$NO_SUCH_VARIABLE` is not initialized until the code is executed.

Comment (Mogztter) [reply test/invoker_test.rb]: I'm now using:

```rb
A_CONST = 10  
A_CONST = 20
```

The above code throws a "runtime" warning on MRI, JRuby and TruffleRuby.

Mogztter:
Upgraded to TruffleRuby 19.2.0: https://github.com/oracle/truffleruby/blob/master/CHANGELOG.md#1920

Gemfile
```diff
@@ -25,3 +25,7 @@ end
 group :ci do
   gem 'simplecov', '~> 0.16.0'
 end
+
+if RUBY_ENGINE == 'truffleruby'
+  gem 'haml', '~> 4.0'
```
eregon: Minor note: it's also possible to use:
```ruby
gem 'haml', '~> 4.0', platforms: [:truffleruby]
```

eregon:
This looks good! :+1:

eregon:
I'd be curious to see the timings once this runs on TravisCI, hopefully Bundler is not that slow anymore on TruffleRuby.

test/invoker_test.rb
```diff
@@ -167,7 +167,10 @@
     begin
       warnings = nil
       redirect_streams do |out, err|
-        invoke_cli_to_buffer(%w(-w -o /dev/null), '-') { $NO_SUCH_VARIABLE || 'text' }
+        invoke_cli_to_buffer(%w(-w -o /dev/null), '-') {
+          A_CONST = 10  
+          A_CONST = 20
```
mojavelinux: I like this way of triggering a warning. I think this should be moved to master now.

.travis.yml
```diff
@@ -18,6 +18,11 @@ matrix:
     env: JRUBY_OPTS='--dev'
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
+  - rvm: truffleruby-19.2.0
+    if: env(TRAVIS_EVENT_TYPE) = cron
```
mojavelinux: Go ahead and comment out this line for the purpose of this PR.

mojavelinux: Also, I believe the line should be:

```
if: type = cron
```

test/invoker_test.rb
eregon: I'd support merging this PR to master rather :)
It's been a long time and lots of efforts, thank you @Mogztter and @mojavelinux!

mojavelinux: I certainly support merging this PR to master once we can verify all the tests are passing. What I'm talking about here is that I like this technique for triggering a warning because we know it is very portable.

.travis.yml
Mogztter: The documentation is really confusing:

> To check whether a build was triggered by cron, examine the TRAVIS_EVENT_TYPE environment variable to see if it has the value cron.

https://docs.travis-ci.com/user/cron-jobs/#detecting-builds-triggered-by-cron

> The following known attributes are available, and can be used in conditions:
type (the current event type, known event types are: push, pull_request, api, cron)

https://docs.travis-ci.com/user/conditions-v1#integration

So according to the documentation both should work but `if: type = cron` is more concise (and easier to read).

Mogztter:
> I'd be curious to see the timings once this runs on TravisCI, hopefully Bundler is not that slow anymore on TruffleRuby.

https://travis-ci.org/asciidoctor/asciidoctor/jobs/576147103

`bundle install` did succeed on the first attempt but still a bit slow 294.85s.

The build failed when running the following command:

```
/home/travis/.rvm/rubies/truffleruby-19.2.0/bin/truffleruby -S bundle exec cucumber -f progress --no-color
truffleruby: /home/travis/.rvm/rubies/truffleruby-19.2.0/bin/bundle:5: `$(' is not allowed as a global variable name (SyntaxError)
```

.travis.yml
mojavelinux: I also tested the TRAVIS_EVENT_TYPE and found it didn't work. But `type = cron` did work.

Mogztter:
I can reproduce this issue on my machine but `bundle exec cucumber -f progress` is running fine.
The line is `SELF_PATH=$(cd "$(dirname "$0")" && pwd -P)/$(basename "$0")` but even MRI can't execute the file `truffleruby-19.2.0/bin/bundle`:

```console
$ ruby -v
ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]
```
```console
$ ruby -S /path/to/truffleruby-19.2.0/bin/bundle
/home/guillaume/.rvm/rubies/truffleruby-19.2.0/bin/bundle:5: `$(' is not allowed as a global variable name
/home/guillaume/.rvm/rubies/truffleruby-19.2.0/bin/bundle:5: syntax error, unexpected end-of-input
SELF_PATH=$(cd "$(dirname "$0")" && pwd ...
```

eregon:
Thanks, Bundler caching in TravisCI (https://docs.travis-ci.com/user/caching/#bundler) should help quite a bit.

Re `bundle:5: $( is not allowed as a global variable name (SyntaxError)`, it's reported as https://github.com/oracle/truffleruby/issues/1697 and basically a bug of RVM breaking TruffleRuby's `bundle` launcher. We still need to figure out how to fix that.
FWIW, what triggers the bug in this case is `lib/cucumber/rake/task.rb` using `path/to/ruby -S bundle ...` instead of just `bundle ...`.

eregon:
I made a PR to RVM to fix `ruby -S bundle`: https://github.com/rvm/rvm/pull/4766
It should be merged soon, and then it's enough to just rerun in TravisCI.

Mogztter:
> I made a PR to RVM to fix `ruby -S bundle`: [rvm/rvm#4766](https://github.com/rvm/rvm/pull/4766)
> It should be merged soon, and then it's enough to just rerun in TravisCI.

Awesome, thanks Benoit 👍

eregon:
Fix for `ruby -S bundle` merged in RVM, please trigger try a rerun in TravisCI.

mojavelinux:
> Thanks, Bundler caching in TravisCI (docs.travis-ci.com/user/caching/#bundler) should help quite a bit.

We don't use Bundler caching because I don't trust it. I want completely clean builds so we know exactly what's going on and what to expect. Tainting the results only taints our understanding.

Mogztter:
@erebor I did but it failed with a SegFault: https://travis-ci.org/asciidoctor/asciidoctor/jobs/576311234 (probably not related)

Mogztter:
Green bars 💚

mojavelinux:
Hooray! Nice work @Mogztter and @eregon!

@Mogztter Go ahead and uncomment the cron condition. Then I'll merge and kick off the scheduled build.

Mogztter:
@mojavelinux Let me know if you want me to squash the commits

mojavelinux:
I'll take care of that via the interface. Let's do this!

Gemfile
mojavelinux: Ah. Nope. That ends up disabling haml for all the other platforms, which isn't what we want.

mojavelinux:
Green bar!

https://travis-ci.org/asciidoctor/asciidoctor/jobs/576533763

You may still be interested in tracking that build to catch warnings and study timings.

mojavelinux:
Thanks again for your support @eregon!

PR #3326, opened 2019-06-07
resolves #3321 add test to assert checkbox cannot be followed by tab

### LINKED ISSUE #3321: Checklist close bracket must be followed by space - Document via lists_test.rb
User story:

Asciidoc is excellent for the documents we write, and we want checklists and following text to line up in the source text and the render. Using a misconfigured editor that embedded tabs lead to this possible issue.

Issue: 

A checklist [x] immediately followed by a \t, optional space, then text, renders the line as text, `[x] my checklist`.
A checklist [x] immediately followed by a space then \t, renders as expected.

- Without an editor revealing whitespace characters, the input text looks normal.
- The behaviour at different list levels, adding a list glyph, then repeating the text including the [x] leads me to think that this is not behaving as expected.
- There is no test to cover this specifically in [lists_test.rb](https://github.com/asciidoctor/asciidoctor/blob/35c1f9a8f7c1cee310d6c1d1ad46ec5ae4702d54/test/lists_test.rb#L5136)
- The behaviour is consistent (as expected) across generated HTML and in VSCode preview using AsciidoctorJS.

Example:

```asciidoctor
* [x]     Closing bracket followed by space, renders as expected.
* [x]	   <tab><space..>after closing bracket. Fail.
**      [x]	   As above, second level: A 2nd level checklist glyph is rendered, plus the text `[x] As above ...`
* [x]     Closing bracket followed by space, renders as expected.
```
Render of the above:
image

Is this working as designed?

mojavelinux:


sturtison:
Your test is far more comprehensive on the [ ], [x] and [*].

The test I wrote seems to be complementary, testing and showing correct and incorrect uses of tab and space in checklist items.

I am concerned about bloating the test code with my additional test.

The pull request was against your personal test branch and I apologise if that is not the correct way to collaborate on revisions, this being my first OS project contribution. The change is small enough for me to drop, redo, and/or suggest another way.

mojavelinux:
I'll take the additional test. It certainly can't hurt. The more coverage the better.

You did it just right. For some reason GitHub doesn't tell me the PR is there, but your comment allowed me to find it.

PR #3329, opened 2019-06-09
Warn on auto-generated ID conflicts. Fix #3147.

### LINKED ISSUE #3147: Warn on automatically generated ID duplicates
At https://github.com/asciidoctor/asciidoctor/pull/2308 explicit ID duplicates warn.

I would prefer if it were possible to also have that for automatically generated IDs.

While it might make TOCs a bit more annoying as you have to manually differentiate them, I would rather have that.

The reason is that when duplicates happen:

- your explicit links `<<>>` might go to an unexpected place and you won't notice
- if you have duplicate titles, it is likely that you actually want to make their ID's more semantic and specific. But you won't notice it without warning.

cirosantilli:
This is just a quick RFC that more or less reaches the behaviour I would like at: https://github.com/asciidoctor/asciidoctor/issues/3147

If you run `asciidcotor a.readme` with:

```
= bla
:idprefix:
:idseparator: -
:sectanchors:
:sectlinks:
:sectnumlevels: 6
:sectnums:
:toc: macro
:toclevels: 6
:toc-title:

toc::[]

== asdf qwer

== asdf qwer
```

it produces my desired warning:

```
asciidoctor: WARNING: auto-generated header id conflict: asdf-qwer
```

If this behaviour were in general accepted, the following changes would need to be done to this patch:

- add testing
- add line number to warning (not sure how)
- possibly convert to `info`. TODO: `logger.info` does not show on terminal, how to enable it from CLI?

### DIFF
--- a/lib/asciidoctor/section.rb
+++ b/lib/asciidoctor/section.rb
@@ -209,6 +209,7 @@ class Section < AbstractBlock
       gen_id = gen_id.slice 1, gen_id.length if pre.empty? && (gen_id.start_with? sep)
     end
     if document.catalog[:refs].key? gen_id
+      logger.warn %(auto-generated header id conflict: #{gen_id})
       ids = document.catalog[:refs]
       cnt = Compliance.unique_id_start_index
       cnt += 1 while ids[candidate_id = %(#{gen_id}#{sep}#{cnt})]


mojavelinux:
I'd consider this if it were either a info and debug message. I also think it should come after the ID is resolved so it can cite the original ID (the duplicate one) and the reconciled ID (the deduplicated one).

lib/asciidoctor/section.rb
```diff
@@ -209,6 +209,7 @@ def self.generate_id title, document
       gen_id = gen_id.slice 1, gen_id.length if pre.empty? && (gen_id.start_with? sep)
     end
     if document.catalog[:refs].key? gen_id
+      logger.warn %(auto-generated header id conflict: #{gen_id})
       ids = document.catalog[:refs]
       cnt = Compliance.unique_id_start_index
       cnt += 1 while ids[candidate_id = %(#{gen_id}#{sep}#{cnt})]
```
mojavelinux: Put the message here instead as:

```
logger.info %(auto-generated ID already in use: #{gen_id}; deduplicated as #{candidate_id}) if logger.info?
```

Also, we'll want to include line information. That may require passing the reader as an additional argument. I'll have to explore that to see if it's feasible.

PR #3330, opened 2019-06-09
Fix documentation for options passed to SyntaxHighlighter constructor

robertpanzer:
Just a very small fix for the documentation of `SyntaxHighlighterFactory::create`.
The document is passed with the key `:document` instead of `:doc`:

https://github.com/asciidoctor/asciidoctor/blob/35c1f9a8f7c1cee310d6c1d1ad46ec5ae4702d54/lib/asciidoctor/document.rb#L1232-L1238


mojavelinux:
Thanks for catching this, @robertpanzer! You're absolutely correct. I kept going back and forth in choosing a name for this option, and the API docs couldn't keep up ;)

Issue #1735, opened 2016-04-27
Just a feature suggestion for KaTeX


dkotrada:
https://khan.github.io/KaTeX/
KaTex should be better than mathjax for pdf generation.

mojavelinux:
I do want to consider KaTeX integration. The challenge, however, is that KaTeX doesn't support AsciiMath to my knowledge. Now that we have the asciimath gem, however, we could convert the AsciiMath expressions to something that KaTeX understands (perhaps either MathML or LaTeX).

Note that for PDF generation, there is https://github.com/tstumm/asciidoctor-mathematical.

jxxcarlson:
I’m going to look into this again.  When I did a while back (at least a year),
KaTeX was too immature to consider.  I just tried a few random math
expressions that had raised concerns on math overflow and stack overflow,
e.g, typesetting matrices and use of some of the AMS fonts.  In the few
tests that I did, the output looked OK.

This is very superficial testing, but it looks like they have made much progress
since then.  I’m going to look into this further to see if there is a way
to use it with https://github.com/asciidoctor/asciidoctor-latex <https://github.com/asciidoctor/asciidoctor-latex> and
http://www.scripta.io. The long rendering times for mathjax with large
documents are indeed a problem, e.g., http://www.scripta.io/compiled/1410 ,

Since scripta is running Asciidoctor-LaTeX on the server side, this could
be a good solution for it. What do you think?

mojavelinux:
Thanks Jim! I look forward to hearing your assessment!

No doubt, KaTeX has a design that fits more with our needs, so in terms of
integration, it opens up a lot of possibilities.

zaxebo1:
katex seems to be really fast!

just a small clarification query:
Say, a person is reading local filesystem's asciidoc documents using asciidoctor.js addon of firefox browser on  a computer which is not connecting to internet.
 In that case , will katex be able to render, or we will have to fallback on mathjax

oddhack:
Pinging to re-evaluate this - KaTeX has evolved considerably in the last year. I believe its current function support list is here: https://github.com/Khan/KaTeX/wiki/Function-Support-in-KaTeX

[...]

PR #3338, opened 2019-06-13
PROTOTYPE katex server side stem math rendering in HTML5

cirosantilli:
Related issue: https://github.com/asciidoctor/asciidoctor/issues/1735

Server side rendering is awesome as it is faster on the browser, so the page doesn't keep reflowing if you have a ton of formulas

This is now just a prototype, first install katex with:

```
npm install -g katex
```

then convert the `a.adoc` test from this PR.

Outcome: the block katex works, inline not yet.

If people think that this is of interest, I am willing to clean it up into a proper version with some help, since this is relatively high interest to me. Otherwise I'll likely just extract into an extension.

TODOs:

- [] inline not working, don't know how to make `def convert` get an `inline_katex` block yet
- [] `stem: katex` not working yet to make it the default for `stem:` throughout document
- [] error handling if katex not installed / other errors
- [] only include katex if `stem: katex` or if there is at least one `katex:` in document. LIkewise for mathjax.
- [] maybe it would be cleaner to implement katex calls with: https://github.com/Shopify/schmooze instead of
- [] proper testing

Bibliography:

- https://stackoverflow.com/questions/39269611/display-mathjax-equations-smoothly-and-only-once-they-are-fully-rendered
- https://github.com/mathjax/MathJax-node
- https://meta.stackexchange.com/questions/268308/implement-server-side-mathjax-rendering
- https://meta.mathoverflow.net/questions/2360/can-we-replace-client-side-mathjax-with-server-side-mathjax

### DIFF
--- a/a.adoc
+++ b/a.adoc
@@ -1,0 +1,46 @@
+= bla
+:stem:
+
+stem: stem:[sqrt(1+1) = 2]
+
+asciimath: asciimath:[sqrt(1+1) = 2]
+
+latexmath: latexmath:[\sqrt{1+1} = 2]
+
+katexmath: katexmath:[\sqrt{1+1} = 2]
+
+stem: stem:[[[a,b\],[c,d\]\]]
+
+asciimath asciimath:[[[a,b\],[c,d\]\]]
+
+latexmath: latexmath:[\begin{bmatrix}a & b\\c & d\end{bmatrix}]
+
+katexmath: katexmath:[\begin{bmatrix}a & b\\c & d\end{bmatrix}]
+
+stem:
+
+[stem]
+++++
+sqrt(1+1) = 2
+++++
+
+asciimath:
+
+[asciimath]
+++++
+sqrt(1+1) = 2
+++++
+
+latexmath:
+
+[latexmath]
+++++
+\sqrt{1+1} = 2
+++++
+
+katexmath:
+
+[katexmath]
+++++
+\sqrt{1+1} = 2
+++++
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -283,7 +283,7 @@ module Asciidoctor
     '====' => [:example, ['admonition'].to_set],
     '****' => [:sidebar, ::Set.new],
     '____' => [:quote, ['verse'].to_set],
-    '++++' => [:pass, ['stem', 'latexmath', 'asciimath'].to_set],
+    '++++' => [:pass, ['stem', 'latexmath', 'asciimath', 'katexmath'].to_set],
     '|===' => [:table, ::Set.new],
     ',===' => [:table, ::Set.new],
     ':===' => [:table, ::Set.new],
@@ -342,11 +342,13 @@ module Asciidoctor
   BLOCK_MATH_DELIMITERS = {
     asciimath: ['\$', '\$'],
     latexmath: ['\[', '\]'],
+    katexmath: ['', ''],
   }
 
   INLINE_MATH_DELIMITERS = {
     asciimath: ['\$', '\$'],
     latexmath: ['\(', '\)'],
+    katexmath: ['', ''],
   }
 
   (STEM_TYPE_ALIASES = {
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -122,6 +122,10 @@ class Converter::Html5Converter < Converter::Base
     end
     result << %(<title>#{node.doctitle sanitize: true, use_fallback: true}</title>)
 
+    if node.attr? 'stem'
+      # TODO only if katex is actually used.
+      result << %(<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.2/dist/katex.min.css" integrity="sha384-yFRtMMDnQtDRO8rLpMIKrtPCD5jdktao2TV19YiZYWMDkUR5GQZR/NOVTdquEx1j" crossorigin="anonymous">)
+    end
     if DEFAULT_STYLESHEET_KEYS.include?(node.attr 'stylesheet')
       if (webfonts = node.attr 'webfonts')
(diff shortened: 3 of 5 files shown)


mojavelinux:
I haven't taken a deep look, but is this a better approach than https://github.com/jirutka/asciidoctor-katex?

cirosantilli:
@mojavelinux ah thanks, I hadn't seen that one. 

After a quick look, basically, the only thing they can be doing fundamentally significantly better is using https://github.com/glebm/katex-ruby which uses https://github.com/rails/execjs to call katex instead of popen3 as I do here (relates to the  https://github.com/Shopify/schmooze point above, but katex ruby gem is even better)

My approach avoids adding a lot of dependencies to this project, their approach likely runs faster in a document with a ton of maths since it should not start a process for every math like I do here.

I will benchmark this on a huge test document vs the existing mathjax to see if I can observe a significant performance difference, if not I would recommend just starting with pipes due to simplicity.

cirosantilli:
I did a benchmark as follows:

```
# N equations.
n=1000

# asciidoctor-katex: asciidoctor 0.2.10, asciidoctor-katex 0.3.0, katex@0.10.2
(printf '= bla\n:stem:\n:docinfo: shared\n\n'; i=0; while [ $i -lt $n ]; do printf '[latexmath]\n++++\n\sqrt{1+1} = 2\n++++\n\n'; i=$((i+1)); done) > bench.adoc
printf '<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.10.2/dist/katex.min.css" integrity="sha384-yFRtMMDnQtDRO8rLpMIKrtPCD5jdktao2TV19YiZYWMDkUR5GQZR/NOVTdquEx1j" crossorigin="anonymous">' > docinfo.html
time asciidoctor -r asciidoctor-katex bench.adoc

# this patch
(printf '= bla\n:stem:\n\n'; i=0; while [ $i -lt $n ]; do printf '[katexmath]\n++++\n\sqrt{1+1} = 2\n++++\n\n'; i=$((i+1)); done) > bench2.adoc
time ./bin/asciidoctor bench2.adoc

# asciidoctor 0.2.10
(printf '= bla\n:stem:\n\n'; i=0; while [ $i -lt $n ]; do printf '[latexmath]\n++++\n\sqrt{1+1} = 2\n++++\n\n'; i=$((i+1)); done) > bench3.adoc
time ./bin/asciidoctor bench3.adoc
```

Results:

- `asciidoctor-katex`: 362 seconds
- this patch: 82 seconds
- `asciidoctor`: 0.1 second

So to my surprise, my naive pipe approach was about 4.5x faster!

The it is interesting to open all the output documents:

- `asciidoctor-katex` and "this patch", look exactly the same, load immediately
- `asciidoctor`: takes 2 seconds to render all math

So yeah, I now found that Katex performance was so bad, I likely won't even be able to use it all... I'll bring this up with the project to see if I'm doing something wrong.

Tested on: Lenovo ThinkPad P51 laptop with CPU: Intel Core i7-7820HQ CPU (4 cores / 8 threads), RAM: 2x Samsung M471A2K43BB1-CRC (2x 16GiB, 2400 Mbps), SSD: Samsung MZVLB512HAJQ-000L7 (512GB, 3,000 MB/s).

mojavelinux:
> So to my surprise, my naive pipe approach was about 4.5x faster!

I'm not all that surprised by this because Ruby system integration is very good. You are going to get the best result by using the native calls rather than introducing layers in between.

mojavelinux:
The main issue with this change is going to be security. Right now, Asciidoctor doesn't use any libraries (that the user doesn't supply) and doesn't make any system calls. This is a hard requirement for Asciidoctor working on GitHub.

We could tie this into the safe mode so that it only works in unsafe mode (to start). We will also probably need a very lightweight adapter / manager class for calling katex so that it can be replaced in different environments with a different strategy (perhaps using a service or something). It would probably be something along the lines of what we have for the syntax highlighter integrations.

cirosantilli:
Yes, this is definitely unsafe only stuff, I'm considering it for GitHub pages.

Although at this speed, I wouldn't even bother supporting it... it would make the dev cycle unbearable.

Shame, since server side math is kind of the holy grail of web maths, so sad.

cirosantilli:
Dan, I've updated [the benchmark comment](https://github.com/asciidoctor/asciidoctor/pull/3338#issuecomment-502045350) with new findings.

As expected KaTeX slowness is only due to startup: once loaded renders are fast, so it is a "perf bug" on asciidoctor-katex's integration / stack.

Then I tried schmooze instead of execjs, and I got good usable perf!

I understand your concerns about this feature, let me know if you think it is worth merging to master or if I should just start with an extension.

The extension I think I already know how to do basically :-) Master merge a bit more involved but I will do with help, as this is math nirvana.

mojavelinux:
That's great news about the perf. 🚀

As to whether this should go in core or be an extension, that's a great
(and tough) question. I really need to think hard about that one, but I
really do want to make this easy on users, so I will take that into
consideration. Regardless, it should be part of the Asciidoctor org).

Thanks for getting this moving! I really appreciate your work. 😁

>

cirosantilli:
A quick progress report:

At https://github.com/cirosantilli/cirosantilli.github.io/blob/fd11b321c5e4075509db2c4d52249c94d90040bd/katex.rb I have pushed the plugin as far as I can go without solving "possibly not easy upstream questions" mentioned in the TODO part of that file:

* make KaTeX `\newcommand` work across math blocks, not supported on KaTeX 0.10.2 upstream: https://github.com/KaTeX/KaTeX/issues/2070
* Equation `#num` prefix to caption descriptions and xrefs. Upstream issue: https://github.com/asciidoctor/asciidoctor/issues/3378
+
This could of course be solved if the plugin were merged into mainline, but it would be cool if it were doable in an extension.

Besides those however, it the extension is already working pretty well, and if those points were implemented, we would have, I believe, the best HTML typesetting system created so far, opening the way to destroy LaTeX and world domination.

Even without those points I think it is already good enough for me to publish it as a gem.

Issue #3394, opened 2019-08-20
Issue loading the syntax_highlighter


aerostitch:
Hi,

Asciidoctor 2.0.10 has finally entered Debian archive and we're seeing a bug report that I'm not sure where the problem would be: https://bugs.debian.org/935196
It looks like for whatever reason it's looking for the syntax_highlighter in /usr/lib/ruby/lib/asciidoctor/syntax_highlighter instead of /usr/lib/ruby/vendor_ruby/asciidoctor/syntax_highlighter

Have you ever experienced that issue?

Thanks,
Joseph

mojavelinux:
Is the Debian package modifying the LIB_DIR (or ROOT_DIR) by chance?

Here's where the syntax highlighter is loaded: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L514

mojavelinux:
There's no relationship between how the syntax highlighter was loaded before 2.0 with how it is loaded after 2.0. The mechanism completely changed. So there's no use in saying "it used to work" because it isn't even the same code. It's more than something new (the autoload) doesn't work correctly in the Debian environment.

mojavelinux:
I noticed that we're not doing autoload the same way as in the converter.

* https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L514-L515
* https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter.rb#L44-L45

See if changing LIB_DIR to __dir__ fixes the issue. Then I'll go ahead and make that change for 2.0.11.

aerostitch:
Thanks a lot fo ryour help, I'll let you know in a bit.

aerostitch:
Your solution worked, thanks! :)
Pushed a PR with the fix.

PR #3395, opened 2019-08-20
Fix autoload of the syntax highlighter in specific installations like Debian

aerostitch:
Tested in Debian, this fixes our issue.
Thanks for the pointer @mojavelinux ! :) (I did the PR so that you didn't have to ;)

Closes: #3394


mojavelinux:
Thanks @aerostitch!

PR #3416, opened 2019-09-13
Fix #3414 Vim Errorformat integration: relative or full path to file

### LINKED ISSUE #3414: Add option to include full path to file in application messages
Imagine I have a document `~/docs/test/test.adoc`.

My current directory is `~/docs` and I issue the `asciidoctor-pdf` command:
```
asciidoctor-pdf "~/docs/test/test.adoc"
```

If there is an error in a file I get a message like:

```
asciidoctor: ERROR: test.adoc: line 20: level 0 sections can only be used when doctype is book
```

File `test.adoc` in the message is without any path specified which makes it impossible to jump to the error from within vim. (It just goes to new `~/docs/test.adoc` file instead).

Is it possible to tell asciidoctor to print full or relative to the directory where asciidoctor was executed file/path?

PS, I tried to specify `--source-dir "~/docs"` but it doesn't change anything.

PPS, vim errorformat I try to use:

`setlocal errorformat=asciidoctor:\ ERROR:\ %f:\ line\ %l:\ %m`

and it looks in vim:

image

habamax:
Hi @mojavelinux , hope I didn't forget anything important.


habamax:
Well, apparently I forgot to run test after fix :( .

habamax:
Hmm, I am lost in `assert_message` failures.

Now instead of hundreds of failures there are only 5:

```
    5) Failure:
  Minitest::Test::TestLogger::TestLogging#test_can_create_an_auto_formatting_message_with_context 
  [C:/Users/maksim.kim/projects/asciidoctor/test/logger_test.rb:196]:
  --- expected
  +++ actual
  @@ -1 +1 @@
  -"file.adoc: line 5: Asciidoctor was here"
  +"C:/Users/maksim.kim/projects/asciidoctor/test/fixtures/file.adoc: line 5: Asciidoctor was here" 

```

Am I doing it right? :)

habamax:
Tried to fix tests... Fixed all except 1:

```
      test 'nested remote include directive that cannot be resolved does not crash processor' do
```
Can't figure out how to satisfy it :(

habamax:
So the last test fail was about treating `@path` as URI which caused File.join to raise exception.

habamax:
Not sure how to unmerge and rebase against the master :(

mojavelinux:
once you submit a PR, you should never need to change it unless there's a direct line conflict with something that went into master. git can sort out how to merge it. GitHub will tell you when there is a conflict.

habamax:
Ok, so now it is in good shape? No need to do rebasing?

mojavelinux:
No need to rebase. But we will need to discuss the code itself ;) I'll review at work tomorrow.

PR #3418, opened 2019-09-16
resolves #3970 semantic HTML 5 converter (paragraph)

### LINKED ISSUE #3970: Use semantic constructs for paragraphs
This issue will be addressed as part of: #242

## Paragraph

I think this is an easy one, we should enclose the content in a [`<p>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p) element.

```html
<p><!-- content --></p>
```

### Role

I think we shouldn't add a default role on paragraph blocks but rather always use specific roles when a `<p>` element is used in other contexts.

### Title

In AsciiDoc, you can assign a title to a paragraph:

```
.Famous intro
A long time ago in a galaxy far, far away.
```

#### strong element

I think we should avoid enclosing the paragraph element in a meaningless container (such as `<div>`).
Since a `<p>` element only permits phrasing content, I believe that `<strong>` is a good choice.
In publishing, the formal name for this construct is a "run-in heading". See:
- https://docs.microsoft.com/en-us/style-guide/scannable-content/headings#using-run-in-headings
- http://www.creativeglossary.com/graphic-design/run-in-heading.html. 

> A run-in heading is a heading positioned upon the very same line as the text. Frequently this text is put in either italic or bold type.

The advantage of strong is that it comes prestyled (compared to span).

```html
<p><strong class="title">Famous intro</strong> A long time ago in a galaxy far, far away.</p>
```

image

In the above example, the strong element has a default role "title".
Instead, we could also use the name "head" but I feel like "title" is probably more obvious.

##### heading element

While it may seem like the right solution, my concern is that we would use the same construct for level 6 section titles and block titles. Also, keep in mind that:

> heading information can be used by user agents to construct a table of contents for a document automatically.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/heading_elements#usage_notes

So it would mean that block titles will be part of the document outline.

In addition, an heading element is not a phrasing content so the heading element will inevitably be outside of the paragraph element:

```html
<h6>Famous intro</h6>
<p>A long time ago in a galaxy far, far away.</p>
```

Mogztter:
Implement the `convert_paragraph` method for the new semantic HTML 5 converter.
Hopefully it will serve to generate discussion on which semantic construct we should use and on implementation details.

This pull request should not be merged into master but into a dedicated branch.

resolves #3970


test/fixtures/sem-html5-scenarios/paragraph-with-id-roles.adoc
```diff
@@ -0,0 +1,2 @@
+[#exciting.lead.awesome]
```
Mogztter: I wonder if we should deduplicate roles.
With the current implementation, if I define the following:

`[#exciting.lead.paragraph]` 

Then `paragraph` will be present twice:

```html
<p id="exciting" class="paragraph lead paragraph">
```

test/fixtures/sem-html5-scenarios/paragraph-with-title.html
```diff
@@ -0,0 +1,4 @@
+<p class="paragraph">
+<strong class="title">Famous intro</strong>
```
Mogztter: Since permitted content in a paragraph are [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content), we could use `<label>` instead of the `<strong>` tag:

```html
<p class="paragraph">
<label class="title">Famous intro</label >
A long time ago in a galaxy far, far away.
</p>
```

> The HTML <label> element represents a caption for an item in a user interface.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

Mogztter: For reference, we cannot use heading tag because it's not permitted in a paragraph.

mojavelinux: Putting the title/caption inside the p is definitely the approach we want to take as it removes the complication (and bad taste) introduced by enclosing the p tag.

I agree the heading tag won't work here. But I don't think label is the right choice either. That's really reserved for a form, so we would be diluting its meaning. I think the safe choice here is either span or strong. The advantage of strong is that it comes prestyled.

What Asciidoctor EPUB3 uses is `<strong class="head">`. Though if "head" seems too obtuse, I would be supportive of the class name "title".

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+    init_backend_traits basebackend: 'html', filetype: 'html', htmlsyntax: syntax, outfilesuffix: '.html', supports_templates: true
+  end
+
+  def convert_embedded node
+    node.content
+  end
+
+  def convert_paragraph node
+    attributes = html_attributes node.id, node.role, 'paragraph'
```
mojavelinux, line 23: I'm not convinced whether the "paragraph" class on the p is necessary. I fear it might leave the same bad taste as the div wrapper in the current HTML converter does. But I can understand the motivation for including it. It allows us to distinguish it from p tags used in other contexts. I wonder if simply marking with the role "block" would serve that purpose.

Note that Asciidoctor EPUB3 does not add a class to the p by default.

mojavelinux: Also note that the DocBook stylesheets do not add an implicit class to normal paragraphs either.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+module Asciidoctor
+# A built-in {Converter} implementation that generates HTML 5 output
+# that maximizes the use of semantic constructs.
+class Converter::SemanticHtml5Converter < Converter::Base
+  register_for 'semantic-html5', 'sem-html5'
```
mojavelinux, line 6: Let's make a separate issue for the backend name. I'm fine if we have a working name in the interim, but I don't think we'll land on semantic-html5 when it's all said and done.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: Then maybe we shouldn't add a default role on paragraph blocks but rather always use specific roles when a `<p>` element is used in other contexts?

test/fixtures/sem-html5-scenarios/paragraph-with-title.html
Mogztter: > But I don't think label is the right choice either.

I came to the same conclusion in another issue. I don't think we should use `<label>`.

Mogztter: > Though if "head" seems too obtuse, I would be supportive of the class name "title".

I can see the reason for that naming. We could also use `lead` since the title is now inside the paragraph and "leading" the text but title is probably less disruptive and more obvious.

test/fixtures/sem-html5-scenarios/paragraph-with-id-roles.adoc
mojavelinux: This gets back to whether we should be adding an implicit role in the first place. But even if we do, I don't think we should worry about it since it's not invalid HTML.

test/fixtures/sem-html5-scenarios/paragraph-with-title.html
mojavelinux: In publishing, the formal name for this construct is a "run-in heading". See https://docs.microsoft.com/en-us/style-guide/scannable-content/headings#using-run-in-headings and http://www.creativeglossary.com/graphic-design/run-in-heading.html. That's how I ended up with "head" in Asciidoctor EPUB3. "lead" actually has a very specific meaning in publishing and not something we can use here.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: > Then maybe we shouldn't add a default role on paragraph blocks but rather always use specific roles when a <p> element is used in other contexts?

Yes! That's a brilliant idea.

test/fixtures/sem-html5-scenarios/paragraph-with-title.html
mojavelinux: The DocBook stylesheets use the strong tag for the run-in heading. I really think that's the best choice.

We may also consider adding the ability to control the "stop" character that follows the run-in heading. Though that certainly should be a separate issue.

elextr: Late to the discussion, but isn't it a significant change to the semantics to change a paragraph title to a run in?

mojavelinux: I would argue it's always been a run-in heading. That's what the DocBook stylesheets produce. So technically, it was the built-in converter that was doing it incorrectly. We've also now done it this way in Asciidoctor EPUB3 for many years, so it's something that is very common already.

elextr: Fair enough, except its now just a piece of bolded text, its actually lost its title semantic.  That means for example a screen reader won't identify it as a title.

elextr: My CSS is pretty poor but `run-in` is a CSSable effect IIUC and it can be applied to an element before the `<p>` so that element could retain `<header>` semantics?

Mogztter: I don't think we should rely on `run-in`. MDN states that this feature is experimental (and should not be used in production code): https://developer.mozilla.org/en-US/docs/Web/CSS/display

The specification is also not very encouraging:

> The following features are at-risk, and may be dropped during the CR period:
> ...
> display: run-in

https://drafts.csswg.org/css-display/#status

If we want to add a relation between the title and the paragraph we could use [`aria-describedb`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) / [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) but we would need to generate an id.

>

We can add `role=heading`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role
I don't own a screen reader, so I'm not sure what are our alternatives if we don't want or cannot use a `<h6>`.


For reference, here's what we produce today:

```html
<div class="paragraph">
  <div class="title">Famous intro</div>
  <p>A long time ago in a galaxy far, far away.</p>
</div>
```

elextr: > I don't think we should rely on run-in. MDN states that this feature is experimental
> Could you elaborate on your proposal and post the HTML it would produce?

See I told you my CSS was poor, and my HTML is not much better :grin:

Really what I'm trying to do is to jog you experts that the semantics of title is being lost if its converted to "just bolded text".  

Related I guess is: what does a title on other blocks like delimited ones translate to in semantic HTML?

> For reference, here's what we produce today:

But of course thats a non-semantic converter :grinning:

Mogztter: > Related I guess is: what does a title on other blocks like delimited ones translate to in semantic HTML?

It depends, I believe that we are going to use [`<figcaption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) on images (and potentially other blocks) and [`<caption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption) on tables, but a lot is still undecided...
Having said that, if your question is "do we want to use the same semantic structure on all blocks?" then the answer is probably no.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: @mojavelinux Not sure how I should phrase it. Maybe we should create an issue to discuss the transition?
- find a backend name for this new converter
- discuss when and how this new converter will become the default one
- find a backend name for the current HTML5 converter if we decide to rename it

mojavelinux: Yes please. I'm fine with using an interim name as long as we're actively discussing what we want it to be (even if we decide to stick with this one). We may, in fact, need two issues. One to decide on the name(s) (which has certain implications in how intrinsic attributes are computed) and one for how we will manage the transition.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize backend, opts = {}
+    @backend = backend
+    if opts[:htmlsyntax] == 'xml'
```
mojavelinux, line 10: I prefer the use of ternary expressions for single variable assignments, which will likely be enforced by the linter when we set it up.

```
syntax = opts[:htmlsyntax] == 'xml' ? 'xml' : 'html'
```

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -0,0 +1,43 @@
...
+#{node.content}
+</p>)
+    else
+      %(<p#{attributes}>
+#{node.content}
+</p>)
+    end
+  end
+
+  def html_attributes id, role, default_role = nil
```
mojavelinux, line 36: I think this method name should be prefixed with common_ to better describe the purpose it serves.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -0,0 +1,43 @@
...
+    roles = []
```
mojavelinux, line 37: Could you set this up as:

```
roles = default_role ? [default_role] : []
```

mojavelinux (changes requested):

mojavelinux:
I left a few comments regarding code style. Other than that, the implementation looks good.

Since we aren't sure what the backend name would be, I would prefer if we registered it only as `semantic-html5` for now. We'll add any aliases in a later change.

As an aside, I have been wanting to move this test suite to rspec, based on the work I did in Asciidoctor PDF. Let's discuss in a separate issue whether we should switch the tests for this converter over to rspec ahead of the rest of the suite. It might make testing it easier.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: I've created two follow-up issues.

Mogztter:
> Since we aren't sure what the backend name would be, I would prefer if we registered it only as semantic-html5 for now. We'll add any aliases in a later change.

Good point!
I've replaced `sem-html5` by `semantic-html5` everywhere to make it easier to update if/when we decide to use another name.

> As an aside, I have been wanting to move this test suite to rspec, based on the work I did in Asciidoctor PDF. Let's discuss in a separate issue whether we should switch the tests for this converter over to rspec ahead of the rest of the suite. It might make testing it easier.

I've created a follow-up issue: #3973

mojavelinux (approved):

mojavelinux:
Let's merge this so we can get this train out of the station and cruising down the tracks. :bullettrain_side:

mojavelinux:
:tada: :champagne:

Mogztter:
Let's goooooooooooo! 😉

PR #3419, opened 2019-09-16
resolves #3248 allow encoding of include file to be specified using encoding attribute

### LINKED ISSUE #3248: Allow Inclusion of non-UTF-8 Files
> **Feature Request**: add an (optional) attribute in the `include::` directive to control the encoding of the included source file, and allow inclusion of non-UTF-8 source files by automatically converting them first.

Rationale:

I'm currently working on a documentation project that involves sourcecode files encoded in ISO-8859-1, which can't be directly included into AsciiDoc socuments.

In order to include the source files (or parts of them) into the documents via the `include::` directive I need to first run a script that converts them to UTF-8 via __iconv__, and then run the Asciidoctor toolchain and include the UTF-8 version instead. Here's a real case example:

- https://github.com/alan-if/alan-docs/blob/master/alanguide/_adoc/generate-inc-files.sh

Where the script creates a copy of the original sourcefiles (eg. "`mysource.alan`"/"`.i`") converted to UTF-8 ("`mysource.utf8_alan`"/"`.utf8_i`").

This introduces an extra layer of complexity and dependencies, especially on Windows which doesn't have a native tool like __iconv__, adds extra files and complicates managing any watch scripts.

If Asciidoctor were to allow an extra attribute to control the encoding of the included file — e.g. `include::path[encoding=iso-8859-1]` it would be much simpler and elegant.

I know that today most source files are expected to be in UTF-8, but some legacy tools still cling on ISO encodings — and, besides, there are many other encodings still in use today. Being an optional extra feature that doesn't break backward compatibility, this would introduce and added benefit to Asciidoctor. My guess is that there should be Ruby libraries to handle encoding conversion.

mojavelinux:


Mogztter:
Definitely a good quality of life improvement 😉

PR #3439, opened 2019-09-30
resolves #3436 only strip trailing whitespace from lines of AsciiDoc include file 

### LINKED ISSUE #3436: Only normalize included content if target ends in a known AsciiDoc extension
The parser only runs the AsciiDoc preprocessor on included content if the target ends with a known AsciiDoc extension (e.g., .adoc). However, it always normalizes the lines (removes trailing whitespace). This can result in data, such as TSV data, being normalized when it shouldn't be.

The two behaviors should be linked. The lines should only be normalized *if* they are going to be parsed as AsciiDoc.

If necessary, in the future, we may consider an option to force the parser to normalize the lines.

mojavelinux:


mojavelinux:
@lurch The code inside of Asciidoctor's parser is highly tuned for speed and to accommodate variations in different versions of Ruby, so it isn't always pretty or DRY. But it's the way it needs to be.

PR #3450, opened 2019-10-12
resolves #3449 upgrade MathJax to version 3.0.0

### LINKED ISSUE #3449: Upgrade to MathJax 3
> MathJax v3 is a complete rewrite of MathJax from the ground up (see What’s New in MathJax v3.0), and so its internal structure is quite different from that of version 2. That means MathJax v3 is not a drop-in replacement for MathJax v2, and upgrading to version 3 takes some adjustment to your web pages.

https://docs.mathjax.org/en/latest/upgrading/v2.html

Mogztter:


CHANGELOG.adoc
```diff
@@ -37,6 +37,7 @@ Improvements::
   * Preserve guard around XML-style callout when icons are not enabled (#3319)
   * Remove redundant test for halign and valign attributes on table cell in DocBook converter
   * Allow encoding of include file to be specified using encoding attribute (#3248)
+  * Upgrade MathJax to 3.0.0 (#3449)
```
Mogztter: Depending on when this feature get merged, we should probably remove "Upgrade MathJax from 2.7.5 to 2.7.6" in the Bug Fixes list otherwise it can be confusing.

lib/asciidoctor/converter/html5.rb
```diff
@@ -256,33 +256,35 @@ def convert_document node
 
     if node.attr? 'stem'
       eqnums_val = node.attr 'eqnums', 'none'
-      eqnums_val = 'AMS' if eqnums_val.empty?
-      eqnums_opt = %( equationNumbers: { autoNumber: "#{eqnums_val}" } )
+      eqnums_val = 'ams' if eqnums_val.empty?
```
Mogztter, line 259: The value is now case-sensitive and strict in MathJax 3. We need to lowercase the value and check if the value is `all`, `none` or `ams`. If the value is invalid, we should use `ams` as a fallback value to be backward compatible with MathJax 2.7.

PR #3457, opened 2019-10-20
resolves #3456 preserve repeating backslashes when generating manpage output

### LINKED ISSUE #3456: Preserve repeating backslashes in manpage output
Currently, none of the following is working:

```
\\
```
```
\\\
```
```
\\\\
```
```
{backslash}{backslash}
```

Initially reported as a question on Stackoverflow: https://stackoverflow.com/questions/56864050/how-to-escape-double-backslashes-so-i-can-get-in-man-page

mojavelinux:


mojavelinux:
I can't think of a better way to do this given we can't use backreferences (because it would break Asciidoctor.js). The risk here is that we start eating into backslashes for another purposes, but at this point those should be guarded.

Mogztter:
Maybe I should add a test on Asciidoctor.js for extra safety?

mojavelinux:
That's a great idea.

>

PR #3463, opened 2019-10-22
resolves #3462 move style tag for convert-time syntax highlighters into head

### LINKED ISSUE #3462: Code highlighter CSS added in the body tag
Why is the code highlighter CSS style block added at the end of `<body>`? I would expect it in `<head>`.

Test source:
```
:source-highlighter: coderay

= Document Title

[source,ruby]
---- 
require 'sinatra'

get '/hi' do
  "Hello World!"
end
----
```

Output (`asciidoctor -b html5 test.adoc`):
```
<!DOCTYPE html>
<html lang="en">
<head>
...
</head>
<body class="article">
...
<style>
/* Stylesheet for CodeRay to match GitHub theme | MIT License | http://foundation.zurb.com */
...
</style>
</body>
</html>
```

mojavelinux:
- move style tag for convert-time syntax highlighters into head
- invoke docinfo? :head method on syntax highlighter adapter after content has been converted

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -31,6 +31,7 @@ Bug Fixes::
 Compliance::
 
   * Add support for muted option to self-hosted video (#3408)
+  * Move style tag for convert-time syntax highlighters (coderay, rouge, pygments) into head (#3462)
 
 Improvements::
 
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -152,8 +152,8 @@ class Converter::Html5Converter < Converter::Base
       end
     end
 
-    if (syntax_hl = node.syntax_highlighter) && (syntax_hl.docinfo? :head)
-      result << (syntax_hl.docinfo :head, node, cdn_base_url: cdn_base_url, linkcss: linkcss, self_closing_tag_slash: slash)
+    if (syntax_hl = node.syntax_highlighter)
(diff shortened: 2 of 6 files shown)


Mogztter:
@mojavelinux What about Highlight.js? 

https://github.com/asciidoctor/asciidoctor/blob/d1aee25c12556cd11d813b8964944c8679c61c41/lib/asciidoctor/syntax_highlighter/highlightjs.rb#L20-L23

The script should be at the end of the `<body>` but the stylesheet should be in the `<head>` right?

mojavelinux:
I decided not to change it for highlight.js since it's a) legal to put the link tag in the body and b) the styles aren't needed until after the script loads. The convert-time highlighters need the style in the head since the HTML is already there when the page loads, waiting to be styled.

Mogztter:
Makes sense, thanks for the explanation 👍 

I was asking because for a very specific use case I needed to override a style defined in the highlight.js theme. But since `github.min.css` is loaded last, I cannot override the style in my custom stylesheet unless I use `!important`.

mojavelinux:
Hmm. Yeah. That is a downside. Though you can use docinfo, if that's an option for you.

Mogztter:
Another issue: https://github.com/Mogztter/asciidoctor-pdf.js/pull/122

I think we should definitely consider loading styles in the `<head>`.

/cc @ahus1

mojavelinux:
Please open a separate issue so we don't lose track of it.

PR #3484, opened 2019-11-14
escape &<> in attributes

Hextremist:
this fixes: "asciidoctor: ERROR: failed to parse formatted text ..."
for asciidoctor-pdf when using e.g. & in an attribute.

Example file:

== Demo
Oh why oh why is {whatever} causing headaches?

[kjell@localhost ~]$ asciidoctor-pdf -a whatever="Fish & Chips" t.adoc
asciidoctor: ERROR: failed to parse formatted text: Oh why oh why is Fish & Chips causing headaches?

I am not sure if this is the right way to fix the problem, but it seems to work fine.

### DIFF
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -232,7 +232,7 @@ module Substitutors
           @document.counter(*args)
         end
       elsif doc_attrs.key?(key = $2.downcase)
-        doc_attrs[key]
+        sub_specialchars doc_attrs[key]
       elsif (value = INTRINSIC_ATTRIBUTES[key])
         value
       else


Hextremist:
My initial comment did not show any ill effects in the output, but try to e.g. use italics or bold in the same line. Output in pdf will be something like <strong>Fish&Chips</strong>

Hextremist:
I reported an issue on asciidoctor-pdf instead:
https://github.com/asciidoctor/asciidoctor-pdf/issues/1524

mojavelinux:
I appreciate what you're trying to accomplish here, but this would violate substitution rules in a major way. At best, the operation would be redundant. At worst, it would completely disable the ability to passthrough content using attributes, which many users do.

I have explained the rule in this issue: https://github.com/asciidoctor/asciidoctor-pdf/issues/1524#issuecomment-578532039. I think this warrants better documentation.

Suffice to say, attribute values passed via the API or CLI must be pre-escaped unless the intention is to pass through content which is not escaped. (You can think of it as though passing an attribute via the API or CLI has an implicit `pass:[]` around the value).

PR #3487, opened 2019-11-19
trigger build of AsciidoctorJ job running as GitHub Action

robertpanzer:
This PR changes the build:dependents task to trigger the upstream build on Github instead of on Travis CI.
To get this running on TravisCI it is necessary to provision a Github auth token in the environment variable `GITHUB_TOKEN`.
This needs to have the `repo` scope.

Even though the tasks of triggering a build in Travis and Github are very similar, I duplicated the `:dependents` task, as I currently think it would make the logic too complex to distinguish between Github and Travis in the same task.
There are a few differences:
- Obviously a different URL to call
- The payload differs
- The request headers differ, including the auth token
- The response code for a successful trigger differs

Testing locally it seems like 2 `task`-s with the same name are called one after the other on `rake build:dependents`.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -45,6 +45,7 @@ Improvements::
 Build / Infrastructure::
 
   * Run test suite on TruffleRuby nightly (*@mogztter*, *@erebor*)
+  * Trigger upstream builds for AsciidoctorJ on Github Actions (*@robertpanzer*)
 
 // tag::compact[]
 == 2.0.10 (2019-05-31) - @mojavelinux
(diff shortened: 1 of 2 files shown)


mojavelinux:
The task "dependents" is now defined twice. Could you move the common logic to a function and then call that function with different arguments depending on whether it is invoking Travis or GitHub Actions?

robertpanzer:
Updated the PR by moving triggering of build at TravisCI and GH actions into the same task and factoring out the common code.

mojavelinux:
Thanks!

PR #3493, opened 2019-11-25
resolves #3491 allow template to override outline

### LINKED ISSUE #3491: Allow plugging "convert_outline"
I'd like to be able to override the TOC conversion behavior, at least some. Right now the html5 converter has `convert_outline doc` but if it did something like `doc.converter.convert doc, 'outline'` then I could plug in my conversions. Is that a thing you'd be ok doing?

nik9000:
Allows plugging templates in to the html5 converter's "outline" by
replacing all direct calls to `convert_outline` with calls to the to the
conversion named `outline`.

Resolves #3491


nik9000:
@mojavelinux, I'm no sure I've added enough tests for this, or if I've put the test that I did add in the right place, but I think is a start.

mojavelinux:
Looking good. I made a few tweaks, but you've got it.

nik9000:
Thanks @mojavelinux !

PR #3497, opened 2019-11-30
resolves #3793 don't apply border-collapse: separate to HTML for table blocks

### LINKED ISSUE #2563: Missing Border in Multirow-Cell
With Asciidoctor 1.5.6.1 in Manjaro, if i create a Table with Cells that spann multiple rows, each second Cell has no right border.

```a
Schedule
========
Testcase

[cols=">1h,^.^4,^.^4",options="header"]
|====
|           |Day 1       |Day 2
|08:00   .2+|Task 1   .4+|Task 4
|09:00
|10:00   .2+|Task 2
|11:00
|12:00    2+|*Lunchhour*
|13:00   .4+|Task 3   .4+|Task 5
|14:00
|15:00
|16:00
|17:00
|====
```
becomes:
schedule

The reason lies in
```html
table.grid-all>*>tr>.tableblock:last-child,table.grid-cols>*>tr>.tableblock:last-child{border-right-width:0}
```
"border-right-width" should be 1px instead of 0.

This behavior does not appear in Asciidoc and Asciidoctor-PDF.
Weirdly, it appears in [AsciidocLive](https://asciidoclive.com/) too (does it use Asciidoctor instead of Asciidoc?)

I dunno how to fix that, except in the HTML-File itself. But it seems easily fixable.

### LINKED ISSUE #3793: Don't apply border-collapse: separate to HTML for table block
For tables without a rowspan or colspan, the borders display the same in HTML regardless of whether border-collapse is separate or collapse. However, when there is a rowspan or colspan, with certain combinations of frame and grid, the border can get doubled at the boundary of a rowspan or colspan. By not applying border-collapse: separate to these tables, it mitigates the problem.

ahus1:
I found that applying border-collapse solves the disappearing cell borders. I tried to trace the history of this CSS settings, but wasn't able to determine the case why this was added here 58149743db9f59d4036879ff73da5b06eaf9ae9e

An example like this

```asciidoc
[options="header"]
|===
| Column 1 | Column 2 | Column 3
.2+|Content in a single cell that spans rows 1 and 2  |Cell in column 2, row 1 .2+|Content in a single cell that spans rows 1 and 2
|Cell in column 2, row 2
|===
```

that rendered before in Chrome like this using the asciidoctor CSS from master 

image

now renders like this

image

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -44,6 +44,7 @@ Bug Fixes::
   * Use custom init function for highlight.js to select the correct `code` elements (#3761)
   * Fix resolved value of :to_dir when both :to_file and :to_dir options are set to absolute paths (#3778)
   * Restore label in front of each bibliography entry in DocBook output that was dropped by fix for #3085 (#3782)
+  * Don't apply border-collapse: separate to HTML for table blocks; fixes double border at boundary of colspan/rowspan (#3793) (*@ahus1*)
 
 Compliance::
 
--- a/data/stylesheets/asciidoctor-default.css
+++ b/data/stylesheets/asciidoctor-default.css
@@ -251,7 +251,7 @@ pre.pygments .lineno::before{content:"";margin-right:-.125em}
 .quoteblock.excerpt,.quoteblock .quoteblock{margin-left:0}
 .quoteblock.excerpt blockquote,.quoteblock.excerpt p,.quoteblock .quoteblock blockquote,.quoteblock .quoteblock p{color:inherit;font-size:1.0625rem}
 .quoteblock.excerpt .attribution,.quoteblock .quoteblock .attribution{color:inherit;text-align:left;margin-right:0}
-table.tableblock{max-width:100%;border-collapse:separate}
+table.tableblock{max-width:100%}
 p.tableblock:last-child{margin-bottom:0}
 td.tableblock>.content{margin-bottom:1.25em}
 td.tableblock>.content>:last-child{margin-bottom:-1.25em}


mojavelinux:
Unfortunately, it's not that simple. That impacts many other frame/grid permutations. See https://github.com/asciidoctor/asciidoctor-stylesheet-factory/pull/42#issuecomment-568645996.

But this does spark the idea of using this technique when it fits. When the frame and grid values are synchronized (all/all, ends/rows, sides/cols), then border collapsing will produce the same result, with the added benefit that it will properly support rowspans and colspans. And that should address the majority of cases since the other permutations are rather esoteric anyway.

I'll submit a proposal.

mojavelinux:
I think you're right though, we never need `border-collapse: separate` since the borders are only being applied selectively. So that can only help.

mojavelinux:
It looks like Firefox and Chrome behave differently. The border is still missing in Firefox with this change, but the appearance is still an improvement since you don't also get the double border.

mojavelinux:
With my proposal combined with yours, we can definitely improve the situation, though it can never be perfect. CSS just doesn't have enough awareness of the table layout to get it exactly right for all combinations.

ahus1:
@mojavelinux - sounds good to me. Will you take the CSS you prepared and close this PR?

mojavelinux:
Yes.

mojavelinux:
I've performed visible inspection of all permutations I can think of. I can confirm that this change does not affect the borders on any tables without a rowspan or colspan. What it does do is prevent the border from doubling at the boundary of a rowpsan or colspan where the frame and grid meet. So we're okay to proceed.

mojavelinux:
It's not correct to say that this solves #2563. It takes steps towards solving it. I'm going to open a separate issue to make it clear this is incremental.

mojavelinux:
Btw, the following CSS would fix a couple more cases in Firefox when the frame and grid are compatible:

```
table.frame-all.grid-all,table.frame-ends.grid-rows,table.frame-sides.grid-cols{border:0}
table.frame-all.grid-all>*>tr:nth-child(n)>.tableblock{border-width:1px}
table.frame-ends.grid-rows>*>tr:nth-child(n)>.tableblock{border-width:1px 0}
table.frame-sides.grid-cols>*>tr:nth-child(n)>.tableblock{border-width:0 1px}
```

This covers:

* frame=all,grid=all (the default)
* frame=ends,grid=rows
* frame=sides,grid=cols

PR #3532, opened 2020-01-08
resolves #1687 multi-row table headings with hrows attribute

### LINKED ISSUE #1687: Multi-row table headings
A patch has been proposed to Asciidoc Python that changes the behaviour of table headers.  It is awaiting documentation, but it appears that it will mean that if `options=header` and any cell in the first row has a vertical span all rows in that span become headers.

Since vertical spans in header rows currently do not produce very sensible output on either Asciidoc Python or Asciidoctor it probably isn't used much, so the change isn't likely to break anything AFAICT.

What does the Asciidoctor community think of the suggestion?

Mogztter:
- `hrows` attribute can be used to define the number of rows in the header (it can also be defined programmatically with `Asciidoctor::Table.header_rowcount` attribute)
- `header` option is now optional if `hrows` attribute has a value greater than one
- `hrows=0` has an higher precedence over  the `header` option (since the attribute is more specific). Meaning that a table with the following definition `[%header,hrows=0]` will have no header.

I think we should implement the same feature for footer using `frows` attribute but I don't want to introduce two features at once in this pull request.

### DIFF
--- a/lib/asciidoctor/table.rb
+++ b/lib/asciidoctor/table.rb
@@ -50,6 +50,9 @@ class Table < AbstractBlock
   # Public: Boolean specifies whether this table has a header row
   attr_accessor :has_header_option
 
+  # Public: An Integer of the number of rows in the header (default: 0)
+  attr_accessor :header_rowcount
+
   # Public: Get the caption for this table
   attr_reader :caption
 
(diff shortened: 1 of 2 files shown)


bwklein:
Would the 2 header rows be repeated if the table breaks across pages in asciidoctor-pdf?
This happens now for single-row headers.

Mogztter:
To be honest, I didn't check.
Since it's possible to create a multi-row header table, I think that asciidoctor-pdf should handle this case. @bwklein Could you please open an issue over there?

PR #3561, opened 2020-02-04
manpage: use monospaced font for literal and listing blocks

kolyshkin:
Most of man page viewing happens on the terminal, which
is monospaced anyway, but one can generate a PDF or PostScript
of a man page as well.

For example:

```console
$ man -Tps man > man.ps
```

or
```console
$ groff -Tps -t -man page.1 > page.ps
$ ps2pdf page.ps page.pdf # optional
```

In this case, if a man page was generated by asciidoctor, all the
blocks that are supposed to be using monospaced font, won't.

Fix is easy.

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -247,7 +247,9 @@ r lw(\n(.lu*75u/100u).'
     result << %(.sp
 .if n .RS 4
 .nf
+.fam C
 #{manify node.content, whitespace: :preserve}
+.fam
 .fi
 .if n .RE)
     result.join LF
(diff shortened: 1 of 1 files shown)


kolyshkin:
I used the following small snipped to test:
```adoc
= foo(1)
:doctype: manpage

== NAME

foo - make bars

== EXAMPLES

 foo -n 5

will generate five bars.

----
foo -o - | gzip -c > foos.gs
----
will compress the resulting bar.

....
foo -P4
....
will use 4 CPUs to produce bar.
```

Using these commands:
```console
$ asciidoctor -b manpage -d manpage foo.adoc
$ man ./foo.1 # to view in terminal
$ man -Tps ./foo.1  > foo.ps # convert man to PS
$ evince foo.ps # use your fav ps viewer
# or conver to PDF
$ ps2pdf foo.ps
$ xpdf foo.pdf # use your fav pdf viewer
```

kolyshkin:
@mojavelinux PTAL

mojavelinux:
Although not frequently used, a quick grep / zgrep of the man pages on my Linux system reveals that this is a common solution to this issue. Since it doesn't affect the rendering of the man page in the terminal, I'll accept into this patch release.

PR #3568, opened 2020-02-11
resolves #3569 add sourcemap flag to the CLI

### LINKED ISSUE #3569: Allow sourcemap option to be switched via CLI
Allow the sourcemap option to be switched on using the CLI, specifically a CLI flag. Although the sourcemap option is not needed by the CLI, it may be used by extensions.

We will need to decide on both a short and long name that does not conflict with any existing flags.

hackingotter:
Fixes #3569

When writing an extension to asciidoctor the source location can be useful information. However it is not populated when the sourcmap option is not enabled which currently can only be done from the ruby API but not the CLI.

This pull request fixes that. The flag is -m because -s and -S are both taken already.

### DIFF
--- a/lib/asciidoctor/cli/options.rb
+++ b/lib/asciidoctor/cli/options.rb
@@ -71,6 +71,9 @@ module Asciidoctor
             'disables potentially dangerous macros in source files, such as include::[]' do |name|
             self[:safe] = SafeMode.value_for_name name
           end
+          opts.on '--sourcemap', 'add source location information to each parsed block (default: false)' do
+            self[:sourcemap] = true
+          end
           opts.on '-s', '--no-header-footer', 'suppress enclosing document structure and output an embedded document (default: false)' do
             self[:standalone] = false
           end
--- a/man/asciidoctor.adoc
+++ b/man/asciidoctor.adoc
@@ -113,6 +113,10 @@ This option may be specified more than once.
   Output an embeddable document, which excludes the header, the footer, and everything outside the body of the document.
   This option is useful for producing documents that can be inserted into an external template.
 
+*--sourcemap*::
+    Enable the sourcemap option on the parser, which instructs the parser to add source location information to each parsed block.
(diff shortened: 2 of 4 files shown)


mojavelinux:
Every PR needs to be associated with an issue. I have created an issue for this feature request. https://github.com/asciidoctor/asciidoctor/issues/3569 Please update this PR to reference that issue.

hackingotter:
Thanks, I didn’t know that. I’ve added it to the description but the title seems immutable

mojavelinux:
Thanks!

lib/asciidoctor/cli/options.rb
```diff
@@ -137,6 +138,9 @@ def parse!(args)
           opts.on('-t', '--timings', 'print timings report (default: false)') do |timing|
             self[:timings] = true
           end
+          opts.on('--sourcemap', 'generate source_location data in the AST. Usefull for extensions (default: false)') do |sourcemap|
```
Mogztter: I think it's "Useful" with one "l":
```suggestion
          opts.on('--sourcemap', 'generate source_location data in the AST. Useful for extensions (default: false)') do |sourcemap|
```

Mogztter: The description might be too low level. For reference, here's the description from the User Manual:

> Keeps track of the file and line number for each parsed block. (Useful for tooling applications where the association between the converted output and the source file is important).

I think we should first describe what it does and then where we can retrieve the data?

hackingotter: I aggree. I changed the description --help returns to `keep track of the file and line number for each parsed block (default:false)` and made the full description in the asciidoctor documentation the same as in the User Manual.

mojavelinux:
I did some cleanup and squashed the commits. Aside from that, this was good to go. It was just waiting for post 2.0.x development, which is now underway. Thanks!

PR #3584, opened 2020-02-24
resolves #2586 document attributes that are unset in the main document should remain unset in the AsciiDoc table cell document

### LINKED ISSUE #2586: Unset attributes don't inherit to AsciiDoc table cell
Attributes which are unset in the main document do not remain unset inside an AsciiDoc table cell.

Example:

```
:example-caption!:

|===
a|
.Title
====
example in table
====
|===
```

The caption on the example should not include the label and number (e.g., Example 1), but it does.

mojavelinux:


mojavelinux:
I'm still thinking this one over. At the very least, the introduction of the DEFAULT_ATTRIBUTES map should be pulled out into a separate PR.

PR #3589, opened 2020-03-06
Don't apply passthrough substitution inside stem macros

### LINKED ISSUE #3409: Conflict between stem and inline passtrough when using + in multiple formulas
There's a conflict between stem and inline passtrough substitution when using at least two formulas with a plus in them.

Minimal example :
`stem:[+] X stem:[+]`
which gets substituted to `stem:[<PASSTROUGH containing ] X stem:[>]`

jirutka:
I just moved processing of the stem substitutions before processing of the passthrough substitutions. It seems that it works well and the side effects of this change are acceptable, i.e. they will probably affect much fewer users than the current undesirable behaviour of `+` signs inside the stem macros.

Fixes #3409

### DIFF
--- a/test/substitutions_test.rb
+++ b/test/substitutions_test.rb
@@ -1955,15 +1955,7 @@ context 'Substitutions' do
       end
 
       test 'should passthrough math macro inside another passthrough' do
-        input = 'the text `asciimath:[x = y]` should be passed through as +literal+ text'
-        para = block_from_string input, attributes: { 'compat-mode' => '' }
-        assert_equal 'the text <code>asciimath:[x = y]</code> should be passed through as <code>literal</code> text', para.content
-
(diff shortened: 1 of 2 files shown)


test/substitutions_test.rb
```diff
@@ -1955,15 +1955,7 @@
       end
 
       test 'should passthrough math macro inside another passthrough' do
-        input = 'the text `asciimath:[x = y]` should be passed through as +literal+ text'
-        para = block_from_string input, attributes: { 'compat-mode' => '' }
-        assert_equal 'the text <code>asciimath:[x = y]</code> should be passed through as <code>literal</code> text', para.content
-
-        input = 'the text [x-]`asciimath:[x = y]` should be passed through as `literal` text'
```
jirutka: What is this obscure syntax anyway? o.O

mojavelinux:
The passthrough syntax has the higher precedence, so this would not be the correct change.

PR #3591, opened 2020-03-07
sections? returns false for anything that isn't itself a section

Mogztter:
resolves https://github.com/asciidoctor/asciidoctor.js/issues/850

Another solution would be to assign a default value (0) in the constructor.
Also, I believe that the internal method `assign_numeral` will fail if `@next_section_index` is `nil`. But since it's an internal method it should only be called on `Section`.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -27,6 +27,7 @@ Bug Fixes::
 
   * Remove unnamespaced selectors in Pygments stylesheet
   * Normalize output from Pygments to use `linenos` class for inline line numbering and trim space after number; update default stylesheet accordingly
+  * Change `AbstractBlock#sections?` to return false when called on block that isn't a Section or Document
 
 Improvements::
 
--- a/lib/asciidoctor/abstract_block.rb
+++ b/lib/asciidoctor/abstract_block.rb
@@ -130,11 +130,13 @@ class AbstractBlock < AbstractNode
 
   # Public: Check whether this block has any child Section objects.
   #
-  # Only applies to Document and Section instances
+  # Acts an an abstract method that always returns false unless this block is an
+  # instance of Document or Section.
+  # Both Document and Section provide overrides for this method.
   #
-  # Returns A [Boolean] to indicate whether this block has child Section objects
+  # Returns false
   def sections?
-    @next_section_index > 0
+    false
   end
 
   # Deprecated: Legacy property to get the String or Integer numeral of this section.
--- a/lib/asciidoctor/document.rb
+++ b/lib/asciidoctor/document.rb
@@ -633,6 +633,13 @@ class Document < AbstractBlock
     end
   end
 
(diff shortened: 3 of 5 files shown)


mojavelinux:
Thanks! Though I'd rather this be written as:

```
(@next_section_index || 0) > 0
```

Another possible approach is to implement this method in AbstractBlock with a fixed return value of false, then override it in Section and Document.

Yet another idea is to have AbstractSection that both Section and Document import. Though that seems a bit drastic for this particular change.

Mogztter:
> Thanks! Though I'd rather this be written as:
(@next_section_index || 0) > 0

I'm fine with that.

> Another possible approach is to implement this method in AbstractBlock with a fixed return value of false, then override it in Section and Document.

I like that, but in this case, we should also initialize `@next_section_index` in the `Document` and `Section` constructor right? Instead of doing that:

```rb
if context == :document || context == :section
    @level = @next_section_index = 0
    @next_section_ordinal = 1
#...
```

What about `reindex_sections` and `assign_numeral`? I feel like they don't belong to AbstractBlock... maybe we could move these two methods on a module and use `include` to include them into `Document` and `Section`?

> Yet another idea is to have AbstractSection that both Section and Document import. Though that seems a bit drastic for this particular change.

If we move `reindex_sections` and `assign_numeral` then it can make sense... You want to add this class in the inheritance hierarchy? 

```rb
class Document < AbstractSection
  # ...
end

class AbstractSection < AbstractBlock
  # ...
end
```

> Another solution would be to assign a default value (0) in the constructor.

Just out of curiosity, what is your thoughts about this solution? Arguably, we assign a default value that don't really make sense but at the same time we avoid the extra check.
I tend to avoid using `nil`, `null` value when possible and use a default value: 0, empty list...

Mogztter:
@mojavelinux How do you want to proceed? I can update the code to use `(@next_section_index || 0) > 0` and we can discuss the other changes at a later time?

mojavelinux:
> What about reindex_sections and assign_numeral? I feel like they don't belong to AbstractBlock... maybe we could move these two methods on a module and use include to include them into Document and Section?

I think I tried to move them at one point, but something didn't work out at the time, so I reverted the change.

I do think that sections? on AbstractBlock should return false and the implementation should be on Document and Section. This causes some duplication, but it's more technically correct. And that may be why I didn't move the other methods. Because these are split hierarchies. But we can just redefine the methods to do the same thing in both Document and Section.

Mogztter:
> I do think that sections? on AbstractBlock should return false and the implementation should be on Document and Section.

Ok, should I update the code accordingly?

mojavelinux:
I updated the code to implement the idea I proposed. As a follow-up, I think we should make a "Sectionlike" mixin module that is added to Section and Document that contains the shared methods that are not part of AbstractBlock. But we can solve this problem without that refactoring.

Mogztter:
Thanks 👍🏻

PR #3609, opened 2020-03-28
resolves #3607 convert video to <videoobject> in DocBook

### LINKED ISSUE #3607: Convert video node to videoobject in the built-in DocBook converter
DocBook has a [videoobject]( https://tdg.docbook.org/tdg/4.5/videoobject.html) element that can be used to describe a video and its associated meta-information.

The following AsciiDoc content:

```adoc
video::e9aoQ9xOmoU[youtube,width=560,height=315]
```

could generate the following DocBook:

```xml
<videoobject>
<videodata fileref="https://www.youtube.com/embed/e9aoQ9xOmoU" width="560"/>
</videoobject>
```

(Please note that according to https://tdg.docbook.org/tdg/4.5/videodata.html, we can only specify a width)

Mogztter:
resolves #3607


mojavelinux:
I'm okay with adding the mapping to the videoobject tag in DocBook, but I don't want to introduce the media abstraction in the same PR. In fact, that's really something that should be discussed as part of the new HTML5 converter (and better conversion helpers in general).

Does DocBook even support Vimeo and YouTube videos?

Mogztter:
> Does DocBook even support Vimeo and YouTube videos?

The goal is to avoid duplicating the logic to produce the video URI.
You can set a `fileref` (URI) on a `videodata` element. I think it's important to use the complete URI produced by the built-in HTML 5 converter in case DocBook is just an intermediate step (in other words, we don't want to lose information).

For instance, if I do the following: AsciiDoc -> DocBook 5 -> HTML 5, ideally I should be able to recreate the `iframe` in the HTML 5 output from the DocBook.

Does it make sense?

mojavelinux:
Ah. That makes sense. So the answer is yes, there is a path.

Where I think this helper should go is into AbstractNode alongside icon_uri. It seems like a lot of logic to put into AbstractNode, but it is just creating a URI (technically a URL) after all, and therefore is equivalent to what we have there. (And other converters like EPUB3 and reveal.js can use it).

However, the helper method should not escape `&` as that assumes the output is HTML or XML. That converter will need to do that part.

Mogztter:
> Ah. That makes sense. So the answer is yes, there is a path.

:+1: 

> Where I think this helper should go is into AbstractNode alongside icon_uri. It seems like a lot of logic to put into AbstractNode, but it is just creating a URI (technically a URL) after all, and therefore is equivalent to what we have there. (And other converters like EPUB3 and reveal.js can use it).

Ok, I put the method into `AbstractNode`.

> However, the helper method should not escape & as that assumes the output is HTML or XML. That converter will need to do that part.

That's a good point!
I'm not sure if a simple `gsub('&', '&amp;')` would be enough? Somehow, I was afraid to replace `&` where it wasn't needed so I've added an `entity_encode` argument to control if we should encode `&` or not... but that's not great :disappointed:

Also I've added `target` as an argument (to make the API consistent with `media_uri` and `image_uri`) but I'm not convinced...

```rb
# node.video_uri
node.video_uri(node.attr('target'))
```

mojavelinux:
One other thing to mention is that I'm not thrilled about AbstractBlock knowing about YouTube and Vimeo. I had that logic in the converter since it's really up to the converter to decide what it supports.

What we may want to do, which keeps coming up again and again, is to have an adapter/provider like with the syntax highlighter (and eventually stem). We need a way to allow integrations to be plugged in without the core objects knowing about them...and to allow others to be registered.

Thus, on deeper thought, I think we need yet another kind of adapter here rather than pulling all this service-specific code into a core object.

Issue #503, opened 2013-07-14
Add attributes for reusable, adaptive, and social content


graphitefriction:
Attributes:
1. Primary Title/Headline/Full Title/Colloquial Title 
   - Size: 100 characters
   - Uses: when article is displayed on its own page
   - matches OpenGraph/Twitter Card meta tags
2. Subtitle
   - optional
   - 1 sentence description of content
3. Short Title
   - Size: 30-50 characters
   - Uses: email subject lines, URL slug generation, display in sidebars, tweets
4. Description/summary
   - Size: 155 characters
   - Uses: If size limit is adhered to, matches summary metadata fields for search engines and social networks
5. miniTeaser
   - Size: 100-120 characters
   - content type in NPR
6. Teaser 
   - Size: 200-400 characters 
   - can be written/overridden individually or auto-generated from first paragraph
7. Excerpt
   - quote from article
   - Uses: as a pull quote on the article page and/or to supplement the summary or teaser

Example Combos:

Tweet: Short title + URL
Twitter Card/Facebook Open Graph: Short title + URL + Primary title + miniTeaser

This is not my work! The data, names, and combos are all thanks to the great information provided by Jeff Eaton in [Big Trouble in Little Content: Planning for Reusable Microcopy](http://www.lullabot.com/blog/articles/big-trouble-little-content-planning-reusable-microcopy).

Character limits for services:

App.net post: 256 chars
Twitter card summary text: 200 chars
Facebook og:description text: 160 chars
Google page description: 155 chars
Tweet: 140 chars
Tweet with link: 116 chars
Subject line in iOS Mail.app: 45 chars

Issue #199, opened 2013-03-02
Add support for twitter cards


mojavelinux:
Twitter supports custom metadata headers in an HTML file that will be displayed below a tweet that links to the URL that has them. Add support in Asciidoctor to transfer custom attributes to the HTML header of the generated document.

https://dev.twitter.com/docs/cards

PR #3618, opened 2020-04-12
Add Open Graph and Twitter card support to html5 backend

nuket:
Not sure if there was further thinking on how to implement issues #199 and #503, but it is something I need in a set of pages I'm working on. So I added this to the html5 backend. I can add documentation and tests too, but before investing time there I'd like to know if this is roughly a reasonable approach.

This might impact Jekyll users of the jekyll-seo-tag plugin (I'm actually one of them), if they declare these attributes in the document header, but usually the SEO config there is part of the `_config.yml` file. In that case they might get double the Open Graph and Twitter tags, but that might not actually be a problem.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -109,6 +109,22 @@ class Converter::Html5Converter < Converter::Base
     result << %(<meta name="keywords" content="#{node.attr 'keywords'}"#{slash}>) if node.attr? 'keywords'
     result << %(<meta name="author" content="#{((authors = node.sub_replacements node.attr 'authors').include? '<') ? (authors.gsub XmlSanitizeRx, '') : authors}"#{slash}>) if node.attr? 'authors'
     result << %(<meta name="copyright" content="#{node.attr 'copyright'}"#{slash}>) if node.attr? 'copyright'
+    # Open Graph tags - Basic and Optional Metadata - https://ogp.me/
+    result << %(<meta property="og:title" content="#{node.attr 'social-title'}"#{slash}>) if node.attr? 'social-title'
+    result << %(<meta property="og:type" content="#{node.attr 'social-type'}"#{slash}>) if node.attr? 'social-type'
+    result << %(<meta property="og:url" content="#{node.attr 'social-url'}"#{slash}>) if node.attr? 'social-url'
+    result << %(<meta property="og:image" content="#{node.attr 'social-image'}"#{slash}>) if node.attr? 'social-image'
+    result << %(<meta property="og:image:width" content="#{node.attr 'social-image-width'}"#{slash}>) if node.attr? 'social-image-width'
+    result << %(<meta property="og:image:height" content="#{node.attr 'social-image-height'}"#{slash}>) if node.attr? 'social-image-height'
+    result << %(<meta property="og:site_name" content="#{node.attr 'social-site-name'}"#{slash}>) if node.attr? 'social-site-name'
+    result << %(<meta property="og:description" content="#{node.attr 'social-description'}"#{slash}>) if node.attr? 'social-description'
+    result << %(<meta property="og:locale" content="#{node.attr 'social-locale'}"#{slash}>) if node.attr? 'social-locale'
+    # Twitter tags - https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary
+    result << %(<meta name="twitter:card" content="#{node.attr 'social-twitter-card'}"#{slash}>) if node.attr? 'social-twitter-card'
+    result << %(<meta name="twitter:site" content="#{node.attr 'social-twitter-site'}"#{slash}>) if node.attr? 'social-twitter-site'
+    result << %(<meta name="twitter:title" content="#{node.attr 'social-title'}"#{slash}>) if node.attr? 'social-title'
+    result << %(<meta name="twitter:description" content="#{node.attr 'social-description'}"#{slash}>) if node.attr? 'social-description'
+    result << %(<meta name="twitter:image" content="#{node.attr 'social-image'}"#{slash}>) if node.attr? 'social-image'
     if node.attr? 'favicon'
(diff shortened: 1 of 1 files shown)


nuket:
I've put these document header attributes to work for me [here](https://vilimpoc.org/travel/tunisia-2019/), [here](https://fromplantoprototype.com/r/noisefloor/), and [here](https://vilimpoc.org/research/GreenOnionBucket/). "View source" to see the output.

The tags work correctly with the [Twitter Card Validator](https://cards-dev.twitter.com/validator) and [Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/).

iranzo:
@nuket can you please also add ld+json ?  (Check my #3883) it should be mostly reuse what you've for twitter and opg.

mojavelinux:
I'm concerned how much extra metadata this is going to add to the HTML output by default (which affects tens of thousands of users). We need to tame this by putting it behind a feature flag like `seo` or `seo-tags` or something. It also needs to be moved to a method. Though I wonder if it would be even better to move it all to an extension and make some sort of extension point. Feel free to share some ideas so we can get a clearer picture of what action to take here.

mojavelinux:
...or maybe we could reuse the existing docinfo extension point and make this a shared library.

iranzo:
Having it out of the box, will increase the 'seo' for HTML generated docs (html is for the web, isn't it?), but having to enable an extra feature is not like a big deal if that will accomodate the users already doing this, and still empower other uses that will get extra benefit from search engines 'understanding' better what the document is about.

If the problem is the extra 'code', an HTML compressor can be used to minify it (and probably should be) but for the whole document, not just headers.

I cannot commend about the way to do this as I'm not that familiar with the asciidoctor code

autonomousapps:
This PR is almost two years old. Is there still interest in adding support for twitter, etc, to the html5 backend?

mojavelinux:
I think this is best done using an extension. I've expressed this concern several times already, so my position is very clear. This is too specific to current social media services for it to be general purpose enough to include in the core converter. There are a lot of attributes being introduced here without any sort of standarization process. I'd rather those attribute names be first defined in the language specification if they are going to be a core part of AsciiDoc.

autonomousapps:
Thanks for the quick response. 

I'm pretty new to the asciidoctor ecosystem, experimenting with it in earnest right now. When I search for "asiidoc Twitter" I find this PR and the linked issue. From that very basic perspective, your position is only clear insofar as both the issue and PR are long neglected. :)

If I may, I'd suggest closing both?

mojavelinux:
I don't have time to do that right now. Given this is a open source project, I'm under no obligation to provide any additional information. When I have the time, I will do so. If someone wants to take this initiative further, and post some information about how to add SEO metadata to the HTML output, they are also free to do so.

gnusupport (changes requested):
I suggest adding schema.org as well.

```
    <!-- Hyperscope Meta Tags -->
    <!-- meta http-equiv=\"content-type\" content=\"text/html; charset=utf-8\"/ -->
    <meta name=\"generator\" content=\"%s\"/>
    <meta name=\"author\" content=\"%s\"/>
    <meta name=\"keywords\" content=\"%s\"/>
    <meta name=\"description\" content=\"%s\"/>
    <!-- Standard Meta Tags -->
    
    <!-- Schema.org -->
    <meta itemprop=\"name\" content=\"%s\" /> 
    <meta itemprop=\"description\" content=\"%s\" />  
    <meta itemprop=\"image\" content=\"%s\" /> 
    <!-- Schema.org -->
    
    <!-- Open Graph -->
    <meta property=\"og:title\" content=\"%s\"/>
    <meta property=\"og:type\" content=\"%s\"/>
    <meta property=\"og:url\" content=\"%s\"/>
    <meta property=\"og:image\" content=\"%s\"/>
    <meta property=\"og:site_name\" content=\"%s\"/>
    <meta property=\"og:description\" content=\"%s\"/>
    <!-- Open Graph -->
` ` `

gnusupport:
> I don't have time to do that right now. Given this is a open source project, I'm under no obligation to provide any additional information. When I have the time, I will do so. If someone wants to take this initiative further, and post some information about how to add SEO metadata to the HTML output, they are also free to do so.

These tags are not necessarily "SEO" related, they are related to marketing and presentation of pages.

When page has OpenGraph or Schema.org or Twitter tags, such page knows which image to present, and by simple including of a link, system fetch image, title, description and present nicer those pages. It is crucial today for many systems, social networks, chat systems and similar.

In other words, visitors learn about the page before visiting the page and it helps to promotion.

gnusupport:
> I think this is best done using an extension. I've expressed this concern several times already, so my position is very clear. This is too specific to current social media services for it to be general purpose enough to include in the core converter. There are a lot of attributes being introduced here without any sort of standarization process. I'd rather those attribute names be first defined in the language specification if they are going to be a core part of AsciiDoc.

Actually not, it is not so specific to current social media. Twitter tags are used in many other pages and networks, it is not for Twitter only. Schema.org and Opengraph are now standards. It is used on too many information channels of various types.

I would actually remove Twitter because it's name is too specific, and keep Schema.org and Opengraph.

gnusupport:
> I'm concerned how much extra metadata this is going to add to the HTML output by default (which affects tens of thousands of users). We need to tame this by putting it behind a feature flag like `seo` or `seo-tags` or something. It also needs to be moved to a method. Though I wonder if it would be even better to move it all to an extension and make some sort of extension point. Feel free to share some ideas so we can get a clearer picture of what action to take here.

It is not "SEO" related so much. It is not good using that term `seo`.

Rather something like:

:schema-org-name: TITLE
:schema-org-description: DESCRIPTION
:schema-org-image: IMAGE

:og-title: TITLE
:og-type: Article
:og-url: LINK
:og-image: IMAGE
:og-site-name: Site Name
:og-description: DESCRIPTION

By using those tags users could decide which one to include.

Or give us just general method to include `<meta ` tags, then I can solve it.

PR #3640, opened 2020-05-04
Emit only XML-relevant declarations in XHTML mode.

marioxcc:


mojavelinux:
Do you have a reference for this? My understanding is that it's valid to use both the DOCTYPE and meta encoding tag in the XML serialization mode of HTML5. Here are two references that support this assertion:

* https://mathiasbynens.be/notes/xhtml5
* https://blog.whatwg.org/xhtml5-in-a-nutshell

marioxcc:
Both are allowed but they serve no purpose.

From [HTML vs. XHTML](https://wiki.whatwg.org/wiki/HTML_vs._XHTML) in the WHATWG wiki:

> `meta` is only allowed as case-insensitive "UTF-8" (and **is ignored** if included).

> The DOCTYPE is **optional.** XML rules for case sensitivity apply (everything is case sensitive). 

From the [WHATWG FAQ](https://github.com/whatwg/html/blob/master/FAQ.md):

> In documents delivered with an XML media type, XML rules for determining the character encoding declaration apply. The <meta> element is never used for determining the encoding of such documents; for those, you should use either the HTTP Content-Type header or the XML declaration to specify the encoding.

From the HTML5 standard, [§ 13.1 “Writing documents in the XML syntax”](https://html.spec.whatwg.org/multipage/xhtml.html#writing-xhtml-documents):

> XML documents may contain a DOCTYPE if desired, but this is not required to conform to this specification. This specification does not define a public or system identifier, nor provide a formal DTD.

DOCTYPE declarations are meaningless in XHTML. They are a part of the obsolete DTD language for XML schemes. HTML 5 has no DTD scheme. DOCTYPE does not give the document type. The document type is given by out-of-band information (i.e.: `Content-Type` HTTP header) or by the name and namespace of the root element.

gdamore:
So I recall that there is this thing called polyglot markup: https://www.w3.org/TR/html-polyglot/

I really believe that it's better to have those elements (even ignored ones) in the markup to allow the documents to be parsed by the widest possible set of user agents.

marioxcc:
> to allow the documents to be parsed by the widest possible set of user agents.

This is incorrect. User agents do not have a choice on whether they interpret the document as XML syntax or HTML syntax. The choice is made server-side and given to the client by the Content-Type HTTP header. https://html.spec.whatwg.org/multipage/introduction.html#html-vs-xhtml says:

> If a document is transmitted with the text/html MIME type, then it will be processed as an HTML document by web browsers. This specification defines the latest HTML syntax, known simply as "HTML".
>
> The second concrete syntax is XML. When a document is transmitted with an XML MIME type, such as application/xhtml+xml, then it is treated as an XML document by web browsers, to be parsed by an XML processor.

Polyglot HTML is a hack to work around incompetent system administrators that incorrectly serve XHTML with the `text/html` media type. Note that in our relevant case the user unambiguously requests AsciiDoctor to use the XML serialization with `-b xhtml5`. At no point is this hack requested.

mojavelinux:
I'm not comfortable removing the DOCTYPE declaration (which I realize is no longer a true DOCTYPE declaration, I get that, but does tell browsers that the document is HTML 5). I'm fine with adding the XML declaration and removing the meta tag.

mojavelinux:
One of the strong cases against adding the XML declaration is that it causes the document to not validate when using either https://validator.w3.org/ or https://html5.validator.nu/. This situation is extremely frustrating for developers in my position because regardless of which markup we choose, users will file issues to say it's wrong. And one cannot deny that many of the subject matter experts disagree that XHTML 5 even exists (certainly the validators don't acknowledge it).

The point of adding the XML mode was to allow XML parsers to postprocess the output. That goal has been accomplished. Whether that output should be served directly to the browsers, well, that really seems to be an open question since we can't get a clear answer from the standards groups.

I really don't want to welcome the debate about whether XHTML 5 exists into this project as it's not our problem to solve. If someone can present a definitive source of rules to follow, I'm happy to adhere to them. What I'm not willing to do is change it at the behest of a single use case. The solution needs to be universal.

At this point, I'm not sure what action to take. I'm awaiting more input.

marioxcc:
> but [the `DOCTYPE` declaration] does tell browsers that the document is HTML 5.

Incorrect. The [relevant section](https://html.spec.whatwg.org/multipage/xhtml.html#the-xhtml-syntax) of the HTML5 standard says no such thing.

> One of the strong cases against adding the XML declaration is that it causes the document to not validate when using either https://validator.w3.org/ or https://html5.validator.nu/.

In that case those validators are broken. XML parsers are [obliged](https://www.w3.org/TR/xml/#sec-prolog-dtd) to accept the XML declaration.

> Whether that output should be served directly to the browsers, well, that really seems to be an open question since we can't get a clear answer from the standards groups.

By exactly the same logic, it is just as much an “open question” whether the HTML serialization should be served to browsers.

> The point of adding the XML mode was to allow XML parsers to postprocess the output.

That implies intention. Whose intention do you mean and why is only that intention relevant?

> I really don't want to welcome the debate about whether XHTML 5 exists into this project

There is no debate. HTML5 is *defined* by a standard that defines a XML serialization. The opinion of “experts”, by whatever arbitrary standard you determine who is an “expert”, is irrelevant.

elextr:
> Incorrect. The relevant section of the HTML5 standard says no such thing.

The page you linked says "XML documents may contain a DOCTYPE if desired, but this is not required to conform to this specification. This specification does not define a public or system identifier, nor provide a formal DTD."

marioxcc:
> > Incorrect. The relevant section of the HTML5 standard says no such thing.
> 
> The page you linked says "XML documents may contain a DOCTYPE if desired, but this is not required to conform to this specification. This specification does not define a public or system identifier, nor provide a formal DTD."

Read what I wrote again. This time try to read *all* of it.

elextr:
The title and the changes of this PR are to not emit the DOCTYPE, but as noted, the standard allows it in the XML.  

So as @mojavelinux says in his first comment, its ok to have it still there.  It doesn't matter if the usefulness is minimal, its still legal.

[This is also being debated on another project I participate in, hence the quick reply, so thanks to @marioxcc for the reference :grin: ]

mojavelinux:
I want to clarify that I'm not trying to win an argument here. Please understand that I just need to work out what change will be acceptable, and not disruptive, to our broad user community. Regardless of what we decide, I have to support it. And I'm still not convinced of what that change is yet.

The fact that the presence of the XML declaration doesn't validate using the W3C validator does not help the case for adding it. The W3C is the authoritative standards organization for web technologies. If that tool cannot recognized and validate an XHTML5 document, that's a real problem that cannot be waved off.

In the change's defense, the MDN docs do show an example of an XHTML document with the XML declaration and without the DOCTYPE declaration: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/XHTML. However, the page then goes on to say there are very few uses of XHTML on the web and links to articles that recommend against using it.

To provide some historical context around the XML mode in Asciidoctor (i.e., backend=xhtml5), it was added to allow the site plugin for Maven (specifically, Doxia) to parse the document. That was the intention of it. To my knowledge, the output the converter currently produces is able to satisfy that requirement.

One thing is for certain is that the guidance around XHTML5 is both limited and ambiguous. And as I stated before, this is not the right forum to resolve that debate. What I'm looking for are clear guidelines to which we need to conform. I'm inclined to follow the W3C validator, which approves the current output of the xhtml5 backend. (Though the example on the MDN docs is nonetheless compelling).

PR #3644, opened 2020-05-09
resolves #3645 escape ellipsis at start of line in manpage output

### LINKED ISSUE #3645: Ellipsis at beginning of line causes line to disappear from manpage output
If an ellipsis falls the the beginning of the line, that line is not visible in the manpage output. The reason is that the ellipsis is not properly escaped, so the man viewer ignores the line.

Example:

```
...and so it begins
```

The line in the manpage output should look like this:

```
\&...and so it begins
```

jnavila:
when an ellipsis is at the start of line, is should be:

 1. converted back to "..."
 2. escaped so that the first period does not trigger a macro

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -34,6 +34,7 @@ Bug Fixes::
   * Removing leading and trailing blank lines in AsciiDoc include file to match assumption of parser (#3470)
   * Activate extensions when :extensions option is set even if Extensions API is not yet loaded (#3570)
   * Don't activate global extensions if :extensions option is false (#3570)
+  * In manpages backend, escape ellipsis at start of line
 
 Compliance::
 
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -704,6 +704,7 @@ allbox tab(:);'
     end
     str = str.
       gsub(LiteralBackslashRx) { %[#{$1}#{'\\(rs' * $2.length}] }. # literal backslash (not a troff escape sequence)
+      gsub(EllipsisCharRefRx, '...'). # horizontal ellipsis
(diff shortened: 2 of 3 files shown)


mojavelinux:
Thanks for raising this issue and for submitting a fix with a test!

I think we can do this without introducing an additional substitution. First, update the EllipsisCharRefRx to check if it starts at the beginning of a line:

```rb
EllipsisCharRefRx = /(^)?&#8230;(?:&#8203;)?/
```

then, do a conditional replacement based on whether it falls at the start of a line:

```rb
gsub(EllipsisCharRefRx) { $1 ? '\\&...' : '...' }. # horizontal ellipsis
```

Could you also make two other changes?

1. Add a test for an ellipsis not at the start of a line (could be the same test)
2. Add an entry to the Bug Fixes category in the CHANGELOG file

jnavila:
> Thanks for raising this issue and for submitting a fix with a test!
> 
> I think we can do this without introducing an additional substitution. 

The patch I wrote does not introduce an additional substitution, but reorder substitutions so that the resolution of ellipsis happens before the escaping of period at start of lines (which will escape the periods of ellispis if the ellipsis happens to be at start of line).
So it's the least disruptive patch, without even testing whether the ellipsis is at the start of line.

> 
> Could you also make two other changes?
> 
> 1. Add a test for an ellipsis not at the start of a line (could be the same test)

Fair enough. 

> 2. Add an entry to the Bug Fixes category in the CHANGELOG file

Will do.

mojavelinux:
You're absolutely correct. I stand corrected. Please disregard that part of the request then.

One other thing. Please remove the added coding line. That's no longer required in Ruby.

jnavila:
Done.

mojavelinux:
Thanks!

PR #3653, opened 2020-05-14
Adds missing Catalan attributes translation + fixes appendix

abelsromero:
* Fixes missing grave accent in `appendix-caption`.
* Adds missing Catalan values.

**Before merging,** I'd like double check `part-refsig`. What's its use?
Depending on it, I think it would be more appropiate to use "apartat", which is specific to designate a document seccion with a function and clearly separated, for examaple, by spacing, title and/or numbering. It also tends to be a short element, like a sub-section of a chapter. "Part" is a very generic term for anything, even without estructure or boundaries.
For exemple, you would say "the pre-installation chapter has 2 'apartats': hardware requirements and software requirements", or "All novels have 3 'parts'; introduction, development and conclusion"...but not the other way arround.

### DIFF
--- a/data/locale/attributes-ca.adoc
+++ b/data/locale/attributes-ca.adoc
@@ -1,9 +1,9 @@
 // Catalan translation, courtesy of Abel Salgado Romero <abelromero@gmail.com> and Alex Soto
-:appendix-caption: Apendix
+:appendix-caption: Apèndix
 :appendix-refsig: {appendix-caption}
 :caution-caption: Atenció
-//:chapter-label: ???
-//:chapter-refsig: {chapter-label}
+:chapter-label: Capítol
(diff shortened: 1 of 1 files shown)


mojavelinux:
In this context, part represents a group of chapters (thus the parent of a chapter). This label is used in the xref text that refers to that part. In English, it would look like "See Part I: Before the War".

abelsalgadoromeroWD:
Given xref can point to several levels, let's go for "part" then.

abelsalgadoromeroWD:
To be clear, can be merge as is it now.

abelsromero:
Given xref can point to several levels, let's go for "part" then. So, no changes required
To be clear, this can be merged as it is now.

mojavelinux:
Thanks @abelsromero!

PR #3661, opened 2020-05-17
resolves #3660 look up asset dir attribute from node with inheritance enabled to resolve asset path prefix

### LINKED ISSUE #3660: Look for imagesdir attribute on node before looking on document
An asset can be:
- icon (inline)
- image (inline and block)
- audio
- video

When resolving an asset path from the target (using the method `AbstractNode#image_uri`, `AbstractNode#media_uri` or `AbstractNode#icon_uri`) we should first look for the asset directory attribute on the node and then look on the document (ie. current behavior).

ggrossetie:
By default, the asset directory is `imagesdir` or `iconsdir` depending on the type of asset.

If the asset directory attribute is defined on the node, then we prefer it. Otherwise, we use the value defined on the `Document`.

I've decided to apply the same rule on all type of assets (video, audio, icon and image). I think it would be inconsistent and confusing to implement this behavior only on images.

resolves #3660

### DIFF
--- a/lib/asciidoctor/abstract_node.rb
+++ b/lib/asciidoctor/abstract_node.rb
@@ -322,14 +322,14 @@ class AbstractNode
   def image_uri target_image, asset_dir_key = 'imagesdir'
     if (doc = @document).safe < SafeMode::SECURE && (doc.attr? 'data-uri')
       if ((Helpers.uriish? target_image) && (target_image = Helpers.encode_spaces_in_uri target_image)) ||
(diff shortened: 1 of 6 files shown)


lib/asciidoctor/abstract_node.rb
```diff
@@ -309,16 +309,17 @@ def icon_uri name
   #
   # Returns A String reference or data URI for the target image
   def image_uri(target_image, asset_dir_key = 'imagesdir')
+    images_base = asset_dir_key ? (attr asset_dir_key, nil, true) : nil
```
Mogztter, line 312: this variable is used in both branches so I've decided to declare it above

PR #3663, opened 2020-05-20
Upgrade TruffleRuby in CI to 20.1

Mogztter:


mojavelinux:
Looks like about a 3 minute improvement in runtime (from 10m30s down to 7m30s). Nice!!

PR #3665, opened 2020-05-21
Use an NBSP between outline number and text

rpavlik:
Makes the TOC more appealing for OpenXR specification, at least. We have sections that start with a long "word".

Before:

Screenshot from 2020-05-21 13-21-46

After:

Screenshot from 2020-05-21 13-21-19

This has bugged me for a while, but hadn't gotten around to poking it until now. Ideally we could do this via stylesheet, which we can override easily, but I'm not sure of a way to do that offhand.

cc @oddhack

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -352,14 +352,14 @@ MathJax.Hub.Register.StartupHook("AsciiMath Jax Ready", function () {
       elsif section.numbered && slevel <= sectnumlevels
         if slevel < 2 && node.document.doctype == 'book'
           if section.sectname == 'chapter'
-            stitle =  %(#{(signifier = node.document.attributes['chapter-signifier']) ? "#{signifier} " : ''}#{section.sectnum} #{section.title})
+            stitle =  %(#{(signifier = node.document.attributes['chapter-signifier']) ? "#{signifier} " : ''}#{section.sectnum}&nbsp;#{section.title})
           elsif section.sectname == 'part'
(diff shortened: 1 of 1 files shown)


mojavelinux:
I don't want to add HTML hacks to the built-in HTML. This is something that needs to be handled by the stylesheet.

rpavlik:
For future reference, my crude attempt to do an equivalent thing in javascript is the following:

```js
const toc = document.querySelectorAll('#toc.toc2 li a');
for (let elt of toc) {
  let words = elt.innerHTML.split(' ');
  /* if this is a numbered section */
  if (words.length > 1 && words[0].endsWith('.')) {
    /* don't split the number from the first word of the title */
    elt.innerHTML = elt.innerHTML.replace(words[0] + ' ', words[0] + '&nbsp;');
  }
}
```

Haven't yet looked to see how hard it would be to add this to the output.

mojavelinux:
It would be better to use CSS than to hack the HTML. CSS has the capability to split words if they don't fit on the line. This is not the proper use of the `&nbsp;` character.

mojavelinux:
I'll poke around a bit and see if I can offer a proper solution.

You certainly shouldn't have to resort to JavaScript. After all, you can modify the converter using the API or templates if you really want to change the HTML. See https://github.com/asciidoctor/asciidoctor.org/issues/80

mojavelinux:
I looked into this and the only way I can see to do it (without using `&nbsp;`) is to wrap the section number and trailing space in a span with the class `sectnum`. Then we can apply the following style to it:

```
#toc .sectnum {
  white-space: nowrap;
}
```

I also think we should add `word-wrap: break-word` to the TOC so that long words are broken instead of causing horizontal scrolling.

mojavelinux:
Since this requires a change to the HTML, it will need to be deferred to at least 2.1.

PR #3700, opened 2020-07-05
resolves #3638 check contents before inlining SVG

### LINKED ISSUE #3638: Failure when inlining 0 byte SVG
Coming from asciidoctor-diagram#280. I noticed that the HTML5 backed fails when it tries to online a 0 bytes SVG.

I believe the tool should instead report a warning that the file size is 0.

_I noticed this issue due to a bogus SVG file created by Asciidoctor Diagram._

***

When running asciidoctor on file with a plantuml diagram I encountered the following issue. (Removing the opts="inline" solves prevent the error, but does not match what I need.)
This diagram worked well until now, and I don't know for sure what happend, especially since I don't know much about ruby.

```
asciidoctor --trace --verbose --base-dir=. --backend=html5 --require asciidoctor-diagram --attribute=icons=font --attribute sectlinks --attribute sectanchors --attribute diagram-svg-type=inline plantuml.smoketest.adoc
```

```adoc
[plantuml,"activity-diagram-example",format="svg",opts="inline"]
.Activity diagram
----
@startuml

start
if (condition A) then (yes)
  :Text 1a;
  :Text 1b/
elseif (condition B) then (yes)
  :Text 2|
  stop
elseif (condition C) then (yes)
  :Text 3<
elseif (condition D) then (yes)
  :Text 4>
else (nothing)
  :Text else}
  note right
    This note is on several
    //lines// and can
    contain <b>HTML</b>
    ====
    * Calling the method ""foo()"" is prohibited
  end note
endif
stop

@enduml
----
```

<details><summary>trace</summary>

```
Traceback (most recent call last):
        28: from /usr/local/Cellar/asciidoctor/2.0.10_1/libexec/bin/asciidoctor:23:in `<main>'
        27: from /usr/local/Cellar/asciidoctor/2.0.10_1/libexec/bin/asciidoctor:23:in `load'
        26: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/bin/asciidoctor:15:in `<top (required)>'
        25: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/cli/invoker.rb:111:in `invoke!'
        24: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/cli/invoker.rb:111:in `each'
        23: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/cli/invoker.rb:128:in `block in invoke!'
        22: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/convert.rb:183:in `convert_file'
        21: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/convert.rb:183:in `open'
        20: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/convert.rb:183:in `block in convert_file'
        19: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/convert.rb:118:in `convert'
        18: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/document.rb:951:in `convert'
        17: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:84:in `convert'
        16: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:227:in `convert_document'
        15: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/document.rb:1020:in `content'
        14: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `content'
        13: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `map'
        12: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `block in content'
        11: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:75:in `convert'
        10: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:54:in `convert'
         9: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:416:in `convert_section'
         8: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `content'
         7: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `map'
         6: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:84:in `block in content'
         5: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/abstract_block.rb:75:in `convert'
         4: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:65:in `convert'
         3: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:624:in `convert_image'
         2: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:1259:in `read_svg_contents'
         1: from /Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:1259:in `each'
/Library/Ruby/Gems/2.6.0/gems/asciidoctor-2.0.10/lib/asciidoctor/converter/html5.rb:1261:in `block in read_svg_contents': undefined method `[]' for nil:NilClass (NoMethodError)
```

</details>

<details><summary> <code>gem query --local</code> </summary>

```
*** LOCAL GEMS ***

asciidoctor (2.0.10)
asciidoctor-diagram (2.0.2)
asciidoctor-html5s (0.5.0, 0.4.1)
bigdecimal (2.0.0, default: 1.4.1)
bundler (2.1.4, default: 1.17.2)
CFPropertyList (3.0.2, 2.3.6)
cmath (default: 1.0.0)
csv (3.1.2, default: 3.0.9)
date (3.0.0, default: 2.0.0)
dbm (1.1.0, default: 1.0.0)
did_you_mean (1.4.0, 1.3.0)
e2mmap (default: 0.1.0)
etc (1.1.0, default: 1.0.1)
fcntl (default: 1.0.0)
fiddle (default: 1.0.0)
fileutils (1.4.1, default: 1.1.0)
forwardable (1.3.1, default: 1.2.0)
hpricot (0.8.6)
io-console (0.5.6, 0.5.5, 0.5.4, default: 0.4.7)
ipaddr (default: 1.2.2)
irb (1.2.3, 1.2.1, default: 1.0.0)
json (2.3.0, default: 2.1.0)
libxml-ruby (3.1.0)
logger (1.4.2, default: 1.3.0)
matrix (default: 0.1.0)
mini_portile2 (2.5.0, 2.4.0)
minitest (5.14.0, 5.11.3)
mustache (1.1.1)
mutex_m (default: 0.1.0)
net-telnet (0.2.0)
nokogiri (1.10.9, 1.10.8, 1.10.7, 1.10.1)
openssl (default: 2.1.2)
ostruct (default: 0.1.0)
power_assert (1.2.0, 1.1.7, 1.1.6, 1.1.5, 1.1.3)
prime (default: 0.1.0)
psych (default: 3.1.0)
rake (13.0.1, 12.3.2)
rdiscount (2.2.0.1)
rdoc (6.2.1, default: 6.1.0)
reline (0.1.4, 0.1.3, 0.1.2)
rexml (3.2.4, 3.2.3, default: 3.1.9)
ronn (0.7.3)
rss (0.2.9, 0.2.8, default: 0.2.7)
scanf (default: 1.0.0)
sdbm (default: 1.0.0)
shell (0.8.0, default: 0.7)
sqlite3 (1.4.2, 1.3.13)
stringio (0.1.0, default: 0.0.2)
strscan (1.0.3, default: 1.0.0)
sync (default: 0.5.0)
test-unit (3.3.5, 3.2.9)
thread_safe (0.3.6)
thwait (default: 0.1.0)
tracer (default: 0.1.0)
webrick (1.6.0, default: 1.4.2)
xmlrpc (0.3.0)
zlib (1.1.0, default: 1.0.0)
```

</details>

***

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.4
BuildVersion:   19E287
```

Mogztter:
- Introduce a new option `warn_on_empty` on the `read_contents` method
   + By default, issue a warning message if the contents of the target is empty: "contents of SVG at: /path/to/empty.svg is empty"
   + Since `read_contents` is called with the `normalize` option in `read_svg_contents` we do not need to check if the contents is blank (i.e. using `strip`) because the contents is already trimmed.
- Make sure that the SVG contents is not empty before doing any processing
- Make sure that the SVG header is valid before updating the `width` and `height` attribute
- Add tests

resolves #3638


mojavelinux:
@Mogztter I've made a few stylistic changes. Let me know if they look good to you and I'll squash and merge.

* I renamed :warn_on_empty to :warn_if_empty
* I made :warn_if_empty false by default
* Moved the tests into block_tests.rb (this puts them with the other SVG tests, though we need to do some reorganizing of these tests in the future)
* Defined contents inside the conditional instead of around it (I'm moving to this style)
* Optimized the start tag match logic (it should only run once if we need it)

lib/asciidoctor/converter/html5.rb
```diff
@@ -1257,13 +1257,17 @@ def convert_inline_quoted node
...
+      return if svg.empty?
       svg = svg.sub SvgPreambleRx, '' unless svg.start_with? '<svg'
-      old_start_tag = new_start_tag = nil
+      old_start_tag = new_start_tag = start_tag_match = nil
       # NOTE width, height and style attributes are removed if either width or height is specified
       ['width', 'height'].each do |dim|
         if node.attr? dim
-          new_start_tag = (old_start_tag = (svg.match SvgStartTagRx)[0]).gsub DimensionAttributeRx, '' unless new_start_tag
+          unless new_start_tag
+            next if (start_tag_match ||= (start_tag_match == false ? false : (start_tag_match = (svg.match SvgStartTagRx) || false))) == false
```
Mogztter, line 1268: It can be simplified no?
The inner ternary condition `start_tag_match == false ? ...` is always false because we already check if `start_tag_match` equals `false` in the outer condition (and do `next` if that is the case).
So in the inner condition `start_tag_match` is always not `false`.

Mogztter: I think the inner condition (i.e. right assignment of `||=`) will be evaluated because `start_tag_match` is false...

Mogztter: But we can probably remove one assignment since we are using `start_tag_match ||=`:

```suggestion
            next if (start_tag_match ||= (start_tag_match == false ? false : ((svg.match SvgStartTagRx) || false))) == false
```

Arguably, we should check if `start_tag_match` is `nil` to avoid the extra `start_tag_match == false ? false : ...`.

Not sure what is the one-line syntax but basically what we are doing here is:

```rb
if start_tag_match.nil?
  start_tag_match = (svg.match SvgStartTagRx) || false
end
next if start_tag_match == false
```

Mogztter: The `||=` operator is nice but since we need to make a distinction between `nil` and `false` values, the code becomes less obvious.

Mogztter:
Thanks!
The changes are fine. I do have one suggestion about the start tag match logic (see above).

> Moved the tests into block_tests.rb (this puts them with the other SVG tests, though we need to do some reorganizing of these tests in the future)

:+1:

lib/asciidoctor/converter/html5.rb
mojavelinux: This code optimizes for the virtuous path that the start tag match is found on the previous run. When it is not set, we have to check if it's false to indicate it failed on a previous attempt. The end result is that it only ever runs ones. It may not be beautiful, but it does work.

mojavelinux: If we were to rewrite it, I would go with something like:

```
next if (start_tag_match ||= (svg.match SvgStartTagRx) || :no_match) == :no_match
```

That way we can leverage the `||=` operator effectively. And it seems clear what is happening.

Mogztter: Are you saying that the following code:

```rb
start_tag_match = ((svg.match SvgStartTagRx) || false) if start_tag_match.nil?
next if start_tag_match == false
```

does more work than:

```rb
next if (start_tag_match ||= (start_tag_match == false ? false : (start_tag_match = (svg.match SvgStartTagRx) || false))) == false
```

As far as I understand Ruby still has to evaluate the value of `start_tag_match` twice right? 
1. `start_tag_match` is trueish -> don't do the assignment (`||=` operation)
2. `start_tag_match` is not equals to `false` -> don't execute `next` (`==` operation)

I guess I could try to a do micro benchmark but the array is only two elements long so the difference will be on the nanosecond range :laughing:

mojavelinux: I rewrote it. See if you like the new way.

Mogztter: That's perfect, the code is concise and the intention is clear :1st_place_medal:

PR #3701, opened 2020-07-05
resolves #3519 compute highlight line ranges on source block relative to value of start attribute

### LINKED ISSUE #3519: highlight lines on source block should be relative to start value
If a start value is set, then the lines to highlight specified by the highlight attribute should be relative to the start value. In other words, if the start value is 5 and highlight=5, then the first line should be highlighted. Otherwise, it gets confusing which lines are being highlighted.

```
[source%linenums,ruby,start=5,highlight=6]
----
get {
  render "Hello, World!"
}
----
```

Note that this scenario is only relevant when linenums are enabled.

Mogztter:
resolves #3519

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -39,6 +39,7 @@ Bug Fixes::
   * Don't register footnote with ID if a footnote is already registered with that ID (#3690)
   * Honor start attribute on ordered list in manpage output (#3714)
   * Warn instead of crashing if SVG to inline is empty (#3638) (*@mogztter*)
+  * Compute highlight line ranges on source block relative to value of start attribute (#3519) (*@mogztter*)
 
 Compliance::
 
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -946,7 +946,7 @@ module Substitutors
     if (linenums_mode = (attr? 'linenums') ? (doc_attrs[%(#{syntax_hl_name}-linenums-mode)] || :table).to_sym : nil)
       start_line_number = 1 if (start_line_number = (attr 'start', 1).to_i) < 1
     end
-    highlight_lines = resolve_lines_to_highlight source, (attr 'highlight') if attr? 'highlight'
+    highlight_lines = resolve_lines_to_highlight source, (attr 'highlight'), start_line_number if attr? 'highlight'
 
     highlighted, source_offset = syntax_hl.highlight self, source, (attr 'language'),
       callouts: callout_marks,
@@ -969,9 +969,10 @@ module Substitutors
   #
   # source - The String source.
   # spec   - The lines specifier (e.g., "1-5, !2, 10" or "1..5;!2;10")
+  # start  - The line number of the first line (optional, default: false)
   #
   # Returns an [Array] of unique, sorted line numbers.
-  def resolve_lines_to_highlight source, spec
+  def resolve_lines_to_highlight source, spec, start = nil
     lines = []
     spec = spec.delete ' ' if spec.include? ' '
     ((spec.include? ',') ? (spec.split ',') : (spec.split ';')).map do |entry|
@@ -992,6 +993,10 @@ module Substitutors
       elsif !lines.include?(line = entry.to_i)
         lines << line
       end
+    end
+    # If the start attribute is defined, then the lines to highlight specified by the provided spec should be relative to the start value.
+    unless (shift = start ? start - 1 : 0) == 0
+      lines = lines.map {|it| it - shift }
     end
     lines.sort
   end
--- a/test/syntax_highlighter_test.rb
+++ b/test/syntax_highlighter_test.rb
@@ -903,6 +903,75 @@ context 'Syntax Highlighter' do
       end
     end
 
+    test 'should line highlight specified lines relative to start value' do
+      input = <<~EOS
+        :source-highlighter: rouge
+
+        [source%linenums,ruby,start=5,highlight=6]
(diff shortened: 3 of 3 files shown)


lib/asciidoctor/substitutors.rb
```diff
@@ -971,11 +971,14 @@ def highlight_source source, process_callouts
   #
   # source - The String source.
   # spec   - The lines specifier (e.g., "1-5, !2, 10" or "1..5;!2;10")
+  # start  - The start attribute
   #
   # Returns an [Array] of unique, sorted line numbers.
-  def resolve_lines_to_highlight source, spec
+  def resolve_lines_to_highlight source, spec, start
```
mojavelinux, line 977: This method is used by Asciidoctor PDF, so this argument needs to be optional.

lib/asciidoctor/substitutors.rb
```diff
@@ -948,7 +948,7 @@ def highlight_source source, process_callouts
     if (linenums_mode = (attr? 'linenums') ? (doc_attrs[%(#{syntax_hl_name}-linenums-mode)] || :table).to_sym : nil)
       start_line_number = 1 if (start_line_number = (attr 'start', 1).to_i) < 1
     end
-    highlight_lines = resolve_lines_to_highlight source, (attr 'highlight') if attr? 'highlight'
+    highlight_lines = resolve_lines_to_highlight source, (attr 'highlight'), start_line_number if attr? 'highlight'
```
mojavelinux, line 951: I'm trying to figure out if it makes more sense to pass in the start value, or to remove it from each value in the result. I suppose it's elegant to pass it in.

lib/asciidoctor/substitutors.rb
```diff
@@ -971,11 +971,14 @@ def highlight_source source, process_callouts
...
   # spec   - The lines specifier (e.g., "1-5, !2, 10" or "1..5;!2;10")
+  # start  - The start attribute
   #
   # Returns an [Array] of unique, sorted line numbers.
-  def resolve_lines_to_highlight source, spec
+  def resolve_lines_to_highlight source, spec, start
     lines = []
     spec = spec.delete ' ' if spec.include? ' '
+    # If a start value is set, then the lines to highlight specified by the provided spec should be relative to the start value.
+    offset = start ? start - 1 : 0
```
mojavelinux, line 981: Let's keep the value nil if the start is 1. That why we avoid an unnecessary calculation per entry.

mojavelinux: I think it makes more sense to remove the offset from each entry after the lines have been resolved. That reduces it to a single place in the code.

```
start == 1 ? lines.sort : (lines.sort.map {|it| it - start + 1 })
```

lib/asciidoctor/substitutors.rb
mojavelinux: `start = 1`

Mogztter: > This method is used by Asciidoctor PDF, so this argument needs to be optional.

The value is not optional by default? In other words, the value will be `nil` if not defined no?

> start = 1

Ok :+1:

mojavelinux: Without the default value, it is required (unlike JavaScript) and thus will throw a runtime error.

lib/asciidoctor/substitutors.rb
Mogztter: `start_line_number` can be `nil`, should we handle this case in `resolve_lines_to_highlight`?

```rb
start && start > 1 ? (lines.sort.map {|it| it - start + 1 }) : lines.sort
```

As an alternative, we could state in the documentation that `start` must be a positive integer greater or equals to 1?

lib/asciidoctor/substitutors.rb
Mogztter: That's a good idea.
I wonder if we should check that we won't produce negative numbers if start is greater than highlight?

mojavelinux: I don't think we should worry about negative numbers. They will effectively be ignored anyway by the syntax highlighter.

Mogztter: Indeed, I've added a test to make sure that the code is robust.

lib/asciidoctor/substitutors.rb
mojavelinux: Good point. In that case, the method argument should be `start = nil` and the logic should be:

```ruby
unless (shift = start ? start - 1 : 0) == 0
  lines = lines.map {|it| it - shift }
end
lines.sort
```

mojavelinux:
Nice work!

CHANGELOG.adoc
```diff
@@ -37,6 +37,7 @@ Bug Fixes::
   * Escape ellipsis at start of line in manpage output (#3645) (*@jnavila*)
   * Don't register footnote with ID if a footnote is already registered with that ID (#3690)
   * Honor start attribute on ordered list in manpage output (#3714)
+  * Highlight lines relative to the start value (#3701) (*@mogztter*)
```
mojavelinux, line 40: Can you update this to read:

```
Compute highlight line ranges on source block relative to value of start attribute
```

Mogztter: Sure, done :smile:

mojavelinux: Thanks!

mojavelinux:
Can you rebase one more time?

Mogztter:
@mojavelinux Rebased on master and commits squashed.

PR #3710, opened 2020-07-12
update issue guidelines

mojavelinux:


mojavelinux:
@djencks @Mogztter let me know what you think about these changes.

CONTRIBUTING.adoc
```diff
@@ -25,12 +25,24 @@ You also assert that you are the original author of the work that you are contri
...
-Most importantly, since Asciidoctor is a text processor, reproducing most bugs requires that we have some snippet of text on which Asciidoctor exhibits the bad behavior.
+Before submitting a bug report or feature request, check to make sure it hasn't already been submitted using the {uri-issues}[search].
+When submitting a bug report, please include a code sample along with any details that may help reproduce the bug, including your gem version, Ruby version, and operating system.
+If the code sample is very long, you can put it in a {uri-gist}[Gist] and link to it.
 
-An ideal bug report would include a pull request with failing specs.
+When reporting the issue, please use an actionable title and a description of the behavior you expect or want.
+An example of a good title is, "Allow attributes to be defined per list item of an ordered or unordered list."
+Please avoid making disparaging or negative remarks in the issue, such as how much pain the software has caused you or that you're disappointed with the exising behavior.
+This type of commentary is not helpful and only serves to set the issue off on the wrong foot.
```
djencks, line 37: I wonder if there's a way to express this more positively.  I feel a bit like this is not following it's own advice, but I haven't thought of better wording yet.  Perhaps more about "tell us what you want, not what you don't want".

mojavelinux: Good point and suggestion. I'll reword.

Mogztter (approved):
I think it has the right balance 👌

mojavelinux:
Thanks!

PR #3712, opened 2020-07-16
resolves #3543 short-circuit xreftext containing a circular reference path

### LINKED ISSUE #3543: stack level too deep on infinite header xref recursion
```
= asdf

[[h2]]
== h2 xref:h3[]

[[h3]]
=== h3 xref:h2[]
```

then at asciidoctor 20.10:

```
asciidoctor main.adoc
```

fails with:

```
stack level too deep
  Use --trace for backtrace
```

On GitHub: https://github.com/isaacs/github/issues/1718

mojavelinux:


mojavelinux:
@Mogztter This change is rather dramatic, but I think the most logical approach. Some browsers tolerate a nested `<a>`, but it doesn't validate. That's why even though I dislike having to check `include? '<a '`, it's really the right thing to do. Without it, the first heading in the use case from the issue would resolve to:

```
<h2 id="a">A <a href="#b">B <a href="#a">[a]</a></a></h2>
```

We could break the cycle but not try to prevent the nested `<a>`, but then it feels like we're just playing games. I'd like to see someone bring forth a case whereby we would really need that logic.

Mogztter:
The HTML 5.2 specification says:

> Content model:
Transparent, but there must be no interactive content or a element descendants.

https://www.w3.org/TR/html52/textlevel-semantics.html#the-a-element

The living specification is a bit more specific:

> Content model:
Transparent, but there must be no interactive content descendant, a element descendant, or descendant with the tabindex attribute specified.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

I found an interesting article about nested links: https://css-tricks.com/nested-links/
The card component example makes a lot of sense but I don't think it applies here on a "simple" link like xref.
Additionally, it could be solved allowing `href` to be defined on a container, see: https://meyerweb.com/eric/html-xhtml/html5-linking.html

In short, I don't think that we should allow nested `<a>` in this context.
In my opinion, this change is reasonable and according to this article:

```html
<!-- this is invalid and won't render as expected -->
<a href="#one">
  Outside
  <a href="#two">
    Inside
  </a>
</a>
```

will actually produce:

```html
<a href="#one">
  Outside
</a>
<a href="#two">
  Inside
</a>
```

> We could break the cycle but not try to prevent the nested <a>, but then it feels like we're just playing games.

Given the current state of the HTML specification, I don't think Asciidoctor should produce nested links.

mojavelinux:
> In short, I don't think that we should allow nested <a> in this context.

:+1:

mojavelinux:
I think we now have something very reasonable. As it turns out, this is not the first time in the converter we have stripped nested links, so there was already an rx available for it.

PR #3717, opened 2020-07-16
resolves #3376 allow a URL macro to have a preceding single or double quote

### LINKED ISSUE #3376: Double quotes around URL macro prevent it from being converted
Enclosing a URL or URL macro in double quotes prevents it from being recognized and converted. For example:

```
"https://asciidoctor.org[Asciidoctor]"
```

However, it's acceptable to put double quotes around an explicit link macro or xref macro.

```
"link:https://asciidoctor.org[Asciidoctor]"

"xref:index.adoc[Home]"
```

The behavior should be consistent.

mojavelinux:


lib/asciidoctor/substitutors.rb
```diff
@@ -542,14 +542,17 @@ def sub_macros text
...
           prefix = '' if prefix == 'link:'
           text = $4
         else
-          # invalid macro syntax (link: prefix w/o trailing square brackets)
-          # FIXME we probably shouldn't even get here...our regex is doing too much
-          next $& if prefix == 'link:'
+          # invalid macro syntax (link: prefix w/o trailing square brackets or enclosed in double quotes)
+          # FIXME we probably shouldn't even get here when the link: prefix is present; the regex is doing too much
+          case prefix
+          when 'link:', '"', ?'
```
Mogztter, line 553: nitpick, the last token `?'` is equivalent to `"'"` or `'\''` right?
I'm not against this notation but maybe we should use it consistently:

```rb
when 'link:', ?", ?'
```

or, if we want to use single quotes everywhere:

```rb
when 'link:', '"', '\''
```

mojavelinux: I do want to switch to it everywhere, but that's for a separate refactoring. For now, I will make these two consistent (and forward looking).

Issue #2535, opened 2018-01-13
Upgrade to FontAwesome 5


Mogztter:
We should decide if we want to use SVG icons with JavaScript or stick with Web Fonts.
Useful links:

 * SVG with JS: https://fontawesome.com/get-started/svg-with-js
 * Web Fonts: https://fontawesome.com/get-started/web-fonts-with-css
 * Upgrading from version 4: https://fontawesome.com/how-to-use/upgrading-from-4

mojavelinux:
I think we should target this upgrade for 1.6.0.

I'm open to using SVG as long as it doesn't require any additional local files. One of the main reasons we introduced font-based icons was to allow icons to be used without having to lug around other files. I'm not stuck on the font-based part. So if we can offer the same experience, I'm all for SVG. We also need to find a way to allow them to be colorized.

Mogztter:
> I'm open to using SVG as long as it doesn't require any additional local files. 

No, as far as I know we just need to use a JavaScript to convert `icon`:

```html
<head>
  <!--load everything-->
  <script defer src="https://use.fontawesome.com/releases/v5.0.4/js/all.js"></script>
</head>
<body>
  <!--user icon in two different styles-->
  <i class="fas fa-user"></i>
  <i class="far fa-user"></i>
  <!--brand icon-->
  <i class="fab fa-github-square"></i>
</body>
```

Having said that the JavaScript file may load more external ressources.

> I think we should target this upgrade for 1.6.0.

Sound fair enough.

> We also need to find a way to allow them to be colorized.

Yes we can: 

```html
<i class="fas fa-circle" style="color:Tomato"></i>
```

Mogztter:
It would be nice to be able to do "server-side" rendering in Asciidoctor.js: https://fontawesome.com/how-to-use/on-the-web/other-topics/server-side-rendering

john-cj:
Is there an option to see which version of Font Awesome is currently used? I already tried

```
asciidoctor -h fontawesome
```

mojavelinux:
https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L348

mojavelinux:
We are on Font Awesome 4 because Font Awesome 5 is basically a different library.

john-cj:
I see, thanks for link.

OrangeDog:
FWIW I would much prefer continuing with a pure-CSS instead of a JS solution.
[...]

PR #3741, opened 2020-08-22
resolves #3694 add support for style attribute to icon macro

### LINKED ISSUE #3694: Add support for set attribute to icon macro
Although the default HTML converter doesn't yet support Font Awesome 5, the library could be integrated downstream. To accommodate the use of Font Awesome 5, the icon macro should honor the set attribute, at least to specify the Font Awesome set.

Example:

```
icon:acorn[set=fas]
```

Should produce:

```
<i class="fas fa-acorn"></i>
```

martingreilinger:
My proposed changes regarding different icon styles for the icon macro, for detailed information see my comment on the issue #3694 itself.

* The contribution guidelines wish for documentation, so I tried adding basic documentation to the `convert_inline_image` method.
  I did not comment on my change directly, as I did not want to repeat the issue description and clutter the source code.
  But if wished, I can add a short summary. 

* Further I tried cleaning up the related test cases by grouping them based on a common test `context`.

Looking forward to hearing some feedback! 

_NOTE_: If my changes get approved, I recommend renaming the issue from _set_ to _style_ to avoid future confusion.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -1174,9 +1174,15 @@ Your browser does not support the video tag.
     end
   end
 
+  # Convert {AbstractNode}s with the {AbstractNode.node_name} +"inline_image"+ to their associated _HTML_ markup
+  #
+  # node  - the {AbstractNode} to convert
+  #
+  # Returns (String) stringified _HTML_ markup representing the inline image node
   def convert_inline_image node
     if (type = node.type || 'image') == 'icon' && (node.document.attr? 'icons', 'font')
-      class_attr_val = %(fa fa-#{node.target})
+      icon_style = node.attr('style', node.document.attr('icon_style',''))[0]
+      class_attr_val = %(fa#{icon_style} fa-#{node.target})
       { 'size' => 'fa-', 'rotate' => 'fa-rotate-', 'flip' => 'fa-flip-' }.each do |key, prefix|
         class_attr_val = %(#{class_attr_val} #{prefix}#{node.attr key}) if node.attr? key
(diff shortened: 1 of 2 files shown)


mojavelinux:
Thank you for taking the time and effort to prepare this change.

Regarding the name, I want to stick with the existing terminology that we established in Asciidoctor PDF, which is `set` not `style`. I don't want to get into different styles for the same icon as that is a feature specific to Font Awesome. Instead, the set dictates the style (and yet is agnostic to icon set).

Please consult the documentation in Asciidoctor PDF for the behavior I'm looking for. https://github.com/asciidoctor/asciidoctor-pdf#font-based-icons

Thus, if you want the brands style for sass, you would use:

```
icon:sass[set=fab]
```

For now, we assume the necessary icons are loaded already, but this could change in the future. Right now, we aren't even supporting Font Awesome 5, so this would likely have to coincide with #2535.

mojavelinux:
> I tried cleaning up the related test cases by grouping them based on a common test context.

Thanks. It's best to do that in a separate commit. And I realize this test suite needs a lot of work. It's quite old ;)

martingreilinger:
@mojavelinux thanks for the feedback, I wasn't aware this was such a big topic at first. 
How would you suggest to move forward with this PR? It wouldn't be a problem for me to change the implementation as you intended it to be originally and move the test cleanup into a separated commit. 
Do you want to close this PR and ill create a new one, or just push the changes here?

Or is this issue still in the _draft_ stage since you mentioned Font Awesome 5 isn't supported yet? Then I'd be fine with closing the PR and maybe follow up on it once everything is settled.

PR #3749, opened 2020-09-05
Update Truffleruby version

gogainda:


mojavelinux:
I only want to test against truffleruby releases. And truffleruby is tested nightly already.

mojavelinux:
> TRUFFLERUBYOPT="--vm.Xmx3g"

why are you proposing to remove this flag?

gogainda:
> I only want to test against truffleruby releases. And truffleruby is tested nightly already.

in the travis config file I see that only some particular version of truffleruby is being used ( btw can I switch to the just truffleruby which is latest stable version?) and I dont see the place were installation of nightly builds is happening, can you point on it?

gogainda:
> > TRUFFLERUBYOPT="--vm.Xmx3g"
> 
> why are you proposing to remove this flag?

latest builds of truffleruby should consume less memory

mojavelinux:
> can I switch to the just truffleruby which is latest stable version

yes

>  I dont see the place were installation of nightly builds is happening

it's this condition:

    if: type = cron OR commit_message =~ /TruffleRuby/ OR commit_message =~ /truffleruby/

> latest builds of truffleruby should consume less memory

we can try it for this PR. Please comment out the conditional to see if it works. Then we'll reenable it before merging.

gogainda:
please check the latest build, it works fine. I would suggest to leave just```rvm: truffleruby-head``` to run it always and get rid of ```if``` statement

mojavelinux:
> please check the latest build, it works fine

you are correct.

> I would suggest to leave just rvm: truffleruby-head

I want to test stable releases.

> to run it always and get rid of if statement

it's too slow. 6 minutes is not acceptable. therefore, we will only run it nightly and on PRs that mention it (like this one).

gogainda:
> I want to test stable releases.

so you need then ```rvm: truffleruby``` it points out to the latest stable realease

mojavelinux:
> so you need then rvm: truffleruby it points out to the latest stable realease

I want to know exactly what version we are testing. In the past, RVM has not been updated in a timely manner and we end up not testing the version we intent to test. In this build, we always specify exact versions.

.travis.yml
```diff
@@ -17,12 +17,9 @@ matrix:
     env: JRUBY_OPTS='--dev'
   - rvm: *oldest_ruby
     env: ROUGE_VERSION='~> 2.0.0'
-  - rvm: truffleruby-20.1.0
+  - rvm: truffleruby
```
mojavelinux, line 20: Update this to truffleruby-20.2.0 and I will merge.

gogainda:
Looks good now

mojavelinux:
Thanks!

PR #3753, opened 2020-09-13
Remove direction property from default stylesheet

abdnh:
Is there any reason direction has to be set explicitly to LTR?
This can be inconvenient when trying to set the document direction to RTL.


mojavelinux:
I don't see any reason why it has to be there. We inherited from the CSS reset in Foundation 4. We can remove it.

PR #3763, opened 2020-09-26
Docbook backend: Add table width attribute

pyssling:


pyssling:
Ok, this is probably wrong. I didn't realize - docbook 5.2 now supports XHTML tables.

PR #3768, opened 2020-09-29
Use vendored libxml2/libxslt libraries instead of systems' ones

tnir:
cf. https://nokogiri.org/tutorials/installing_nokogiri.html#install-with-included-libraries-recommended

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,6 @@ matrix:
     if: type = cron OR commit_message =~ /TruffleRuby/ OR commit_message =~ /truffleruby/
 env:
   global:
-  # use system libraries to speed up installation of nokogiri
-  - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
   - PYGMENTS_VERSION='~> 1.2.0'
   - SOURCE_DATE_EPOCH=1521504000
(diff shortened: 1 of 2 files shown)


tnir:
@mojavelinux What do you think this?

mojavelinux:
The reason I suggest linking against the system libraries is because it makes installation must faster (7s vs 35s). That's because if you use the vendored library, it must be compiled. Linking against the system libraries is critical for CI, so I don't want to change it there. But what I could do is explain the suggestion to use system libraries and make it clear it's not required. In other words, you can install nokogiri without having libxml2/libxslt.

mojavelinux:
See cabde9c25fe040221ad71bb6aaf8754d2e5ec099.

tnir:
Thanks for updating the doc. Unlike the ancient years, we do have many caching mechanism of Rubygems on most modern CI systems, not so many developers do not have to do what you described in cabbe9c. (cf. 2.7 sec in total for `bundle install` at https://travis-ci.org/github/jekyll/jekyll/jobs/733146010#L297 after megring jekyll/jekyll#8409, which is similar to this PR.)

Anyway, again thanks for fixing it around this. Let me close this PR.

PR #3769, opened 2020-09-29
Use Ubuntu bionic (18.04 LTS) instead of trusty (14.04 LTS)

tnir:
14.04 LTS has reached its end of support lifecycle and 20.04 LTS is still relatively new.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 git:
   # use depth 2 just in case two refs get pushed at once (like a tag)
   depth: 2


mojavelinux:
The reason I hadn't done this is because the Pygments tests fail on JRuby 9.1. However, I think it's possible to downgrade the OS per entry in the matrix, so I can hold the JRuby 9.1 run back.

mojavelinux:
Yep, all I need to do is locked the dist for the JRuby 9.1 build to trusty. I'll merge, then make that updated. Thanks!

PR #3782, opened 2020-10-17
docbook backend: restore reftext after anchor

pyssling:
Was previously dropped - seemingly by accident - add to unit test.


mojavelinux:
Wow, you're absolutely right. I very surprised this regression got overlooked. But as you have pointed out, the test was not thorough enough.

mojavelinux:
Thanks!

PR #3795, opened 2020-10-28
resolves #3387 rework table borders to leverage border collapsing

### LINKED ISSUE #3387: Use cell borders instead of table border
I think we should use cell borders (configured to collapse) instead of an outer border defined on the table. The rendering looks the same but the main advantage is that if the table is spread on multiple pages then the border will not be visible when the row ends at the end of a page (see the bottom of the page) and the repeated table header will have a top border (see the top of the page).

Initial discussion: https://github.com/Mogztter/asciidoctor-pdf.js/pull/10

mojavelinux:


mojavelinux:
I verified that this change fixes the case originally reported here: https://github.com/Mogztter/asciidoctor-web-pdf/pull/10.

It's interesting to note that the same issue still appears in Firefox when printing, though I can see no way of fixing that. It doesn't seem to use any border no matter where we put it.

Mogztter:
Thanks 👍

> It's interesting to note that the same issue still appears in Firefox when printing, though I can see no way of fixing that. It doesn't seem to use any border no matter where we put it.

That's unfortunate but it's probably a bug in Firefox.

mojavelinux:
Yep, that's what I'm thinking.

PR #3805, opened 2020-10-30
resolves #3804 link the notitle and showtitle attributes

### LINKED ISSUE #3804: Link the showtitle and notitle attributes
Currently, the showtitle and notitle attributes both affect whether the document title is included in embedded output, but they work independently (and thus don't always agree). These two attributes should be linked so they do not operate independently. In other words, setting showtitle should unset notitle and setting notitel should unset showtitle.

mojavelinux:


mojavelinux:
@Mogztter This may require a change in the browser extension. I noticed that it sets the showtitle attribute directly after the document is loaded. That attribute will no longer be checked (and bypasses this linking). Instead, the code should delete the notitle attribute. To make it compatible with past releases, though, you can also still assign the showtitle attribute.

Mogztter:
I've created a "reminder issue" 😅

PR #3807, opened 2020-10-31
iso 639 code for Korean is ko, not kr

jnavila:
kr is for Kanuri


mojavelinux:
Confirmed according to https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?iso_639_1=ko. Thanks!

PR #3816, opened 2020-11-01
resolves #3815 add CSS class to support pre-wrap role to preserve leading, trailing, and repeating spaces in phrase

### LINKED ISSUE #3815: Add class for pre-wrap role to preserve leading, trailing, and repeating spaces in phrase
There are cases when leading, trailing, and repeating spaces an in phrase (typically one representing computer code) are significant. Add CSS support for a built-in role named pre-wrap that preserves these spaces.

mojavelinux:


mojavelinux:
This change also prevents the roles from being assigned to a pre element, since that conflicts with existing styles.

PR #3821, opened 2020-11-03
enable GitHub Actions to run CI jobs on Windows

mojavelinux:


mojavelinux:
It looks like there are problems running the test suite on JRuby on Windows. More investigation will be needed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Configure Bundler
+      run: |
+        bundle config --local path .bundle/gems
+        bundle config --local without docs lint coverage
+    - name: Install dependencies
+      run: |
+        bundle --jobs 3 --retry 3
+    - name: Test
+      run: |
+        bundle exec ruby -w $(bundle exec ruby -e 'print File.join Gem.bindir, %q(rake)') test:all
```
slonopotamus, line 47: I forgot again, why you're doing this dance instead of just `bundle exec rake test:all`?

mojavelinux: So that I can enable `-w` which Rake doesn't provide.

slonopotamus: `bundle --verbose` is doing something different?

mojavelinux: I'm not sure what you are asking. I need Ruby warnings turned on, and before anything is loaded. There are various ways to do it (some that toggle the switch to late). This is the most reliable.

(I don't understand why rake doesn't support -w, but that is not a battle I feel like fighting).

mojavelinux: I think using RUBY_OPT works too.

mojavelinux: Nope, RUBY_OPT doesn't work in this case.

slonopotamus: Okay, forget it then.

mojavelinux: Exactly. I tried for hours to try to enable script warnings when calling rake any other way. This is the only way that works.

slonopotamus:
Oh, I remembered that Pygments hangs on JRuby + Windows. Do you run tests with Pygments?

test/reader_test.rb
```diff
@@ -855,6 +855,7 @@ class ReaderTest < Minitest::Test
       end
 
       test 'unreadable file referenced by include directive is replaced by warning' do
+        next if windows? # JRuby on Windows runs this even with the conditional on the block
```
slonopotamus, line 858: Is there a reason why you're not using Rspec `skip``? It allows to specify message and makes it clearly visible in test results which tests were skipped.

mojavelinux: Sure, I could do that too. Thanks for the idea. I'm just focusing on getting the test suite to pass. Perfection is not my goal right now.

mojavelinux: I either need to switch away from IO.popen or just disable the CLI tests that are failing. The reason I'm using IO.popen is because there was a bug in popen3 that was causing the test to fail in some version of JRuby. But that could be outdated now. The point is, I didn't pick stuff randomly. Getting these CLI tests to work on all Rubies and all platforms takes hours, if not days. And I hate it every time I have to touch them.

mojavelinux:
I've decided to break this up so that the migration from AppVeyor and adding JRuby to the matrix are handled separately.

PR #3867, opened 2020-11-26
Add the 'horizontal' node.style similar to html5.rb

phidebian:
Hi, 
This is a tiny patch to manpage.rb to get it on pare with html5.rb regarding dlist formating, i.e list of options, exit statis list etc....
I join a tiny `.ad` file here to demo the fix.

I am not fluent in ascidoctor development enough, to provide QA tests, neither I can fix the documentation thats seems to be behind de developement code, because I dicovered the `horizontal` style by reading the source code and find no evidence of it in the docco.

With this fix and `[horizontal]` placed in the `.ad` file I got this
rendering from man(1).

```
OPTIONS
       -s     short option name, doc start on the same line.

       -t     short option name, doc start on the same line yet continue to
              next line with some more usefull informations that can be safely
              ignored.

       -l, --long
              Long option name, doc start on the next line yet continue to
              next line with some more usefull informations that can be safely
              ignored.

EXIT STATUS
       0      Success.

       1      Failure (syntax or usage error; configuration error; document
              processing failure; unexpected error).

```
Cheers,
Phi
[man-short.txt](https://github.com/asciidoctor/asciidoctor/files/5603497/man-short.txt)

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -201,6 +201,9 @@ r lw(\n(.lu*75u/100u).'
         result << %(.sp
 #{counter}. #{manify terms.map {|dt| dt.text }.join ' '}
 .RS 4)
+      when 'horizontal'
+        result << %(.TP
+#{manify terms.map {|dt| dt.text }.join(', '), whitespace: :normalize})
       else
         result << %(.sp
 #{manify terms.map {|dt| dt.text }.join(', '), whitespace: :normalize}
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -554,14 +554,38 @@ Your browser does not support the audio tag.
       end
       node.items.each do |terms, dd|
         result << '<tr>'
-        result << %(<td class="hdlist1#{(node.option? 'strong') ? ' strong' : ''}">)
+
+        len=0 
+        labelspan=(node.attr 'labelspan').to_i
+        if(labelspan!=0)      
+          terms.each do |dt|  
+            termstr=dt.text
+            termstr.gsub!(/<[^>]+>/,'')
+            if(termstr.length>len)
+              len=termstr.length;
+            end
+          end
+
+          if(len<labelspan)
+            result << %(<td class="hdlist1#{(node.option? 'strong') ? ' strong' : ''}">)
+          else
+            result << %(<td class="hdlist2#{(node.option? 'strong') ? ' strong' : ''}"  colspan="2">)
+          end
+        else
+          result << %(<td class="hdlist1#{(node.option? 'strong') ? ' strong' : ''}">)
+        end
+
         first_term = true
         terms.each do |dt|
           result << %(<br#{slash}>) unless first_term
           result << dt.text
           first_term = nil
         end
-        result << '</td>'
+        if( (labelspan==0) || (len<labelspan) )
+          result << '</td>'
+        else
+          result << '</td></tr><tr><td></td>'
+        end
         result << '<td class="hdlist2">'
         if dd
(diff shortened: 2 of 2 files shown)


mojavelinux:
Thanks. I'd definitely like to see support for the horizontal list in the manpage backend.

The horizontal dlist is covered by this section in the docs: https://asciidoctor.org/docs/user-manual/#description-list

phidebian:
thanx

mojavelinux:
Please make a separate PR for the change to the HTML5 converter. I would like to get the horizontal style support in the manpage converter first. Then we can discuss improvements to the HTML5 converter.

mojavelinux:
Regarding the change to the HTML, the problem with this approach is that we cannot know for sure how many characters the label takes up just by looking at the formatted text. That's because the text may have character references and double-wide characters. So this is not going to be 100% accurate. The HTML layout change seems like something that should be done in a custom converter. Better HTML semantics may even make it possible with CSS. I'm also not convinced about the attribute name labelspan. That doesn't feel intuitive or descriptive. Something like labelmaxchars would be better (restricting the column to that many chars and overflowing as needed).

mojavelinux:
I just tried it and it's possible to achieve this layout using CSS floats. However, it requires using a role so the styles could be applied only when you want them.

Here's the AsciiDoc I tested:

```
[.smart-hdlist]
short term:: the description of this term
a much longer term:: the description of this term
```

and here's the CSS that goes with it:

```css
<style>
.smart-hdlist dt {
  margin-right: -10%;
  min-width: 10%;
  float: left;
  clear: left;
}

.smart-hdlist dd {
  min-width: 90%;
  margin-left: 10%;
  padding-left: 1.125em;
  float: left;
}
</style>
```

You could use calc to avoid the change to margin-left:

```css
.smart-hdlist dt {
  margin-right: -10%;
  min-width: 10%;
  float: left;
  clear: left;
}

.smart-hdlist dd {
  min-width: calc(90% - 1.125em);
  margin-left: calc(10% + 1.125em);
  float: left;
}
</style>
```

mojavelinux:
Here is the width expressed in ex, which is roughly equivalent to the width of a character.

```css
<style>
.smart-hdlist dt {
  margin-right: -10ex;
  min-width: 10ex;
  float: left;
  clear: left;
}

.smart-hdlist dd {
  min-width: calc(100% - 11ex - 1.125em);
  margin-left: calc(10ex + 1.125em);
  float: left;
}
</style>
```

For some reason, the min-width for the description needs to be 1ex smaller. It seems to be a browser quirk when using calc with 100%.

mojavelinux:
Another idea is to add a margin left on the dl, then put a negative margin left on the dt element. That should prevent the term from overflowing the right margin of the page.

lib/asciidoctor/converter/manpage.rb
```diff
@@ -201,6 +201,9 @@ def convert_dlist node
         result << %(.sp
 #{counter}. #{manify terms.map {|dt| dt.text }.join ' '}
 .RS 4)
+      when 'horizontal'
+        result << %(.TP
+#{manify terms.map {|dt| dt.text }.join(', '), whitespace: :normalize})
```
mojavelinux, line 206: I tried this and it doesn't seem to have any effect. Can you clarify how the TP macro works?

mojavelinux:
This state of this PR is not adequate for merging. First, it has changes do not fit within the scope of the title. It's changing both the man page converter and the HTML 5 converter. The change to the HTML 5 converter should be separate. I also didn't get a response about what the TP macro does in troff. The PR is also missing tests. Therefore, I'm closing this as incomplete.

I'm still in favor of adding support for the horizontal dlist layout in the man page converter. However, the change needs to be done properly. And to do that, it should start by filing an issue where we can discuss the change before submitting a PR.

PR #3874, opened 2020-12-02
feat: enable running tests as root

mikemckiernan:
I've been using rootless containers with podman
and I end up running the tests as root.


mojavelinux:
I've observed this failure several times, but never put together why it was happening. Thanks!

PR #3899, opened 2020-12-30
#3641 first attempt to implement Rouge linenumbers inline

### LINKED ISSUE #3641: Implement support for rouge-linenums-mode (add inline mode)
I'm using asciidoctor 2.0.10 to generate HTML. I use Rouge for syntax highlighting.

According to https://asciidoctor.org/docs/user-manual/#rouge I should use 
```
:rouge-linenums-mode: inline
```
for sourceblocks that have long lines. However, that option doesn't seem to work. It doesn't make any difference if I use 'table' or 'inline'.

In both cases the codeblock renders as a table with 2 cells, one containing a '<pre>' tag with all the line numbers, one with the lines as spans wrapped in 1 '<pre>' tag.

On a large block of long lines the linenumbers end way earlier than the actual lines.

This code: https://rouge-ruby.github.io/docs/Rouge/Formatters/HTMLLineTable.html suggests there should be a row with a linenumber + source code line per line.

But that class is never used in https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/syntax_highlighter/rouge.rb

I'm not good in Ruby, so I might be overlooking something. If so, please let me know.

hepabolu:
The tests won't run completely because they fail on the annotations. I haven't been able to get past the failing tests, so I don't know what else is wrong.

resolves #3641

### DIFF
--- a/lib/asciidoctor/rouge_ext.rb
+++ b/lib/asciidoctor/rouge_ext.rb
@@ -18,6 +18,33 @@ module Asciidoctor; module RougeExt; module Formatters
     end
   end
 
+  class HTMLLineTable < ::Rouge::Formatter
+    def initialize delegate, opts
+      @delegate = delegate
+      @start_line = opts[:start_line] || 1
+    end
+
+    def stream tokens
+      lineno = @start_line - 1
+      buffer = [%(<table class="linenotable"><tbody>)]
+      token_lines(tokens) do |line_tokens|
+        lineno += 1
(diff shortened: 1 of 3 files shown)


hepabolu:
The tests fail because the tests for the annotations fail.

mojavelinux:
Superseded by #4129. I prefer to use our own simple implementation rather than try to use the more elaborate one that Rouge has. Besides, the one from Rouge introduces ID conflicts in the document when there is more than one source block with line numbering enabled.

PR #3901, opened 2021-01-02
speedup CI by using Bundler cache

slonopotamus:
Before: 40s to install Ruby + Bundler and [around 9.5-10 min for bundle install on Windows + Ruby 3.0](https://github.com/asciidoctor/asciidoctor/runs/1635368583?check_suite_focus=true#step:10:1) (slowest platform currently, not sure why. Does it build libxslt from sources for Nokogiri instead of using prebuilt binary?)

After: [around 50s for Ruby + Bundler + bundle install combined on Windows + Ruby 3.0](https://github.com/slonopotamus/asciidoctor/runs/1637276403?check_suite_focus=true#step:7:1)

I'm not 100% happy with this change because we lose ability to do custom `bundle config` before installing deps, but benefits are pretty huge. @eregon do you want a feature request for this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,22 +53,6 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
-    - name: Install Ruby
-      uses: ruby/setup-ruby@v1
-      with:
-        ruby-version: ${{ matrix.ruby }}
-    - name: Configure Bundler
-      run: |
-        bundle config --local path .bundle/gems
-        bundle config --local without docs
-    - name: Configure Nokogiri installation (Linux)
-      if: matrix.os == 'ubuntu-latest'
-      run: |
-        bundle config --local build__nokogiri --use-system-libraries
-        sudo apt-get install libxslt1-dev
-    - name: Configure Nokogiri installation (macOS)
-      if: matrix.os == 'macos-latest'
-      run: bundle config --local build__nokogiri --use-system-libraries
     - name: Set AsciiMath version
       if: matrix.asciimath-version
       run: echo 'ASCIIMATH_VERSION=${{ matrix.asciimath-version }}' >> $GITHUB_ENV
@@ -78,8 +62,20 @@ jobs:
     - name: Set Rouge version
       if: matrix.rouge-version
       run: echo 'ROUGE_VERSION=${{ matrix.rouge-version }}' >> $GITHUB_ENV
-    - name: Install dependencies
-      run: bundle --jobs 3 --retry 3
+    - name: Install libxslt for Nokogiri (Linux)
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt-get install libxslt1-dev
+    - name: Install Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: ${{ github.event_name != 'schedule' }}
+    - name: Install dependencies without cache
+      if: github.event_name == 'schedule'
+      shell: bash
+      run: |
+        bundle config --local path vendor/bundle
+        bundle install --jobs 4
     - name: Run tests
       run: bundle exec ruby -w $(bundle exec ruby -e 'print File.join Gem.bindir, %q(rake)') test:all
     - name: Build dependents
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,10 +49,17 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       COVERAGE: ${{ matrix.primary }}
+      BUNDLE_WITHOUT: docs
       SOURCE_DATE_EPOCH: '1521504000'
     steps:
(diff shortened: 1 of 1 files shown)


eregon:
> I'm not 100% happy with this change because we lose ability to do custom `bundle config` before installing deps,

That's just `bundle config --local without docs` in this case, right?
One can set an env var to keep exclude such groups:
```
BUNDLE_WITHOUT="development:test"
```
I guess we should document this in setup-ruby, it's the easiest and most efficient way, PR welcome.

mojavelinux:
By using the cache, we throw out the verification that Asciidoctor can be installed and tested from a clean slate. Pulling from a cache can give us a false positive. I suppose we could consider disabling the cache for nightly builds so we can still get that confirmation.

We do need to figure out how to set bundler options though, because I don't want to change how we run the test suite. Can we control the install path using an env var?

slonopotamus:
> Can we control the install path using an env var?

setup-ruby has reasons why they use `vendor/bundle` path. See `bundle install --deployment` [docs](https://bundler.io/man/bundle-install.1.html#DEPLOYMENT-MODE)

mojavelinux:
I'm fine with vendor/bundle. What matters to me is that it's local to the
project. I don't much care for the folder name vendor/bundle as it's
obscure, but as this is CI, we don't see it anyway.

It does seem odd that any of this is linked to using the cache, though.
These seem like orthogonal concerns.

eregon:
> It does seem odd that any of this is linked to using the cache, though. These seem like orthogonal concerns.

What is cached is the contents of `vendor/bundle`, so that's how they are related.

> By using the cache, we throw out the verification that Asciidoctor can be installed and tested from a clean slate.

`bundle install` is still run, but gems will already be in vendor/cache, unless there is a new release of a gem/dependency in the Gemfile. And of course gems in the cache must have been installed sucessfully from scratch at some point.
I've never seen a case where it would install with the cache but wouldn't without, so I think that's very rarely an issue.
But if you wanted to verify that, some scheduled job not using the cache and doing just `bundle install` would be one way, indeed.

mojavelinux:
> I've never seen a case where it would install with the cache but wouldn't without, so I think that's very rarely an issue.

It may be an edge case, but it's a real issue. If you rely on a cache, then you aren't testing it the way the user will use it. And we have run into issues with this in the past. One case was a corrupt gem in rubygems.org. But I do agree we don't need this level of verification on every commit.

slonopotamus:
I'm not sure I want to add all the complexity needed to run periodic builds without cache. Someone could possibly investigate why Ruby 3 + Windows is so slow to install deps and if it is caused by Nokogiri lacking prebuilt binaries, there's already [nokogiri-1.11.0](https://github.com/sparklemotion/nokogiri/releases/tag/v1.11.0) that adds prebuilt binaries for Ruby 3.0. Though at the same time, nokogiri-1.11.0 drops support for Ruby 2.3 + 2.4 (they both are EOL).

eregon:
My understanding from above was that @mojavelinux was fine with this PR, isn't it? Or did I misunderstand something?

If nightly builds without cache is a requirement, it could be done by adding a conditional `bundle install` step, the condition can be something like:
```yaml
if: ${{ github.event_name }} == 'schedule'
```

slonopotamus:
We also need to disable cache in nightly build somehow. Is it possible to specify an _expression_ inside `bundle-cache:` parameter? Or we'll have to duplicate whole step and conditionally execute only one of them?

eregon:
I think `bundler-cache: ${{ github.event_name != 'schedule' }}` should work.

slonopotamus:
Okay, lemme try this...

mojavelinux:
> I think `bundler-cache: ${{ github.event_name != 'schedule' }}` should work.

:100:

mojavelinux:
I did a bit of rearranging and also left in the --use-system-libraries flag for nokogiri since that's a separate change. I'm happy with how this works so I'm prepared to merge.

.github/workflows/ci.yml
```diff
@@ -62,20 +69,16 @@ jobs:
...
         bundler-cache: ${{ github.event_name != 'schedule' }}
-    - name: Install dependencies without cache
+    - name: Install dependencies (scheduled build only)
       if: github.event_name == 'schedule'
-      shell: bash
       run: |
-        bundle config --local path vendor/bundle
-        bundle install --jobs 4
+        bundle config --local path .bundle/gems
+        bundle --jobs 3 --retry 3
```
slonopotamus, line 81: Err... `install` got lost?

mojavelinux: install is the default command, so there's not need to specify it. I tend not to use it in the CI jobs in the Asciidoctor ecosystem, so I was just being consistent with what I've done elsewhere.

slonopotamus: > install is the default command

I see, didn't know that.

mojavelinux:
The build truly is faster and I'm very happy with the arrangement we came up with. Thanks for seeing this through, @slonopotamus! And thanks for the input @eregon!

PR #3909, opened 2021-01-08
resolves #3969 upgrade pygments.rb to 2.0.0

### LINKED ISSUE #3969: Migrate Pygments integration to pygments.rb 2
The current integration with Pygments for syntax highlighting (provided by the `SyntaxHighlighter::PygmentsAdapter` class) depends on pygments.rb 1.2.x. The 1.2.x version of pygments.rb bundles an older version of Pygments, requires Python 2 (not Python 3), is prone to timeout errors invoking Python, and is considered deprecated. Fortunately, there's a new version available that's being maintained by a contributor in the Asciidoctor community (`@slonopotamus`). We should modernize this integration adapter by upgrading to this new version.

Since pygments.rb is an optional dependency of Asciidoctor, upgrading the integration will require the user to make at least two changes:

1. Upgrade the pygments.rb gem (Asciidoctor itself cannot activate this change)
2. Ensure Python 3 is available on the system
3. If using a custom version of Pygments, ensure it's compatible with Python 3

Since these upgrades will produce a result that is backwards compatible, I'm comfortable saying we can make this change in a minor version of Asciidoctor. I'm not super concerned about (3) since it's a very rare situation and also the responsibility of the user. (It's outside what we can control).

We do need to communicate very clear in the release notes that users who are using Pygments for syntax highlighting will need to take action for their build to continue to work. But once they do, they should expect the same result.

slonopotamus:


docs/modules/syntax-highlighting/pages/pygments.adoc
```diff
@@ -8,16 +8,14 @@
...
 The pygments.rb gem manages calls to Pygments, which is an external program that runs using Python.
 
 TIP: You do not need to install Pygments itself.
 It comes bundled with the pygments.rb gem.
 
-IMPORTANT: You must have Python 2 installed to use pygments.rb.
-Python 3 is not compatible with the pygments.rb gem.
-Check that you have a `python2` (Linux), `python` (macOS), or `py -2` (Windows) executable on your PATH.
-(On macOS, verify that the `python` executable uses Python 2 by running `python -V`).
+IMPORTANT: You must have Python 3 installed to use pygments.rb.
```
mojavelinux, line 17: I think we should clarify here that you can have either Python 2 or Python 3. But if you must use Python 2, then you have to use the older (unsupported version) of pygments.rb. Of course we are discouraging it, but the information should help those who are trying to migrate.

slonopotamus: All this page needs a major rework. It tells users to do scary things to workaround the fact that pygments.rb cannot use non-bundled Pygments version currently.

mojavelinux: We're not doing a general review of the docs right now (and beside, most of this will be removed now that docs.asciidoctor.org is online anyway). So the revisions happen later. For now, we just cannot make assertions that aren't entirely true. Python 3 is not all of a sudden required. It is required if you use pygments.rb 2. If you use pygments.rb 1, then Python 2 is required. That's all I'm asking.

slonopotamus: Sure, no problem. Besides that, tests need to be fixed before this can proceed anywhere :)

slonopotamus:
```
    test 'should not hardcode inline styles on lineno spans when linenums are enabled and source-highlighter is pygments' do
      input = <<~'EOS'
      :source-highlighter: pygments
      :pygments-css: inline
      :pygments-linenums-mode: inline

      [source%linenums,ruby]
      ----
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      puts 'Hello, World!'
      exit 0
      ----
      EOS

      output = convert_string_to_embedded input, safe: :safe
      assert_css 'table.linenotable', output, 0
      assert_css 'pre', output, 1
      assert_includes output, '<span class="lineno"> 1 </span>'
      assert_includes output, '<span class="lineno">10 </span>'
    end
```

I don't understand this test.
_Why_ lineno style should not be inline?
Does it expect `StyledLinenoSpanTagRx` from `lib/asciidoctor/syntax_highlighter/pygments.rb` to un-inline it back?
My ma [told me](https://stackoverflow.com/a/1732454) never touch HTML with regexes :)

mojavelinux:
Because it makes it difficult to style/control from the stylesheet. And it's not something we want Pygments controlling because it's idea of layout is very different from ours.

slonopotamus:
I see, you want to undo pygments styling and apply your own. Should I then rewrite regex so it matches both pygments.rb 1.2.1 and 2.0.0 CSS?

mojavelinux:
That will be necessary. I'm very surprised that it has changed. That means we're going to need to look closer at this integration because we do groom the HTML a bit.

slonopotamus:
There are three changes WRT line numbers:

1. Now text color is specified for line numbers
2. They stopped setting top/bottom paddings, that's https://github.com/pygments/pygments/commit/20fb8fd85f37d79b37ac79f14774b9cc4ae77731
3. There's no more space after number itself

slonopotamus:
Okay, I reworked this PR so there are no changes to tests. Difference in markup is handled within highlighter class.

mojavelinux:
Thanks for taking care of that, Marat.

I filed an issue to track this change since it will need to be clearly communicated.

Am I correct in saying that the adapter will only work with pygments.rb 2 (and not pygments.rb 1.2)? If so, I'm fine with that. Or does the adapter attempt to work with either version?

slonopotamus:
Adapter works with both of them. Old regex is still [there](https://github.com/asciidoctor/asciidoctor/pull/3909/files#diff-2b67ee5ada7d09e201ee17af1f0bc041c9f8b5bd5ec9d3585a41257a8a08b255R143).

slonopotamus:
I don't know whether you want to continue testing against 1.x on CI. If yes, we need to adjust build matrix somehow.

mojavelinux:
Thanks for clarifying that. So the question becomes, do we want to keep supporting pygments.rb 1.2.x? One possibility is to support it for 2.0.x, then drop it for 2.1.x. Another is to support it indefinitely (which seems problematic given that it's unlikely there will be more pygments.rb 1.2.x releases). Another possibility is to wait until 2.1.x to merge this and drop support for pygments.rb 1.2.x at the same time (though that seems drastic since we could just merge it now).

mojavelinux:
> Old regex is still there.

The regular expression should be selected based on the version of pygments.rb rather than just trying each one in turn. Can you update the code to detect the version and only use the old regex if pygments.rb is < 2?

mojavelinux:
> I don't know whether you want to continue testing against 1.x on CI. If yes, we need to adjust build matrix somehow.

Yes. Though updating the matrix is easy. We just need to add an additional entry and set the pygments-version on that entry. There's no need to run that through every version of Ruby since we are are testing the integration, not Ruby.

slonopotamus:
I'd like to share some more thoughts on this topic.

Asciidoctor does not signal user which versions of optional dependencies it supports (via gemspec of whatever). I actually think this is a problem. pygments.rb is trying to be a good citizen and released a major version bump because it _does break stuff_. But given that Asciidoctor says "just run `gem install pygments.rb`", users now get the latest and the greatest pygments.rb 2.x.

> The regular expression should be selected based on the version of pygments.rb rather than just trying each one in turn. Can you update the code to detect the version and only use the old regex if pygments.rb is < 2?

Regex selection depends not on pygments.rb version but instead on version of _underlying Pygments_. And several distros package pygments.rb 1.x in such a way that it talks to a modern Pygments. So, while depending on pygments.rb version would work for _bundled_ Pygments, it would be wrong in other setups. Actually, it means that on those distros Asciidoctor regexes are not working for a long time already.

> it's unlikely there will be more pygments.rb 1.2.x releases

Yep, this is very unlikely.

mojavelinux:
> Regex selection depends not on pygments.rb version but instead on version of underlying Pygments. 

Thanks for that clarification.

Is there any other way we can detect the version of Pygments? Perhaps via a separate call that routes through pygments.rb?

> Asciidoctor does not signal user which versions of optional dependencies it supports (via gemspec of whatever). I actually think this is a problem.

We can address this in the documentation. Ultimately, the documentation states what we do and do not support. There's no way in Ruby to specify a version range for optional dependencies that gets enforced, so that's just a reality we have to deal with. Plus, as you mentioned, packagers will often ignore those versions anyway, which brings us back to the documentation being the authority.

slonopotamus:
> Is there any other way we can detect the version of Pygments?

There isn't such API currently. I've created pygments/pygments.rb#226 but it is too late. Do you _actually_ want to rely on specific version numbers? This is a similar case to what people do with browser compatibility. Instead of "if this is firefox newer than X, we can use Y", they do "if we have Y, we can use it". So, if there's text to replace, just replace it.

mojavelinux:
> Do you actually want to rely on specific version numbers?

Not really. The scenario seemed simpler in my head (pygments.rb 1 vs 2). But, as you've explained, it's more complex than that.

I'm inclined to go with your approach and reasoning, though I'd like to take a day or two to give it some more thought. I certainly don't want to overthink it since the logic is what I would consider to be low risk (it won't break something critical in the page if the match stops working one day).

slonopotamus:
I thought about this regex issue _a bit more_... Do we actually want pygments to use inline styling? Maybe what we want instead is to *always* use CSS classes, but put them either in a separate CSS file or into document header (so this becomes what we call "inline CSS")?

mojavelinux:
The inline style is the best approach for embedded documents and I know there are both users and tools out there that rely on it. We could consider a third mode, but I don't think it would be wise to remove this mode.

mojavelinux:
I figured out how to rewrite the lineno spans using a single regexp that works reliably across pygments.rb 1.x and 2.x. In fact, it's more robust than it was.

Regarding the `<span></span>` reference, Pygments has always used this to denote the start of the highlighted code, so I think we can safely rely on that. We could get more precise if we wanted, but for now I'm happy taking this approach.

mojavelinux:
...oops, missed a case. Looking into it.

mojavelinux:
I also restored the docs about the Pygments timeout, but made it clear it's only relevant when using pygments.rb 1.x.

mojavelinux:
@slonopotamus Do you approve of my changes? If so, do you agree this can now be merged?

slonopotamus:
Hehe, GitHub doesn't allow me to approve changes because this initially was my own PR :) Yeah, let's go with it.

mojavelinux:
Thanks Marat!

PR #3944, opened 2021-02-14
resolves #3939 counter should not override locked attributes

### LINKED ISSUE #3939: counter should not override locked attributes
`counter` has a loophole that allows to override any attribute (with an arbitrary value) even when this attribute is locked.

Here's a simple reproduction case:

```console
$ asciidoctor -v
Asciidoctor 2.0.12 [https://asciidoctor.org]
Runtime Environment (ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

```console
$ echo '{counter2:imagesdir:/noop} image:hello.png[]' | asciidoctor -a imagesdir=/img -s -
<div class="paragraph">
<p> <span class="image"><img src="/noop/hello.png" alt="hello"></span></p>
</div>
```

As you can see above, the `imagesdir` attribute is locked via the CLI (using `-a imagesdir=/img`) but the `counter2` attribute reference overrides the value from `/img` to `/noop`.
As a result, the `src` attribute on the `<image>` element is `/noop/hello.png` instead of `/img/hello.png`.

Mogztter:
resolves #3939

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 Bug Fixes::
 
+  * Don't allow counter and counter2 directives in attribute reference to override locked attributes (#3939)
   * Update default stylesheet to remove dash in front of cite on nested quote block (#3847)
   * Don't escape hyphen in manname in manpage output
   * Remove extra .sp before content of verse block in manpage output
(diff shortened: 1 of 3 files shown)


Mogztter:
The following test is failing on macOS...? :thinking: 

```
Minitest::Test::TestSyntaxHighlighter::TestPygments#test_should_restore_callout_marks_to_correct_lines_if_table_line_numbering_is_enabled
```

mojavelinux:
I modified the fix to return the attribute value if locked. That way, it still functions as an attribute reference. It just doesn't count.

PR #3952, opened 2021-02-20
resolves #3953 refactor creation of lexer and formatter

### LINKED ISSUE #3953: Extract methods for creating lexer and formatter in Rouge adapter
To make the Rouge adapter easier to extend, extract methods for creating the lexer and formatter, respectively.

Oblomov:
As a followup to the discussion on
https://discuss.asciidoctor.org/Interpret-links-or-other-AsciiDoc-syntax-only-in-specific-lines-of-a-source-block-td8499.html#a8517
this patch moves the lexer and formatter selection of the Rouge
syntax highlighter onto their own functions. This should make it easier
to extend the highlighter with custom lexers and/or highlighters without
having to replicate much of the logic in the highlight method.

### DIFF
--- a/lib/asciidoctor/syntax_highlighter/rouge.rb
+++ b/lib/asciidoctor/syntax_highlighter/rouge.rb
@@ -13,34 +13,18 @@ class SyntaxHighlighter::RougeAdapter < SyntaxHighlighter::Base
   end
 
   def highlight node, source, lang, opts
-    if lang.include? '?'
-      # NOTE cgi-style options only properly supported in Rouge >= 2.1
-      if (lexer = ::Rouge::Lexer.find_fancy lang)
-        unless lexer.tag != 'php' || (node.option? 'mixed') || ((lexer_opts = lexer.options).key? 'start_inline')
-          lexer = lexer.class.new lexer_opts.merge 'start_inline' => true
-        end
-      end
-    elsif (lexer = ::Rouge::Lexer.find lang)
-      lexer = lexer.tag == 'php' && !(node.option? 'mixed') ? (lexer.new start_inline: true) : lexer.new
-    end if lang
-    lexer ||= ::Rouge::Lexers::PlainText.new
     @style ||= (style = opts[:style]) && (style_available? style) || DEFAULT_STYLE
-    if opts[:css_mode] == :class
-      @requires_stylesheet = true
-      formatter = ::Rouge::Formatters::HTML.new inline_theme: @style
+    @requires_stylesheet = true if opts[:css_mode] == :class
+
+    lexer = create_lexer node, source, lang, opts
+    formatter = create_formatter node, source, lang, opts
+
+    highlighted = formatter.format lexer.lex source
+    if opts[:number_lines] && opts[:callouts]
+      [highlighted, (idx = highlighted.index CodeCellStartTagCs) ? idx + CodeCellStartTagCs.length : nil]
     else
-      formatter = ::Rouge::Formatters::HTMLInline.new (::Rouge::Theme.find @style).new
+      highlighted
     end
-    if (highlight_lines = opts[:highlight_lines])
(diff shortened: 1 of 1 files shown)


mojavelinux:
Very nice proposal!

I would like to suggest that we change the prefix for these method names. I prefer either `prepare_` or `initialize_`. Since the logic is selecting as well as initializing, `prepare_` seems to make more sense. Let me know what you think, or you if you have any other suggestions.

Oblomov:
I don't have a particularly strong feeling about the name, but I would avoid `initialize_` because I feel that would make more sense if we were help in constructing some internal variable with the corresponding name. `prepare`_ OTOH, does make sense, especially for the formatter that _does_ set some formatting-related variables. I'll go with that.

mojavelinux:
Excellent!

mojavelinux:
As soon as I wrote that, I though to myself "what about `create_`"?

I'm trying to find a name that is going to be the most intuitive. I've definitely seen "create" used frequently for this scenario. And it does create a new instance (unless the implementation caches the instance).

Oblomov:
I can see the rationale for `create_`, but I think `prepare_` is better, especially for the formatter, considering that the method does some side stuff (e.g. set `@style` and `@requires_stylesheet`), and can lead to the creation of a chain of formatters. However, as I said, I do not a particularly strong opinion, so I can change it to create_ if you deem it more appropriate. I'm guessing you might still want some time to think about it? 8-)

mojavelinux:
Thanks for emphasizing that distinction. I think you've made a good case for keeping prepare_ as the prefix.

I'm going to give the code one more thorough review to *prepare* (hahaha) for merging.

mojavelinux:
Aha. Now I see where things are getting mixed up. The following two lines should be moved back to `highlight`:

```ruby
@style ||= (style = opts[:style]) && (style_available? style) || DEFAULT_STYLE
```

and

```ruby
@requires_stylesheet = true if opts[:css_mode] == :class
```

That way, the methods that create the lexer and formatter don't have side effects.

And to avoid the dependency on `@style`, it could be amended to the opts:

```ruby
formatter = prepare_formatter node, source, lang, (opts.merge :style, @style)
```

So it would look like this:

```ruby
def highlight node, source, lang, opts
  @style ||= (style = opts[:style]) && (style_available? style) || DEFAULT_STYLE
  @requires_stylesheet = true if opts[:css_mode] == :class
  lexer = create_lexer node, source, lang, opts
  formatter = create_formatter node, source, lang, (opts.merge :style, @style)
  ...
end
```

That feels the most natural to me. (If you think it's okay to depend on `@style`, feel free to make that case).

Could you also move these new methods to just above the line `module Loader`? Since they are auxiliary methods, they should be below the methods from the interface.

And one other small nitpick. I generally like to avoid the use of "return", so the `highlight` method should end like this:

```ruby
if opts[:number_lines] && opts[:callouts]
  [(highlighted = formatter.format lexer.lex source), (idx = highlighted.index CodeCellStartTagCs) ? idx + CodeCellStartTagCs.length : nil]
else
  formatter.format lexer.lex source
end
```

(be careful to only use 2 spaces for indentation).

Oblomov:
Good point about the management of the inner variables, I'll go with your proposed change.

For the location, my aim was to minimize the patch, so it made sense to keep the code blocks exactly where they were. This is the same reason for the presence of the `return` (I forgot to shift it back by one though).

While we're at it, I'm actually thinking about making a single call to the `formatter.format lexer.lex source` that is used in both sides of the branch.

Oblomov:
I've opted to keep the use of `@style` in `create_formatter`, since it's used in multiple places and it allows for lighter syntax. It would also allow users to override `@style` in a derived class without having to remember to replicate the info the in the `opts` array.

mojavelinux:
This looks great to me. Thanks again!

Oblomov:
Thank you for the review, and glad to have given my 2 thousandths to the quality of Asciidoctor 8-)

Issue #1720, opened 2016-04-05
Introduce a header insertion point for docinfo


mojavelinux:
Introduce a new docinfo location, header, which expands docinfo content above other content. In the HTML backend, this would go directly under the document title. In the DocBook backend, this would go above other content.

Look for a docinfo file with the prefix docinfo-header to expand to this location.

mojavelinux:
Note that it's possible to implement this today by feeding a custom document template to the template converter.

mojavelinux:
Open question: In the HTML backend, should the content go inside the `<div id="header">` element, and if so, should it be inserted as the first child or the last child?

lodestone:
What else goes in the existing `<div id="header>` section?

mojavelinux:
The document title followed by the author and revision info. Here's a snippet:

```html
<div id="header">
<h1>Document Title</h1>
<div class="details">
<span id="author" class="author">Author Name</span><br>
<span id="email" class="email"><a href="mailto:author@example.org">author@example.org</a></span><br>
<span id="revnumber">version 1.0.0,</span>
<span id="revdate">2016-04-05</span>
</div>
</div>
```

lodestone:
Ok, I just tried to educate myself on docinfo as much as possible. I'm not sure I succeeded very well, but I wondered the following:

* What advantages does including it in the `#header` section have? Is it logically associated with that content or simply associated by name space (aka "header").

mojavelinux:
I'll turn the question around. Where would you expect the docinfo header content to go.

The docinfo footer goes below the `<div id="footer">`. That tells us that the docinfo header should go below (or above) the `<div id="header">` to be consistent.

lodestone:
From your question-turning-around, I would say ***after*** `<div id="header">`.

mojavelinux:
Roger that! We'll move forward with after unless we hear counter arguments.

ulehnerhs:
I would prefer that content right _before _ `<div id="header">`. 

In my specific case I would need it to add a custom header with a logo, some navigation and last release date right at the top of the documentation. So it should appear right before the document title.

Adding the custom content after the document title wouldn't make sense in my eyes. When generating a pdf or other paper format document, I would expect a customized and corporate designed top  before the title and the toc, preamble and all that stuff right after the title. 
[...]

PR #3954, opened 2021-02-25
Add documentation for docinfo header

mrotteveel:
In #1720 a docinfo header was introduced, but this was not documented in the user manual.

Replaces https://github.com/asciidoctor/asciidoctor.org/pull/942


mojavelinux:
Thanks!

PR #3956, opened 2021-02-27
Docs: add a pandoc command using docker and update command.

dacog:
- Update pandoc command by removing deprecated --atx-headers. Use --markdown-headings=atx instead.
- Add pandoc command with docker to be able to use pandoc without installing it.
- Update pandoc command in general to work with the latest version as in https://pandoc.org/getting-started.html#step-6-converting-a-file

### DIFF
--- a/docs/modules/migrate/pages/ms-word.adoc
+++ b/docs/modules/migrate/pages/ms-word.adoc
@@ -16,10 +16,19 @@ However, in this case, it's a good choice.
 
 To perform the conversion from MS Word docx to AsciiDoc, you need to perform the following command:
 
+.pandoc < 2.11.2
+ 
+ $ pandoc --from=docx --to=asciidoc --wrap=none --atx-headers \
+     --extract-media=extracted-media input.docx > output.adoc
+ 
+.pandoc (2.11.2 or newer)
+ 
  $ pandoc input.docx -f docx -t asciidoc --wrap=none --markdown-headings=atx \                                                                             
     --extract-media=extracted-media  -o output2.adoc
-     
-If you use docker you can also use pandoc with docker without installing it:
+
+.pandoc (2.11.2 or newer) and docker
+
+If you use docker you can also use the latest version of pandoc with docker without installing it:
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks for proposing updates to this page.

> Update pandoc command by removing deprecated --atx-headers. Use --markdown-headings=atx instead.

This seems odd to me. We aren't making Markdown. We are making AsciiDoc. So why would the name of the flag be specific to Markdown?

Also, I'm using the latest version of pandoc available on Fedora (pandoc 2.9.2.1) and I don't see this new flag.

```
pandoc --help| grep atx
                        --atx-headers
```

Is this perhaps an unreleased deprecation? Suggesting it prematurely could make this command fail for readers trying to following along.

...on closer examination, it was added in the pandoc 2.11.2 release, which only happened in November. It's too soon to suggest this change. We need to wait until the version in Linux distributions is at least 2.11.2 (and that includes Debian).

dacog:
Hi @mojavelinux ,

Thank you taking the time to review the changes. 

I understand what you mean.  

Would it be ok then if I add the new syntax with a note that it is for a new version (as well as the docker one which already uses the new version) and leave the old syntax there as first command?

Although I understand what you mean, I think that just leaving the command for the old version can also confuse the users, as they might have installed the newer version on a non-Linux environment (as it was my case using docker on macOS). 

I can update my merge request as proposed if you think this would be better. If you prefer it another way, please just let me know 😄 .

Thank you,

Best regards,
Diego

mojavelinux:
> Would it be ok then if I add the new syntax with a note that it is for a new version (as well as the docker one which already uses the new version) and leave the old syntax there as first command?

Absolutely. You can leave the first command, then offer the second command for those who have upgraded to the new version. You might even format it like:

```
.pandoc < 2.11.2
$ old command

.pandoc (2.11.2 or newer)
$ new commmand
```

dacog:
@mojavelinux I have updated the changes as per your feedback. Please let me know if that would be ok.

mojavelinux:
This looks great. Thanks!

PR #3963, opened 2021-03-06
Fixed code quality issues

withshubh:
### Description
Hi :wave: 
I ran the [DeepSource](https://deepsource.io) analysis on the forked copy of this repo and found some interesting [code quality issues](https://deepsource.io/gh/withshubh/asciidoctor/issues/?category=recommended).

#### Summary of changes

- For first assignment, use `=` instead of `||=`
- Prefer rescuing `StandardError` over `Exception`
- Remove useless access modifiers
- Fix improper use of block delimiter
- Remove redundant string coercion
- Remove empty `else` clauses
- Remove bad use of `alias`
- Add .deepsource.toml

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,13 @@
+version = 1
+
+test_patterns = ["test/**/*_test.rb"]
+
+exclude_patterns = [
+    "bin/**",
+    "docs/**",
+    "man/**"
+]
+
+[[analyzers]]
+name = "ruby"
+enabled = true
--- a/benchmark/benchmark.rb
+++ b/benchmark/benchmark.rb
@@ -76,13 +76,13 @@ when 'userguide'
   sample_file = ENV['BENCH_TEST_FILE'] || 'sample-data/userguide.adoc'
   backend = ENV['BENCH_BACKEND'] || 'html5'
   fetch_userguide if sample_file == 'sample-data/userguide.adoc' && !(File.exist? sample_file)
-  result = Benchmark.bmbm {|bm|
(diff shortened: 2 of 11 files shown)


mojavelinux:
I appreciate you looking out for the code quality. However, as you can see, the tool didn't get everything right. I'm generally pretty skeptical of tools that apply arbitrary rules to a code base without understanding what that code does. That's not to say I don't think they have their place, but it takes time and configuration to get to a point where we can trust the tool.

A tool I already have a lot of experience with is RuboCop. I would prefer to use that on Asciidoctor. In fact, that is the plan for >= 2.1.

Let's look at just a few of the mistakes this tool made:

* ::Exception ⇒ StandardError - first, this removes the leading `::`, which I prefer to have there to make it clear it's a qualified type; second, it integration with AsciidoctorJ, which can throw exceptions which are not caught by StandardError
* @resolving_xref ||= to @resolving_xref = - this breaks the recursive logic, which needs to keep track of when this variable has already been already set by the outer call
* alias ⇒ alias_method - these are two different language features and they are not interchangeable; we rely on alias so that the mapping is available to the interpreter and can be replaced by later declarations
* removing to_s in interpolated strings - while objects are automatically coerced in interpolated strings, it does not communicate the intent; I prefer the explicit call to .to_s
* removing the private keyword - it's there for a reason, to mark private access; if we remove it, then what it was marking is no longer private

In this entire PR, I only see one change I would want to apply, which is to remove an explicit nil return inside an empty else block. But that's more of style issue than a code quality issue. So I'm going to decline this PR for something that is more carefully thought out.

withshubh:
Thank You for reviewing @mojavelinux :sparkles:

Issue #3967, opened 2021-03-14
Use semantic constructs for the author(s) and revision


Mogztter:
This issue will be addressed as part of: https://github.com/asciidoctor/asciidoctor/issues/242

### Author(s)

When there are a more than one author, we should probably use an unordered list:

```html
<ul class="byline">
  <li><!-- author 1 --></li>
  <li><!-- author 2 --></li>
</ul>
```

When there's only a single author, we could probably use a `<p>` element:

```html
<p class="byline">
  <!-- author -->
</p>
```

If we want to use same structure regardless of the number of authors, then we could use a generic container like `<div class="byline">`?

In the above examples, I'm using the class `byline` because it seems to be a well-established term: https://en.wikipedia.org/wiki/Byline
For reference, the Asciidoctor EPUB3 is using it: https://github.com/asciidoctor/asciidoctor-epub3/blob/3bfee503ee959119c1fe0097b1346fe49ff7fdea/lib/asciidoctor-epub3/converter.rb#L356

An author consists of a name and an optional email.
Using a `<span>` element for the author's name and a `<a>` element for the author's email seem sufficient:

```html
<span class="author">John Doe</span> <a rel="author" href="mailto:john@writer.org">john@writer.org</a>
```

Here are complete examples:

```html
<ul class="byline">
  <li><span class="author">Kismet Lee</span></li>
  <li><span class="author">John Doe</span> <a rel="author" href="mailto:john@writer.org">john@writer.org</a></li>
</ul>
```

```html
<p class="byline">
  <span class="author">John Doe</span> <a rel="author" href="mailto:john@writer.org">john@writer.org</a>
</p>
```

#### Email rel="author"

MDN suggests that we should add the link type "author" (`rel="author"`) on the author's email:

```html
<a rel="author" href="mailto:john@writer.org">john@writer.org</a>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

In addition, some resources claim that contact information should be enclosed in a `<address>` element:

```html
<address>
  <span class="author">John Doe</span><a rel="author" href="mailto:john@writer.org">john@writer.org</a>
</address>
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

Since the email address is optional, I'm not convinded that this is actually a good idea to enclose the author in an `<address>` element when the email address is present.

#### Author format

I think, it might be a good idea to have a dedicated method to format an author.
[...]

Issue #2749, opened 2018-05-02
Distinguish subtitle from primary level-0 title in html5 output


christofferholmstedt:
I'm new to Asciidoctor and currently trying it out by writing a few book reviews. I'm writing in vim and using Firefox plugin Ascidoctor.js Live Preview, asciidoctor from command line to generate html5 as well as asciidoctor-pdf to get a PDF.

As noted in the documentation; subtitles are "not distinguished from the main title in the html5 output" [1]. **How do I achieve this?** (the output in PDF looks good).

### Found discussions about this
In issue asciidoctor/asciidoctor.org#297 it is mentioned that there should be no big blocker to implement this separation for HTML5 backend. **Is this bug report in the correct repository for such a functionality and how can I help?**

In the same issue it is also mentioned that a "slim template" could do the job. As I'm new to Asciidoctor I have no clue what a slim template is. Searching through the user-manual [2] I get no hits on "slim template" so I must be searching for the wrong thing. Any pointers on how to get started with "slim templates"?

[1] https://asciidoctor.org/docs/user-manual/#document-subtitle
[2] https://asciidoctor.org/docs/user-manual/

znerd:
This appears to be related to [asciidoctor.org issue 423](https://github.com/asciidoctor/asciidoctor.org/issues/423).

@mojavelinux: Is this ticket appropriate for tracking the progress of `subtitle` support on the HTML5 backend, or do we need a separate “clean” ticket for that?

Either way, here are some relevant questions:

1. Is it possible to implement this without changes to the core?
2. If so, then can it be done only with a custom Slim template, or does it also require custom parsing?
3. If so, then is there a (reusable) example somewhere?
4. What should be the styling in the default template?

mojavelinux:
This is another great example of something to be included in #242. (Though it doesn't have to wait for that, but it does make the most sense).

It's certainly possible to implement this without a change to core. Remember that all the HTML that is generated can be customized using templates or a custom converter.

You can find info about how to make custom templates here: https://github.com/asciidoctor/asciidoctor.org/issues/80

In this case, you would need to override the document template since that's where the header is defined. You can use this file as a starting point: https://github.com/asciidoctor/asciidoctor-backends/blob/master/slim/html5/document.html.slim

Another possible way to do it is to use a postprocessor to patch the generated HTML. It's not as clean to do it that way, but certainly less work.

nkalvi:
[...]

PR #3979, opened 2021-03-24
ref #2749 distinguish title and subtitle

Mogztter:
### What I did

* Introduce a method to generate the header
* Introduce a method to generate the document title (title + subtitle)
* Generate a *simplified header* that includes the document title + subtitle on the embedded output

### Limitations

The implementation is *deliberately* incomplete to make it easier to review.
A more complete header will be implemented as part of https://github.com/asciidoctor/asciidoctor/issues/3967

#### generate methods

Currently, `generate_*` methods can return `nil`.
More specifically, `generate_header` returns `nil` when `noheader` attribute is defined or when there's no header block. Similarly, `generate_document_title` returns `nil` when `notitle` attribute is defined.

Should we move conditions outside of these methods?

```rb
def generate_document_title node
  doctitle = node.doctitle partition: true, sanitize: true
  attributes = common_html_attributes node.id, node.role
  %(<h1#{attributes}>#{doctitle.main}#{doctitle.subtitle? ? %( <small class="subtitle">#{doctitle.subtitle}</small>) : ''}</h1>)
end
```

If not, should we return empty instead of `nil`? My understanding is that returning `nil` is (somehow) more idiomatic Ruby?

#### Roles and id

It's possible to assign and id (and roles!) to a document title.

```
[#start.dummy]
= Document Title
```

Currently, the converter will produce:

```html
<header>
<h1 id="start" class="dummy">Document Title</h1>
</header>
```

I wonder if we should assign roles to the parent element (i.e., in this case the `<header>` element)? And, more broadly, should we define a rule to make it consistent and predictable? 

ref #242

### DIFF
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -12,7 +12,12 @@ class Converter::SemanticHtml5Converter < Converter::Base
   end
 
   def convert_embedded node
-    node.content
+    result = []
+    if (header = generate_header node)
+      result << header
+    end
+    result << node.content
+    result.join LF
   end
 
   def convert_paragraph node
@@ -26,6 +31,25 @@ class Converter::SemanticHtml5Converter < Converter::Base
       %(<p#{attributes}>
 #{node.content}
 </p>)
+    end
+  end
+
+  def generate_header node
+    if node.header? && !node.noheader
+      result = ['<header>']
+      if (doctitle = generate_document_title node)
+        result << doctitle
+      end
+      result << '</header>'
+      result.join LF
+    end
+  end
+
+  def generate_document_title node
+    unless node.notitle
+      doctitle = node.doctitle partition: true, sanitize: true
+      attributes = common_html_attributes node.id, node.role
+      %(<h1#{attributes}>#{doctitle.main}#{doctitle.subtitle? ? %( <small class="subtitle">#{doctitle.subtitle}</small>) : ''}</h1>)
     end
   end
 
--- a/test/fixtures/semantic-html5-scenarios/document-title-subtitle.adoc
+++ b/test/fixtures/semantic-html5-scenarios/document-title-subtitle.adoc
@@ -1,0 +1,2 @@
+= Document Title: Subtitle
+:showtitle:
--- a/test/fixtures/semantic-html5-scenarios/document-title-subtitle.html
+++ b/test/fixtures/semantic-html5-scenarios/document-title-subtitle.html
@@ -1,0 +1,3 @@
+<header>
+<h1>Document Title <small class="subtitle">Subtitle</small></h1>
+</header>
--- a/test/fixtures/semantic-html5-scenarios/document-title-with-id-roles.adoc
+++ b/test/fixtures/semantic-html5-scenarios/document-title-with-id-roles.adoc
@@ -1,0 +1,3 @@
+[#start.dummy]
+= Document Title
+:showtitle:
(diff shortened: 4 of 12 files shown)


lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -29,6 +34,25 @@ def convert_paragraph node
     end
   end
 
+  def generate_header node
+    if node.header? && !node.noheader
```
Mogztter, line 38: All combinations are covered:

| `header?` | `noheader` | `notitle` | Expected Result |
|-|-|-|-|
| `true` | `true` |  `true` | No `<header>` |
| `true` | `true` |  `false` |  No `<header>` |
| `true` | `false` |  `false` |  Complete `<header>` |
| `true` | `false` |  `true` |  `<header>` without title |
| `false` | `true` |  `true` |  No `<header>` |
| `false` | `true` |  `false` |  No `<header>` |
| `false` | `false` |  `true` |  No `<header>` |
| `false` | `false` |  `false` |  No `<header>` |

Mogztter:
According to https://developers.google.com/style/headings-targets we should use the following to make headings into link targets:

```html
<section id="introduction-to-everything">
  <h2>Introduction to everything</h2>
</section>
```

So we should probably move the role and id to the parent element:

```html
<header id="start" class="dummy">
  <h1>Document Title</h1>
</header>
```

@mojavelinux Thoughts?

mojavelinux:
> we should probably move the role and id to the parent element:

```
<header id="start" class="dummy">
  <h1>Document Title</h1>
</header>
```

This is tricky. You'll always have one camp who thinks it should be on the wrapper, and one that thinks it should be on the heading itself. I tend to be in the first camp. But there is a compromise we could consider, which is to have a system where we repeat the id and class on the heading with a prefix. But we won't truly know if we need that until we start developing the CSS for this HTML...so it might be best to wait and see.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: I'll be honest, this whole notitle thing has always been really weird and ill-defined. And it clashes with noheader. The idea of it, originally, was to disable the display of the h1 title and only show the body of the document (almost always for embedded output). But that was before we had a well-defined header. And in PDF, it's meaning is to disable the title page, which is really the purpose the header is serving in HTML. I'm sort of leaning towards treating it like noheader, which is to just disable the `<header>` (or ignoring it).

mojavelinux (approved):
If you'd like, we can revisit the interpretation of notitle in a separate issue. As for this partial header, I'm happy with what is there thus far.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: In HTML, it can make sense to hide the document title but still show document metadata (i.e., revision and author(s)). In PDF, not so much...

> And in PDF, it's meaning is to disable the title page, which is really the purpose the header is serving in HTML.

We also have the attribute `title-page` which serves the same purpose no?

Anyway, `notitle` is probably too fine grained and as you've mentioned it clashes with `noheader`. Also, if you want to hide the document title in HTML, it requires 1 line of CSS:

```css
header > h1 { display: none; }
```

> I'm sort of leaning towards treating it like noheader, which is to just disable the \<header> (or ignoring it).

When you say disable/ignore the `<header>`, you mean do no generate/produce the `<header>` element?

Mogztter:
> If you'd like, we can revisit the interpretation of notitle in a separate issue. As for this partial header, I'm happy with what is there thus far.

Follow-up issue created!

Thanks for your review, it's greatly appreciated. Even if I can merge the pull request after a few days, it's comforting to know that someone is watching (and that I'm actually going in the right direction 😃)

PR #4015, opened 2021-04-14
resolves #4013 increment counter even if attribute is locked

### LINKED ISSUE #4013: Rework counter logic to use and increment counter even if attribute is locked
If the attribute used for the counter is locked, the counter itself should still be incremented. In other words, the locked attribute can still be used to control the start value of the counter, but the attribute isn't subsequently updated.

### LINKED ISSUE #4014: Use attribute as seed value for counter even if not already registered as a counter
If the counter directive references an existing attribute, it should use that attribute as the seed value even if the attribute isn't already configured as a counter or a seed value is passed to the counter method.

Currently, a seed value is passed in for chapter-number and appendix-number. As a result, the start value of the counter cannot be controlled via the attribute.

mojavelinux:
resolves #4013 increment counter (but not the corresponding attribute) if attribute is locked

also resolves #4014 use attribute as seed value for counter even if not already registered as a counter

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,6 +13,13 @@ endif::[]
 This document provides a high-level view of the changes introduced in Asciidoctor by release.
 For a detailed view of what has changed, refer to the {uri-repo}/commits/master[commit history] on GitHub.
 
+== Unreleased
+
+Improvments::
+
+ * Increment counter (but not the corresponding attribute) if attribute is locked (#4013)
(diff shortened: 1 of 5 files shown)


mojavelinux:
This still solves #3939, but doesn't disable the functionality of the counter in the process. It also allows the chapter and appendix number to be modified using the `chapter-number` and `appendix-number` attributes without any special hacks.

PR #4021, opened 2021-04-16
ref #3967 generate author(s)

Mogztter:
This is the first part of #3967 where we generate semantic constructs for the author(s).

### DIFF
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -87,14 +87,14 @@ class Converter::SemanticHtml5Converter < Converter::Base
   end
 
   def format_author node, author
-    with_context 'author' do
+    in_context 'author' do
       %(<span class="author">#{node.sub_replacements author.name}#{author.email ? %( #{node.sub_macros author.email}) : ''}</span>)
     end
   end
 
-  def with_context name, &block
+  def in_context name
     (@convert_context ||= []).push name
-    result = block.call
+    result = yield
     @convert_context.pop
     result
   end
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -74,21 +74,21 @@ class Converter::SemanticHtml5Converter < Converter::Base
 
     if node.authors.length == 1
       %(<p class="byline">
-#{format_author node, node.authors.first, 1}
+#{format_author node, node.authors.first}
 </p>)
     else
       result = ['<ul class="byline">']
-      node.authors.each_with_index do |author, index|
-        result << "<li>#{format_author node, author, index + 1}</li>"
+      node.authors.each do |author|
+        result << "<li>#{format_author node, author}</li>"
       end
       result << '</ul>'
       result.join LF
     end
   end
 
-  def format_author node, author, index
+  def format_author node, author
     with_context 'author' do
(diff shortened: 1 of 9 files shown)


lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -34,12 +34,40 @@ def convert_paragraph node
     end
   end
 
+  def convert_inline_anchor node
```
Mogztter, line 37: Taken from the `Html5Converter`.
The implementation is incomplete because it's not the focus of this pull request.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -53,10 +69,52 @@ def generate_document_title node
...
+        result << "<li>#{format_author node, author, index + 1}</li>"
+      end
+      result << '</ul>'
+      result.join LF
+    end
+  end
+
+  def format_author node, author, index
+    with_context 'author' do
+      %(<span class="author" data-index="#{index}">#{node.sub_replacements author.name}#{author.email ? %( #{node.sub_macros author.email}) : ''}</span>)
```
Mogztter, line 91: Not sure if the `data-index` is really useful... I guess it's relatively to compute the index in JavaScript and in CSS we can use [:nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) to select an element based on its position in a group of siblings.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -53,10 +69,52 @@ def generate_document_title node
...
+    end
+  end
+
+  def with_context name, &block
+    (@convert_context ||= []).push name
+    result = block.call
+    @convert_context.pop
+    result
+  end
```
Mogztter, line 100: Utility method to make the code a bit more readable.
I still don't know if it's better to use `&block` + `block.call`, implicit + `yield`, `Proc`, `lambda`... ?

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -53,10 +69,52 @@ def generate_document_title node
...
+
+  def format_author node, author, index
+    with_context 'author' do
+      %(<span class="author" data-index="#{index}">#{node.sub_replacements author.name}#{author.email ? %( #{node.sub_macros author.email}) : ''}</span>)
+    end
+  end
+
+  def with_context name, &block
+    (@convert_context ||= []).push name
```
Mogztter, line 96: I'm using an array in case we need nested context. Arguably, we could use a string...?

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: I agree. This seems redundant.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: You should always use yield unless the situation calls for something different. For example, if you need to bind the block to another context, then you need something like instance_exec + block. But that's rare. I use blocks a lot in Asciidoctor PDF, so you can use that as a reference.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: The array seems reasonable here. I may rename this to in_context rather than with_context. You are not adding more context (variables) to the scope, which is usually what with_context means. in_context implies that you are inside some sort of context (but not contributing to it), which is more appropriate.

mojavelinux (approved):

PR #4022, opened 2021-04-16
ref #3967 generate revision

Mogztter:
This is the second part of #3967 where we generate revision.

### DIFF
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -118,9 +118,9 @@ class Converter::SemanticHtml5Converter < Converter::Base
 </thead>
 <tbody>
 <tr>
-<td data-title="#{node.attr 'version-label', ''}">#{node.attr 'revnumber', ''}</td>
+<td data-title="#{node.attr 'version-label'}">#{node.attr 'revnumber'}</td>
 <td data-title="Date">#{revision_date}</td>
-<td data-title="Remark">#{node.attr 'revremark', ''}</td>
+<td data-title="Remark">#{node.attr 'revremark'}</td>
 </tr>
 </tbody>
 </table>)
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+autoload :Date, 'date' unless RUBY_ENGINE == 'opal'
+
 module Asciidoctor
 # A built-in {Converter} implementation that generates HTML 5 output
 # that maximizes the use of semantic constructs.
@@ -56,6 +58,9 @@ class Converter::SemanticHtml5Converter < Converter::Base
       if (authors = generate_authors node)
         result << authors
       end
+      if (revision = generate_revision node)
+        result << revision
+      end
       result << '</header>'
       result.join LF
     end
@@ -84,6 +89,35 @@ class Converter::SemanticHtml5Converter < Converter::Base
       result << '</ul>'
       result.join LF
     end
+  end
+
+  def generate_revision node
+    return unless (node.attr? 'revnumber') || (node.attr? 'revdate') || (node.attr? 'revremark')
+
+    revision_date = if (revdate = node.attr 'revdate')
+      date = ::Date._parse revdate
+      date_text = "#{date[:year]}-#{date[:mon].to_s.rjust 2, '0'}-#{date[:mday].to_s.rjust 2, '0'}"
+
+      %(<time datetime="#{date_text}">#{revdate}</time>)
+    else
+      ''
+    end
+    %(<table class="revision">
+<thead>
+<tr>
+<th>Version</th>
+<th>Date</th>
+<th>Remark</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td data-title="#{node.attr 'version-label', ''}">#{node.attr 'revnumber', ''}</td>
+<td data-title="Date">#{revision_date}</td>
+<td data-title="Remark">#{node.attr 'revremark', ''}</td>
+</tr>
+</tbody>
+</table>)
   end
 
   def format_author node, author
--- a/lib/asciidoctor/converter/semantic-html5.rb
+++ b/lib/asciidoctor/converter/semantic-html5.rb
@@ -96,9 +96,15 @@ class Converter::SemanticHtml5Converter < Converter::Base
 
     revision_date = if (revdate = node.attr 'revdate')
       date = ::Date._parse revdate
-      date_text = "#{date[:year]}-#{date[:mon].to_s.rjust 2, '0'}-#{date[:mday].to_s.rjust 2, '0'}"
-
-      %(<time datetime="#{date_text}">#{revdate}</time>)
+      if (date.has_key? :year) || (date.has_key? :mon) || (date.has_key? :mday)
+        date_parts = []
+        date_parts << "#{date[:year]}" if date.has_key? :year
+        date_parts << "#{date[:mon].to_s.rjust 2, '0'}" if date.has_key? :mon
+        date_parts << "#{date[:mday].to_s.rjust 2, '0'}" if date.has_key? :mday
+        %(<time datetime="#{date_parts.join '-'}">#{revdate}</time>)
+      else
+        revdate
+      end
     else
       ''
     end
--- a/test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
@@ -1,4 +1,4 @@
 = Document Title
 Author Name
 :revnumber: LPR55
-:version-label!:
+:!version-label:
--- a/test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
@@ -1,0 +1,4 @@
+= Document Title
+Author Name
+:revnumber: LPR55
+:version-label!:
--- a/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd-yyyy.adoc
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd-yyyy.adoc
@@ -1,0 +1,4 @@
+= Document Title
+Author Name
+:revdate: October 31, 2021
+:revnumber: 1.0
--- a/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd-yyyy.html
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd-yyyy.html
@@ -1,0 +1,21 @@
+<header>
+<p class="byline">
+<span class="author">Author Name</span>
+</p>
+<table class="revision">
+<thead>
+<tr>
+<th>Version</th>
+<th>Date</th>
+<th>Remark</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td data-title="Version">1.0</td>
+<td data-title="Date"><time datetime="2021-10-31">October 31, 2021</time></td>
+<td data-title="Remark"></td>
+</tr>
+</tbody>
+</table>
+</header>
--- a/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd.adoc
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd.adoc
@@ -1,0 +1,4 @@
+= Document Title
+Author Name
+:revdate: January, 4th
+:revnumber: 1.0
--- a/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd.html
+++ b/test/fixtures/semantic-html5-scenarios/revision-with-date-MMMM-dd.html
@@ -1,0 +1,21 @@
+<header>
+<p class="byline">
+<span class="author">Author Name</span>
+</p>
+<table class="revision">
+<thead>
+<tr>
+<th>Version</th>
+<th>Date</th>
+<th>Remark</th>
+</tr>
+</thead>
+<tbody>
+<tr>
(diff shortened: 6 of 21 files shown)


lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -40,11 +42,47 @@ def generate_header node
...
+      header << '<th>Date</th>'
+      date = ::Date._parse revdate
+      date_text = "#{date[:year]}-#{date[:mon].to_s.rjust 2, '0'}-#{date[:mday].to_s.rjust 2, '0'}"
+      body << %(<td data-title="Date"><time datetime="#{date_text}">#{revdate}</time></td>)
+    end
+    if (revremark = node.attr 'revremark')
+      header << '<th>Remark</th>' if node.attr? 'revremark'
+      body << %(<td data-title="Remark">#{revremark}</td>)
+    end
+    %(<table class="revision">
```
Mogztter, line 72: I quite like the table structure.
One drawback is that the number of columns is flexible so we cannot use the column index to apply a specific style. In other words, we don't know for sure what the first column contains.
In this context, it could be useful to add a class or a data attribute to identify columns.

Do you think we should always generate all columns (i.e., use a fixed number of columns)?

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
+require 'date'
```
mojavelinux, line 2: I tend to prefer autoload. That way, it's loaded on demand.

```
autoload :Date, 'date'
```

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: One thing we lose by using the table is the concise byline we see here: https://beanvalidation.org/2.0/spec/ and here: https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html. Clearly, a revision table is also desired in both of those cases.

We also run up against the case that it's only possible to specify a single revision. Though perhaps this converter is making room for that eventuality. Still, I wonder if the table is something that can be turned on using an attribute like `revtable`, or `:revstyle: table`.

Mogztter: > One thing we lose by using the table is the concise byline we see here: beanvalidation.org/2.0/spec and here: jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html. 

It's possible to display the table as a single line using a few lines of CSS:

```html
<table class="revision">
<thead>
<tr>
<th>Version</th>
<th>Date</th>
<th>Remark</th>
</tr>
</thead>
<tbody>
<tr>
<td data-title="Version">1.2.3</td>
<td data-title="Date"><time datetime="2021-10-31">October 31, 2021</time></td>
<td>Fall incarnation<td>
</tr>
</tbody>
</table>
```

```css
table.revision > thead {
  display: none;
}
table.revision tr > td:first-child::before {
  content: attr(data-title) ' '
}
table.revision tr > td:first-child::after {
  content: ','
}
table.revision tr > td:nth-child(3):not(:empty)::before {
  content: ' |'
}
```

And here's the result:

111903820-4847b400-8a44-11eb-8229-dc5be80635c3


With a few extra lines of CSS, it's also possible to support multiple revisions:


```css
table.revision tbody > tr {
  display: inline-flex;
}
table.revision tbody > tr:not(:first-child)::before {
  content: ' - '
}
```

multi-revs


> We also run up against the case that it's only possible to specify a single revision. Though perhaps this converter is making room for that eventuality. Still, I wonder if the table is something that can be turned on using an attribute like revtable, or :revstyle: table.

That's a good idea.
If you have more than one revision you will probably want to display them as a table.

We could also configure the location of the revision(s) table (similar to what we do with the Table Of Contents):

- `preamble`
- `macro`
- `header` (default) 


```
:revstyle: table
:revision: preamble
:revision-title: Revisions
```

demo-tablestyle-revs

Even if we won't implement that in the near future, it seems flexible enough to accommodate "advanced" features.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: I should probably hide it from Opal using:

```rb
unless RUBY_ENGINE_OPAL
  autoload :Date, 'date'
end
```

Not sure if it's working properly with the following syntax:

```rb
autoload :Date, 'date' unless RUBY_ENGINE_OPAL
```

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: > Do you think we should always generate all columns (i.e., use a fixed number of columns)?

I think we should, it will make things easier and we can use `:not(:empty)` to style accordingly.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: Yes. You can see how I do it here: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L5-L15

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: > > Do you think we should always generate all columns (i.e., use a fixed number of columns)?

> I think we should, it will make things easier and we can use :not(:empty) to style accordingly.

Agreed. Go for it.

mojavelinux: > We could also configure the location of the revision(s) table (similar to what we do with the Table Of Contents):

That's a great idea. If it feels out of scope, I'm fine with making it a separate issue.

Based on what you showed, I'm prepared to move forward with the table as you have it (with the fixed number of columns).

Mogztter: I'm glad you like it!
Yes, let's do that as a follow-up issue 😃

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -86,6 +91,35 @@ def generate_authors node
     end
   end
 
+  def generate_revision node
+    return unless (node.attr? 'revnumber') || (node.attr? 'revdate') || (node.attr? 'revremark')
+
+    revision_date = if (revdate = node.attr 'revdate')
+      date = ::Date._parse revdate
```
mojavelinux, line 98: One minor nitpick. It's not required that the revdate be a valid date. Date._parse already handles an invalid date gracefully. If the hash comes back empty, the datetime attribute should be dropped (or perhaps don't use the time element at all).

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -86,6 +91,35 @@ def generate_authors node
...
+<thead>
+<tr>
+<th>Version</th>
+<th>Date</th>
+<th>Remark</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td data-title="#{node.attr 'version-label', ''}">#{node.attr 'revnumber', ''}</td>
```
mojavelinux, line 115: Perhaps the version-label should control the column header instead of being added per cell. It also seems a bit redundant to add data-title to each cell. wdyt?

mojavelinux (changes requested):

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: I didn't known that it could return an empty hash. It can also return an incomplete hash!

```rb
p Date._parse("2eme janvier de l'an 511") # {:mon=>1, :mday=>2}
p Date._parse("l'an 511") # {:yday=>511}
p Date._parse('cinq cent onze av. J.-C.') # {}
```

> If the hash comes back empty, the datetime attribute should be dropped (or perhaps don't use the time element at all).

Since the `datetime` format is really flexible, we can still generate the datetime attribute if a value is missing.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#valid_datetime_values

```html
<time datetime="01-02">2eme janvier de l'an 511</time>
```
```html
<time datetime="511">l'an 511</time>
```
```html
cinq cent onze av. J.-C.
```

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: I'm using `data-title` to preprend the version label using CSS:

```css
table.revision tr > td:first-child::before {
  content: attr(data-title) ' '
}
```

It would be nice if we could get the value of the header for a given cell in CSS but as far as I know that's not possible 😞 

> It also seems a bit redundant to add data-title to each cell. wdyt?

That's true but I guess that's also true if we want to apply a custom style on the first cell in a table. In this case, we would need to add a class on each cell of every row.

Alternatively, we could generate dynamic CSS:

```ruby
css = %(table.revision tr > td:first-child::before {
  content: #{node.attr 'version-label', ''} ' '
})
```

But we should be cautious because if we do that then we won't have a fully static CSS...

> Perhaps the version-label should control the column header instead of being added per cell.

Do you think we should also use `#{node.attr 'version-label', ''}` in the table header? If so, I believe that "Remark" and "Date" should also be configurable (for i18n).

ytbigvinny288 (approved):

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: Thanks for explaining. Now I understand.

I agree that putting the version label in a data attribute makes styling it more flexible (which is a core goal of this converter). I think I'd rather see it as a span element, though, rather than on the cell.

```
<td><span data-title="#{node.attr 'version-label'}">#{node.attr 'revnumber', ''}</span></td>
```

I don't think the other cells should have data-title attributes since these are already provided by the table header.

> Do you think we should also use #{node.attr 'version-label', ''} in the table header? If so, I believe that "Remark" and "Date" should also be configurable (for i18n).

Perhaps we will want that. But it should be a separate issue since it entails adding new attributes...names which we then have to support.

mojavelinux: Btw, you don't need the fallback string when resolving the version-label attribute:

```
node.attr 'version-label'
```

That's only needed if you want to call a method on the result (though even then, Ruby 2.5 has a safe navigation operator).

test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
```diff
@@ -0,0 +1,4 @@
+= Document Title
+Author Name
+:revnumber: LPR55
+:version-label!:
```
mojavelinux, line 4: The preferred syntax for negating an attribute is to put the `!` before the name.

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: > Btw, you don't need the fallback string when resolving the version-label attribute:

I didn't know that the string representation of `nil` is `''`, I will remove it 😄 

> Perhaps we will want that. But it should be a separate issue since it entails adding new attributes...names which we then have to support.

👍🏻 

> I agree that putting the version label in a data attribute makes styling it more flexible (which is a core goal of this converter). I think I'd rather see it as a span element, though, rather than on the cell.

Any rationale behind this decision?

test/fixtures/semantic-html5-scenarios/revision-with-version-without-label.adoc
Mogztter: 🙃

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: > I don't think the other cells should have data-title attributes since these are already provided by the table header.

Unless, you are hiding the table header.
It gives the opportunity to tweak the design:

```css
thead {
  display: none;
}
tbody > tr {
  display: flex;
  flex-direction: column;
}
tbody > tr > td {
  display: flex;
  gap: 5px;
}
tbody > tr > td:nth-child(1)::before,
tbody > tr > td:nth-child(2)::before {
  content: attr(data-title) ' ';
  display: block;
  font-weight: bold;
}
```

image

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: I'm fine with a partial date. But if the Hash it empty, I think it should be dropped.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: > Any rationale behind this decision?

Because "before" on a table cell fundamentally changes the DOM structure of the HTML. It essentially adds another element next to the table cell, which is really not semantic at all.

> Unless, you are hiding the table header.
It gives the opportunity to tweak the design:

That seems like some pretty drastic CSS hacking. I would think it would be better in that case to extend the converter. I feel like we are pushing the semantics too far here.

Mogztter: > I feel like we are pushing the semantics too far here.

The `data-title` attribute is all about CSS styling. Semantically, revisions are well structured in a table.

> Because "before" on a table cell fundamentally changes the DOM structure of the HTML. It essentially adds another element next to the table cell, which is really not semantic at all.

You mean inside the table cell?
I'm struggling to understand what value provides the `<span>` element:

##### before

```html
<td>
  ::before
  revnumber
</td>
```

##### after

```html
<td>
  <span>
    ::before
    revnumber
  </span>
</td>
```

On one hand, we add a pseudo-element in a `<td>` element and on the other hand we add a pseudo-element in a `<span>` element (inside a `<td>` element).

I'm voting in favor of leaving this for later and removing the `data-title` attribute for now.
We are making adjustments and speculating on how to style a component rather than focusing on establishing a well structured HTML.

I think we've agreed that the revision

mojavelinux: I did some more testing and realized I was incorrect. The pseudo-element does go *inside* the table cell. So it's fine to put data-title on the td itself.

I see your point about adding data-title to all the cells to accommodate advanced layouts. So I'm willing to go with and see how it plays out.

mojavelinux (approved):

PR #4036, opened 2021-04-22
resolves #4035 don't assign empty role to inline phrase

### LINKED ISSUE #4035: Don't assign empty role to inline phrase
As a result of the optimization applied in 2b93ed1178544afe03230c7aa42f0c6239e567d5, it became possible for the parser to assign an empty role to an inline phrase. If the role is empty (after removing leading and trailing spaces), it should not be assigned to the phrase.

Consider this example:

```
[,]#text#
```

A role should not be assigned to the inline phrase in this case.

mojavelinux:


ytbigvinny288 (commented):
.

ytbigvinny288 (approved):

Issue #3981, opened 2021-03-27
Use semantic constructs for sections


Mogztter:
This issue will be addressed as part of: https://github.com/asciidoctor/asciidoctor/issues/242

### Sections

I think we should use [`<section>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) as a parent element.
I can't think of any other element that would make more sense.

```html
<section>
  <!-- ... -->
</section>
```

#### Headings

Here's MDN usage recommandation:

> To reiterate, each `<section>` should be identified, typically by including a heading (`<h1>`-`<h6>` element) as a child of the `<section>` element, wherever possible.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section#usage_notes

This recommandation fits quite well with the AsciiDoc syntax:

```
== Section Title

This is a paragraph.
```

```html
<section>
  <h2>Section Title</h2>
  <p>This is a paragraph.</p>
</section>
```

#### Wrapping content

I don't think we need to wrap the content of a section inside a generic container such as the `<div>` element:

```html
<section>
  <h2>Section Title</h2>
  <p>First paragraph.</p>
  <p>Second paragraph.</p>
</section>
```

#### Nesting sections

I did some research and I think it's fine to nest section elements:

```html
<section>
  <h2>Section Title</h2>
  <section>
    <h3>Subsection Title</h3>
  </section>
</section>
```

##### Using CSS

```css
body {
  counter-reset : h2;
}

h2 {
  counter-reset : h3;
}

h3 {
  counter-reset : h4;
}

h4 {
  counter-reset : h5;
}

h5 {
  counter-reset : h6;
}

h2:before {
  content : counter(h2,decimal) ".\0000a0";
  counter-increment : h2;
}

h3:before {
  content : counter(h2,decimal) "." counter(h3,decimal) ".\0000a0";
  counter-increment : h3;
}

h4:before {
  content : counter(h2,decimal) "." counter(h3,decimal) "." counter(h4,decimal) ".\0000a0";
  counter-increment : h4;
}
```

Note: `\0000a0` is a non-breaking space.

One benefit is that the `counter()` CSS function takes a counter style name as the second parameter.
This can be useful when you want to mix counter styles.
Here, I want to use roman numerals on chapters, decimal numbering on sections and lower alpha on subsections:

[...]

Issue #3702, opened 2020-07-07
Configure Format of Section Numbers


VeitSanner:
For the document I am writing, I need to set the section numbers to the format `<level 1>.<level 2>`, e.g. `2.2`. 

The default number format for AsciiDoctor seems to be `<level 1>.<level 2>.`, e.g. `2.2.`.

Is there a way to modify the numbering format, so that the trailing dot after the last number is not printed?

mojavelinux:
This customization was first proposed in Asciidoctor PDF. https://github.com/asciidoctor/asciidoctor-pdf/issues/858#issuecomment-529310500 In that issue, I suggested that the change should be made to core (here). So let's consider the design proposed in that issue.

VeitSanner:
@mojavelinux this would resolve my request. Thanks for the fast reply.

xor2003:
Need it too

retifex:
I love Asciidoc, but unfortunately, for me this is a showstopper bug that prevents me from using Asciidoc for writing my term papers. I would therefore also be happy if I could finally number my headings in Asciidoc according to the pattern "1.1", without a dot at the end. As others have pointed out, that is a very common requirement.

retifex:
For the time being, concerning PDF output I use the [workaround](https://github.com/asciidoctor/asciidoctor-pdf/issues/858#issuecomment-337217358) by @janosvitok, which works perfectly.
Unfortunately I'm neither a Ruby nor an AsciiDoc expert. Can this workaround be used or modified so it works for HTML output? Or is there another existing workaround for HTML? (Besides doing regex search&replace in my favorite text editor afterwards)

retifex:
Alright, sorry for spamming the thread. I'm really a noob, but i found out this.
Theoretically, you could hack `lib/asciidoctor/converter/html5.rb`, modifying the following two lines (as of AsciiDoctor 2.0.18).
Line 363:
```ruby
-          stitle = %(#{section.sectnum} #{section.title})
+          stitle = %(#{section.sectnum('.', false)} #{section.title})
```
Line 397:
```ruby
-        title = %(#{node.sectnum} #{node.title})
+        title = %(#{node.sectnum('.', false)} #{node.title})
```
Maybe some Ruby/AsciiDoctor expert can make this a proper patch that can be required via CLI instead of hacking the original source file. Thanks in advance!

Issue #2934, opened 2018-10-25
Inline anchor in section title breaks sectlinks


mojavelinux:
When the sectlinks attribute is set on the document, adding a secondary, inline anchor in a section title breaks the automatic link surrounding the section title.

Example:

```
:sectlinks:

[#foo]
== [[fu]]Foo
```

This happens because anchor (a) tags can't be nested in HTML.

The workaround is to change from using an a tag to a span tag for anchors.

The same problem will happen anywhere an anchor is added inside a link.

mojavelinux:
There's absolutely no difference between using `<a>` or `<span>` for the purpose of making an anchor (e.g., `<span id="fu"></span>`), except that using `<span>` circumvents the problem described.

mojavelinux:
Although this should not affect any existing documents, it does substantially change the HTML generated. We could consider using `<span>` for anchors on in section titles. The downside of that approach is then you end up with two different strategies throughout the document.

mojavelinux:
What I'm trying to determine is whether this change could be introduced in a patch or minor version, or whether this would warrant a major version.

mojavelinux:
I have an idea for a slightly less invasive change, one that could even be backported.

If the converted title starts with `<a id=`, then skip over any anchors and wrap the text that follows in the link tag. In other words, the example above would produce:

```
<h2 id="foo"><a id="fur"></a><a class="link" href="#foo">Foo</a></h2>
```

Then, there's no chance of breaking existing publications.

mojavelinux:
I'd like to make this fix for 2.0.x, then we can revisit whether to change all `<a>` tags used as anchors (ironically) to `<span>` tags in 2.1.x.

PR #4039, opened 2021-04-23
ref #3981 modern HTML section

Mogztter:
### What I did

- Introduce a method to generate section numbering
- Wrap section numbering in a `<span>` element (in case you want to tweak the design)
- Use `<section>` element (ID and roles are assigned to the `<section>` element)
- Do not wrap section content in a container

### Limitations

I didn't implement section level 0 since it will require additional work (potentially wrapping book parts in an `<article class="part">` element as suggested by @mojavelinux).

ref #3981


lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -49,6 +81,24 @@ def convert_inline_anchor node
     end
   end
 
+  def generate_section_numbering node
+    level = node.level
+    doc_attrs = node.document.attributes
+    if node.numbered && level <= (doc_attrs['sectnumlevels'] || 3).to_i
+      if level < 2 && node.document.doctype == 'book'
+        if node.sectname == 'chapter'
+          %(#{(signifier = doc_attrs['chapter-signifier']) ? "#{signifier} " : ''}<span class="section-numbering">#{node.sectnum}</span>)
```
mojavelinux, line 90: I think this class name should be shortened from section-numbering to sectnum. This is a class name which is going to appear very frequently, so we need to be mindful of how much extra visual and virtual weight this adds to the page. I'm open to other suggestions as well.

test/fixtures/semantic-html5-scenarios/section-with-chapter-signifier.adoc
```diff
@@ -0,0 +1,17 @@
+:doctype: book
+:sectnums:
+:noheader:
+:chapter-signifier: Chapter
+
+[chapter]
+= Part one
```
mojavelinux, line 7: I would prefer fixtures to be well formed and for attribute entries to be below the document title. Here, the part is being reclassified as a chapter, which isn't even technically supported by AsciiDoc. This document should be:

```
= Document Title
:doctype: book
:sectnums:
:noheader:
:chapter-signifier: Chapter

== Introduction

=== Background

==== Problem

== Literature Review

=== Model
```

test/fixtures/semantic-html5-scenarios/section-with-chapters.adoc
```diff
@@ -0,0 +1,13 @@
+:doctype: book
+:sectnums:
+:noheader:
+
+= Chapter One
```
mojavelinux, line 5: This document is also not well-formed. What is marked as "Chapter One" is, in fact, the document title and parts are below chapters.

test/fixtures/semantic-html5-scenarios/section-with-id.adoc
```diff
@@ -0,0 +1,4 @@
+[#tigers-subspecies]
```
mojavelinux, line 1: If a fixture is going to start at a section level below level 1, it should be marked as a fragment document.

mojavelinux (changes requested):

test/fixtures/semantic-html5-scenarios/section.html
```diff
@@ -0,0 +1,6 @@
+<section id="_section_title">
+<h2>Section Title</h2>
+<p>
+A normal paragraph.
+</p>
```
mojavelinux, line 5: I know this isn't part of this PR, but I don't like that there are newlines directly inside the `<p>` tag. This strikes me as something people are going to complain about as being weird HTML, which is what we're trying to get away from. Should I file a separate issue?

test/fixtures/semantic-html5-scenarios/section-with-id-roles.adoc
```diff
@@ -0,0 +1,4 @@
+[#exciting.lead.awesome]
```
mojavelinux, line 1: Perhaps `#exciting.awesome.feature` instead. I don't want to overload the use of the term "lead".

mojavelinux:
> Introduce a method to generate section numbering

:+1: 

> Do not wrap section content in a container

I think this is the right choice. We only ever use it once in the default stylesheet, and that's only to address a quirk with the preamble.

> Use `<section>` element (ID and roles are assigned to the `<section>` element)

There has been some confusion in the past as to why the role isn't on the h[n] tag. However, if we provide a consistent structure, I believe it's easy to target that tag using a CSS selector. And we can always add an autogenerated ID or role later if we determine it needs to be on both. I think adding it to section is the right place to start.

> Wrap section numbering in a `<span>` element (in case you want to tweak the design)

I like that. Though, as I said in the review, I think the name of the class is too verbose. I prefer "sectnum". For what it's worth, the WHATWG uses the class name "sectno". (See https://html.spec.whatwg.org/multipage/introduction.html#is-this-html5?). However, it might be better to be consistent with our own terminology.

mojavelinux:
Btw, I think https://html.spec.whatwg.org/multipage/introduction.html#is-this-html5 is a useful reference to consult in general while developing this converter.

lib/asciidoctor/converter/semantic-html5.rb
```diff
@@ -49,6 +81,24 @@ def convert_inline_anchor node
     end
   end
 
+  def generate_section_numbering node
```
mojavelinux, line 84: While we're changing the section number output, we might want to consider offering a preview of the fix for https://github.com/asciidoctor/asciidoctor/issues/3702 (click through to the comment for the proposal). This is controlled by the converter anyway when it calls `Section#sectnum`. For parts, it should look for `partnumsuffix` first, then `sectnumsuffix`, then fall back to the current values.

I'm leaning towards `sectnumseparator` and `sectnumsuffix` (as complements to `idseparator` and `idprefix`). I had proposed dropping the `sect` prefix on the name, but someone argued that numbers in captions should be treated differently, so we're not sure about that yet.

mojavelinux:
Also consider using `<span>` instead of `<a>` for anchors. See https://github.com/asciidoctor/asciidoctor/pull/3201

test/fixtures/semantic-html5-scenarios/section.html
Mogztter: Yes sure 👍🏻

test/fixtures/semantic-html5-scenarios/section-with-id.adoc
Mogztter: Are you suggesting that we use a naming convention (i.e., for instance a prefix `fragment-section-id.adoc`) or that we use two distinct set of fixtures in order to convert fragment documents or complete documents?

test/fixtures/semantic-html5-scenarios/section-with-chapter-signifier.adoc
Mogztter: I think my goal was to create an headless document with section level 1. I want fixtures to be as independent as possible. If I add a document title then this test will break if I do some changes on document title (even though sections are fine).

Maybe we should improve the "test engine" in order to validate a given portion of the HTML produced (and not systematically the complete HTML)?

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: It makes me think that we should try to adhere to a naming convention. For now, I'm using kebab case without any abbreviation or shortening.

I'm fine with `sectnum`, it's also consistent with the `sectnums` attribute.

So the rule would be short/abbreviated words in lowercase with no separator (and preferably consistent with existing AsciiDoc attributes)?

Mogztter:
> Also consider using `<span>` instead of `<a>` for anchors. See #3201

Section anchors should use `<a>` because we have a target (`href`). For inline anchors, we can indeed use `<span>` but I'm not there yet 😉

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: I'm inclined to say yes to that convention. Though we also have to be mindful of the fact that it won't be perfect, and we may have to make exceptions. But we'll do the best we can to make it intuitive.

test/fixtures/semantic-html5-scenarios/section-with-chapter-signifier.adoc
mojavelinux: > Maybe we should improve the "test engine" in order to validate a given portion of the HTML produced (and not systematically the complete HTML)?

That's one idea. The other is to rely on the noheader attribute for cases like these.

test/fixtures/semantic-html5-scenarios/section-with-id.adoc
mojavelinux: I'm saying to avoid warnings, a document which doesn't start with a top-level section should define the `fragment` attribute at the top of the document. See https://docs.asciidoctor.org/asciidoc/latest/attributes/document-attributes-reference/#section-title-and-table-of-contents-attributes (Technically, it's not required, but considered a good practice IMO).

mojavelinux:
> Section anchors should use <a> because we have a target (href). For inline anchors, we can indeed use `<span>` but I'm not there yet 

:+1: Yes, that's exactly what I'm suggesting. I'm referring to anchor as the target, not anchor as in a self link.

test/fixtures/semantic-html5-scenarios/section-with-chapter-signifier.adoc
Mogztter: > The other is to rely on the noheader attribute for cases like these.

I like the `noheader` approach, let's keep it simple for now 😄

test/fixtures/semantic-html5-scenarios/section-with-id.adoc
Mogztter: Oh I wasn't aware of this attribute, noted!

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: Should we create a list/table with all the built-in classes/roles that we are using in this new converter? So far:

| Name | Context |
| --- | --- |
| `link` | Section links |
| `anchor` | Section anchors |
| `title` | Paragraph title |
| `sectnum` | Section numbering |
| `subtitle` | Document subtitle |
| `byline` | Author(s) block |
| `author` | Author line |
| `revision` | Revision table |

lib/asciidoctor/converter/semantic-html5.rb
Mogztter: I've added #3702 to the [Create Modern HTML Converter project](https://github.com/asciidoctor/asciidoctor/projects/1).
I would rather keep that for a follow-up pull request.

Mogztter:
I've added https://github.com/asciidoctor/asciidoctor/issues/2934 to the [Create Modern HTML Converter project](https://github.com/asciidoctor/asciidoctor/issues/2934) to make sure that we will address this issue. As mentioned in the issue we might address this issue sooner, in this case we will need to "backport" this fix to the new converter.

lib/asciidoctor/converter/semantic-html5.rb
mojavelinux: That's a good idea. And it's also forward-looking for the specification. Because I would like the specification to be more clear about what the HTML "API" is so to speak when composing a stylesheet. And this might become part of that.

Mogztter: For now, I don't want to do anything fancy, just a plain list. Should I add this (hidden) page in `docs` (i.e., without a navigation)? in https://github.com/asciidoctor/asciidoctor/tree/master/docs/modules/api/pages?

mojavelinux: Since docs aren't published from this branch, it could really go anywhere for now. We might want to think about point it in a devdocs folder for now, then deciding later if and where it might fit in the user docs.

test/fixtures/semantic-html5-scenarios/section.html
Mogztter: Done: https://github.com/asciidoctor/asciidoctor/issues/4057

Mogztter:
@mojavelinux could you please review it again, I think I've made all the requested changes.

mojavelinux (approved):

PR #4040, opened 2021-04-25
resolves #3860 don't include trailing punctuation in target of autolink

### LINKED ISSUE #3860: If a bare URL is followed by a period and a closing parenthesis, the period should not be treated as a part of the URL
```
(The homepage for the Asciidoctor Project is https://asciidoctor.org.)
```

The period is treated as a part of the URL.

A workaround that I use:

```
(The homepage for the Asciidoctor Project is link:++https://asciidoctor.org++[].)
```

Probably a bug.

mojavelinux:


Mogztter (approved):
👍🏻

PR #4047, opened 2021-05-03
resolve #4046 change ifeval directive to resolve to false if comparison operation cannot be performed

### LINKED ISSUE #4046: >=: undefined method `>=' for nil
This originates from:

```
ifeval::[{release} >= 4.5]
```

where `{release}` is not defined anywhere. It builds without any errors when using CLI (`asciidoctor -B . -a toc -d book <file>`), but not with JavaScript:

```js
const asciidocOptions = {
  doctype: 'book',
  // asciidoctor: WARNING: include file is outside of jail; recovering automatically
  //safe: 'server',
  safe: 'unsafe',
  sourcemap: true,
  base_dir: repoDir,
  attributes: {
    icons: 'font',
    toc: 'macro'
  }
}
const doc = adoctor.loadFile(assembly, asciidocOptions);
doc.convert();
```

Is there any option that avoids the exception?

Thanks!

mojavelinux:


lib/asciidoctor/reader.rb
```diff
@@ -958,11 +958,11 @@ def preprocess_conditional_directive keyword, target, delimiter, text
...
           if text && EvalExpressionRx =~ text.strip
-            lhs = $1
-            op = $2
-            rhs = $3
+            lhs = resolve_expr_val $1
             # regex enforces a restricted set of math-related operations (==, !=, <=, >=, <, >)
-            skip = ((resolve_expr_val lhs).send op, (resolve_expr_val rhs)) ? false : true
+            op = $2
+            rhs = resolve_expr_val $3
+            skip = (lhs.send op, rhs) ? false : true rescue true
```
Mogztter, line 965: I was going to say, is it possible to have other errors besides `NoSuchMethod` (such as `ArgumentError`).
`rescue true` will catch any error?

mojavelinux: I can't think of any other errors at the moment. But regardless, if the operation cannot be performed, it means the condition is not true. So it is correct to skip in this case.

Mogztter (approved):

PR #4060, opened 2021-05-15
manpage: fix apostrophe portability

felipec:
As you can see from the bug referenced in the preamble, the point is to
not convert \' to \(aq directly, because that only works in groff. To
make it portable \*(Aq should be used instead, and that will be
converted to \(aq in groff and ' elsewhere.

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=507673

### DIFF
--- a/lib/asciidoctor/converter/manpage.rb
+++ b/lib/asciidoctor/converter/manpage.rb
@@ -743,7 +743,7 @@ allbox tab(:);'
       .gsub('&#8658;', '\(rA')  # rightwards double arrow
       .gsub('&#8203;', '\:')    # zero width space
       .gsub('&amp;', '&')       # literal ampersand (NOTE must take place after any other replacement that includes &)
-      .gsub('\'', '\(aq')       # apostrophe-quote
+      .gsub('\'', '\*(Aq')      # apostrophe-quote
       .gsub(MockMacroRx, '\1')  # mock boundary
       .gsub(ESC_BS, '\\')       # unescape troff backslash (NOTE update if more escapes are added)
       .gsub(ESC_FS, '.')        # unescape full stop in troff commands (NOTE must take place after gsub(LeadingPeriodRx))
--- a/test/manpage_test.rb
+++ b/test/manpage_test.rb
@@ -256,6 +256,16 @@ context 'Manpage' do
       assert_includes output, 'go\\(emto'
     end
 
+    test 'should replace quotes' do
+      input = <<~EOS.chop
+      #{SAMPLE_MANPAGE_HEADER}
+
+      'command'
+      EOS
+      output = Asciidoctor.convert input, backend: :manpage
+      assert_includes output, '\*(Aqcommand\*(Aq'
+    end
+
     test 'should escape lone period' do
       input = <<~EOS.chop
       #{SAMPLE_MANPAGE_HEADER}


mojavelinux:
Can you point to a reference where *(Aq is defined? Can you also add a test?

What tool do I use to test this macro without using groff?

mojavelinux:
Thanks for adding the test!

felipec:
> Can you point to a reference where *(Aq is defined?

It's in line 64 [manpage.rb#L64](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/manpage.rb#L64).

> Can you also add a test?

Sure. Done.

> What tool do I use to test this macro without using groff?

I don't know, I suppose you would need a system using Solaris, or something like that.

mojavelinux:
> > Can you point to a reference where *(Aq is defined?

> It's in line 64 manpage.rb#L64.

I meant where is *(Aq defined as a portable replacement for \(aq? I'm looking for a reference site.

> > What tool do I use to test this macro without using groff?

> I don't know, I suppose you would need a system using Solaris, or something like that.

Then how does this have anything to do with Debian?

felipec:
> > It's in line 64 manpage.rb#L64.
> 
> I meant where is *(Aq defined as a portable replacement for (aq? I'm looking for a reference site.

`*(Aq` is an alias you define at the top of every generated man page. You decided that name. You can replace all the `Aq` instances with `Fo` and it will still work.

This is what you write on top of every man page:

```groff
.ie \n(.g .ds Aq \(aq
.el       .ds Aq '
```

I'm no expert in groff, but this is the equivalent of

```ruby
if GROFF
  Aq = %[(aq]
else
  Aq = %[']
end
```

> Then how does this have anything to do with Debian?

This has absolutely nothing to do with Debian. You decided to put a link to the Debian bug tracker as reference, but there's also a reference to the groff mailing list: [\(aq not working in OpenSolaris environment](https://lists.gnu.org/archive/html/groff/2009-02/msg00013.html).

There's plenty of other references.

Another reference is how they fixed the problem in docbook: [Fixed bug #2412738 (apostrophe escaping) by applying the submitted patch.](https://github.com/docbook/xslt10-stylesheets/commit/fb553434265906ed81edc6d5f533d0b08d200046l).

They fixed it in precisely the same way, in fact, the patch came from Debian.

You presumably copied the header of a man page generated with asciidoc, but used `\(aq` instead of `\*(Aq` in the actual body.

Generate a man page using asciidoc, you'll see the difference.

For more history and references check [this patch I sent to git.git](https://lore.kernel.org/git/20210515115653.922902-2-felipe.contreras@gmail.com/).

mojavelinux:
> You decided to put a link to the Debian bug tracker as reference

I can assure you that I did not add the link that is in the description of this PR. And no one else has edited this PR. So it must have been you who added that link.

mojavelinux:
> You presumably copied the header of a man page generated with asciidoc, but used `\(aq` instead of `\*(Aq` in the actual body.

Aha! That's what I needed to know. I did not think to look in the header generated by this manpage converter. Now that you pointed me to it, I see where the definition is coming from. Now it makes sense.

mojavelinux:
> I don't know, I suppose you would need a system using Solaris, or something like that.

I'm still confused how you came to report this bug if it doesn't affect the system you are using. But it really doesn't matter, because the problem you've pointed out is, in fact, a valid one for non-groff users. So we can proceed with this fix.

felipec:
It has been there since the first commit: 7bddc416.

The alias is there in asciidoc-backends too: [document.rb](https://github.com/asciidoctor/asciidoctor-backends/blob/master/erb/manpage/document.erb).

Except in their code that did actually use `\*(Aq`: [helpers.rb](https://github.com/asciidoctor/asciidoctor-backends/blob/master/erb/manpage/helpers.rb):

```ruby
.gsub('\'', '\\*(Aq')
```

felipec:
> I'm still confused how you came to report this bug if it doesn't affect the system you are using.

It doesn't affect me. I'm a git developer and I'm trying to make sure the git documentation generates proper man pages with asciidoctor.

The git code had a workaround for the docbook bug and I spent several hours digging into discussions and bug reports in different projects from 2008 to 2010.

After I did all that I checked how you handled this issue I noticed that you hadn't.

mojavelinux:
> It has been there since the first commit

Yep, I'm not questioning that. I just didn't know/remember (many years later) that it was coming from there. When I first started with this convertor, groff was all cryptic to me. Now the syntax makes a lot more sense, but I haven't studied it line-by-line in a long time.

> I'm a git developer and I'm trying to make sure the git documentation generates proper man pages with asciidoctor.

Gotcha.

> After I did all that I checked how you handled this issue I noticed that you hadn't.

I do appreciate you taking the time to bring the fix to this project. So thanks again!

PR #4063, opened 2021-05-17
rewrite Reader#read_lines_until

mojavelinux:
Give this refactoring a chance to run through CI.


Mogztter:
@mojavelinux does it have any impact on the performance? I guess `read_lines_until` is a hotspot no?
I will also check on Asciidoctor.js 😉

mojavelinux:
Yes, I believe it will have an impact on performance, though I can't say by how much. What the code was doing was very inefficient, and this change cuts out both the unnecessary while loop and unnecessary checks.

PR #4064, opened 2021-05-18
resolves #4062 import source and build for default stylesheet

### LINKED ISSUE #4062: Import source and build for default stylesheet into this repository
The default stylesheet (for the HTML output) is currently generated by the asciidoctor-stylesheet-factory project. However, that project uses a build which relies on unmaintained software. It's also arduous and obscure to have to make changes to files into another project and import the generated result into this repository to be committed. The whole process will be a lot simpler if that build is located in this project.

To avoid introducing unanticipated changes to the stylesheet, we'll need to work backwards from the minified/distributed end result until we get to something that's maintainable. This issue focuses on the first stage in that migration, which is described below:

1. Unminify data/stylesheets/asciidoctor-default.css (this will become the editable source file)
    * The source file will be maintained in plain CSS with some help from PostCSS
2. Set up and configure a npm-based build that uses postcss + cssnano to produce the same or similar minified result as before
3. Use the build to generate data/stylesheets/asciidoctor-default.min.css
4. Update the processor to read the default styles from the data/stylesheets/asciidoctor-default.min.css file instead of data/stylesheets/asciidoctor-default.css file
    * If this doesn't work out, the source file could be named data/styelsheets/asciidoctor-default.src.css instead
5. Configure the gemspec to omit the stylesheet source file from the gem

We'll probably want to avoid committing the minified file to the repository and generating it on demand for the test suite and release, but we should do that in a later stage.

mojavelinux:


mojavelinux:
Note that this build currently produces the exact default stylesheet that's in HEAD. Once we merge this build, we can continue to improve the minified styles, using git to track the changes. For example, we'll want to switch to autoprefixer instead of using prefix property names explicitly.

Mogztter:
It looks good!

Do we want to make sure that `data/stylesheets/asciidoctor-default.css` is not out of sync? If so, we could add a new workflow that generates the stylesheet (`npm run build`) and check if there's any local changes (`[ -z "$(git status --porcelain)" ];`)?
It will also ensure that the stylesheet generation/build is working.

Apart from that 👍🏻

stylesheets/lib/postcss-minify-selectors.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const selectorParser = require('postcss-selector-parser')
+
+module.exports = (opts) => {
+  return {
+    postcssPlugin: 'postcss-minify-selectors',
+    Rule (rule) {
```
Mogztter, line 8: I think we should explain this custom rule.
Do we need it in order to generate the exact same default stylesheet? Is it an intermediate step? Do we want to keep it?

stylesheets/lib/postcss-rule-per-line.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = (opts) => {
+  return {
+    postcssPlugin: 'postcss-rule-per-line',
+    OnceExit (root) {
+      root.walk((node) => {
+        if (node.type.endsWith('rule') && node.prev()) node.raws.before = '\n'
+      })
+      root.raws.after = '\n'
+    },
```
Mogztter, line 11: Ditto.
I guess the goal is to add a newline before and after each rule (otherwise the minified CSS is a single line) right? But is that temporary? Or should we remove it at a later stage?

stylesheets/postcss.config.js
```diff
@@ -0,0 +1,19 @@
...
+    cssnano: {
+      preset: [
+        'default',
+        {
+          discardComments: { exclude: true },
+          minifySelectors: { exclude: true }, // this one we'll do ourselves
+          minifyFontValues: { exclude: true },
+          mergeRules: { exclude: true }, // TODO reenable
+          uniqueSelectors: { exclude: true },
+          cssDeclarationSorter: { exclude: true },
```
Mogztter, line 12: I think we should explain why each rule are disabled. Do we want to re-enable `discardComments`, `minifyFontValues`, `uniqueSelectors` and `cssDeclarationSorter`? If not, why?

[`cssDeclarationSorter`](https://cssnano.co/docs/optimisations/cssdeclarationsorter/) is probably not worth it, since, as far as I understand, the benefit would be relatively low. 

We could also add a link on the documentation:

- https://cssnano.co/docs/optimisations/minifyfontvalues/ 
- https://cssnano.co/docs/optimisations/cssdeclarationsorter/
- ...

stylesheets/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules/
+/package-lock.json
```
Mogztter, line 2: you didn't commit `package-lock.json`, is that intended?

mojavelinux:
My main concern is whether we've organized the files correct. Do you think the stylesheets directory is the right choice, or do you have another idea in mind? I do want to keep it out of the gem, which is why I didn't put it in either data or lib.

stylesheets/lib/postcss-minify-selectors.js
mojavelinux: I agree, the rules need some explanation. The purpose of this rule is two fold:

* Minify the selectors without sorting them
* Don't break pseudo elements (I don't know why cssnano insists on using the incorrect syntax)

stylesheets/lib/postcss-rule-per-line.js
mojavelinux: I find single-line minified files to be impossible to inspect. I've always broken the rules out on each line in the default stylesheet, and it has proven to be a good strategy. So I'm sticking with it here.

stylesheets/postcss.config.js
mojavelinux: The main reason for tuning the rules is to be able to start with a 1-to-1 match with what we have now. We don't want this step (the switch to PostCSS+cssnano) to introduce any changes. We want to take those changes in as separate commits.

* discardComments: We only have three comments at the top and we want to keep those; if we add comments elsewhere, this will need to be adjusted
* minifyFontValues switches to numeric weights and I don't like that because it makes the stylesheet harder to build on
* cssDeclarationSorter is pointless for minification; it's just sorting the property names, which introduces HUGE changes
* uniqueSelectors: this is just adding a lot of changes, so we're not ready for it yet...but something to consider

> We could also add a link to the documentation

:+1:

stylesheets/.gitignore
mojavelinux: For a dev build, I prefer not to commit the lock file since it really doesn't matter. I'd rather detect any problems we might encounter in CI.

mojavelinux:
> Do we want to make sure that data/stylesheets/asciidoctor-default.css is not out of sync?

Great idea.

I'll implement the changes you suggested and request a follow-up review.

Mogztter:
> My main concern is whether we've organized the files correct. Do you think the stylesheets directory is the right choice, or do you have another idea in mind? I do want to keep it out of the gem, which is why I didn't put it in either data or lib.

`stylesheets` is fine, what are your concerns?

.github/workflows/ci.yml
```diff
@@ -85,6 +85,18 @@ jobs:
         bundle --jobs 3 --retry 3
     - name: Run tests
       run: bundle exec ruby -w $(bundle exec ruby -e 'print File.join Gem.bindir, %q(rake)') test:all
+    - name: Check default stylesheet
+      if: matrix.primary
+      run: |
+        cd stylesheets
+        npm i --silent --no-audit --no-audit --no-fund --no-package-lock --no-progress
+        npm run build
+        cd ..
```
Mogztter, line 94: you can use `--prefix` to avoid `cd stylesheets` and `cd ..`


```suggestion
        npm i --silent --no-audit --no-audit --no-fund --no-package-lock --no-progress --prefix stylesheets
        npm run build --prefix stylesheets
```

Mogztter: Another solution would be to run `npm i` and `npm run build` in a dedicated task with `working-directory` (and then check if there's any uncommitted changes in another task)

stylesheets/.gitignore
Mogztter: OK 👍🏻

mojavelinux:
> stylesheets is fine, what are your concerns?

Just that it's a new top-level directory and I want to get the name right. I thought about `src/stylesheets` (kind of like you have in asciidoctor.js, but I wasn't sure because lib and src then kind of looks confusing at first glance). I've also consider stylesheets-src (or src-stylesheets), which I might be more comfortable with. Keep in mind, this will eventually host both stylesheet builds, and perhaps any others we end up adding.

Mogztter:
> I've also consider stylesheets-src (or src-stylesheets), which I might be more comfortable with

Keep it simple, `stylesheets` is fine. 

> and perhaps any others we end up adding.

Unless we have a pretty good idea of what the "others" will be, I think we shouldn't try to extrapolate.
Having said that, we could opt for a more generic name like "resources" if we feel like "stylesheets" is too specific.

PR #4074, opened 2021-06-05
Fix documentation about shared-header

gemmaro:
Added the missing `shared-header` in the docinfo list ([Asciidoctor Docs > Docinfo Files > Enabling docinfo](https://docs.asciidoctor.org/asciidoctor/latest/docinfo/#enable)).


mojavelinux:
Thanks for catching and fixing this!

PR #4083, opened 2021-06-13
Fix typo cooresponding -> corresponding

Mogztter:


mojavelinux:
Nice catch. I always struggle with the spelling of that word ;)

PR #4084, opened 2021-06-13
Upgrade Truffle Ruby to 21.1.0

Mogztter:


mojavelinux:
Thanks!

Mogztter:
Truffle 21.1.0 seems to be way faster than 20.0.2:

|Truffle 21.1.0|Truffle 20.0.2|
|--|--|
|image|image|


The fastest build I found with Truffle 20.0.2 took 3m19s (slower more than 4m). With Truffle 21.1.0 the build took only 2m33s.

mojavelinux:
Excellent news! :tada:

mojavelinux:
What I think that means is that we could now enable it on every run since it's not taking any longer than JRuby.

mojavelinux:
> What I think that means is that we could now enable it on every run since it's not taking any longer than JRuby.

:rofl: We already did that.

Mogztter:
Ah ah ah!
Still less CI time so 🎉

PR #4092, opened 2021-06-27
resolves #4071 use semantic constructs for listing blocks

### LINKED ISSUE #4071: Use semantic constructs for listing blocks
This issue will be addressed as part of: https://github.com/asciidoctor/asciidoctor/issues/242

### Preformatted text

Listing blocks display their rendered content exactly as they were entered by the user.
The [`<pre>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) fits the description as it represents a block of preformatted text.

> Some examples of cases where the pre element could be used [...] Including fragments of computer code, with structure indicated according to the conventions of that language. [...]

https://html.spec.whatwg.org/multipage/grouping-content.html#the-pre-element

### Code

When a language is defined (source block), we should wrap the content in a `<code>` element.
Please note that it only applies when no syntax highlighter is defined.

The language should be added on the `<code>` element as a data attribute: `<code data-lang="ruby">`.
Not sure if we should also add the language as a class, something like: `<code class="language-ruby" data-lang="ruby">`...?
Arguably, we don't need it because we can use data attribute as CSS selector:

```css
/* Selects if the attribute has a particular value */
code[data-lang="ruby"] {
}

/* Selects if the attribute is present at all */
code[data-lang] {
}
```

But at the same it might be useful?

### Figure

> The figure element represents some flow content, optionally with a caption, that is self-contained (like a complete sentence) and is typically referenced as a single unit from the main flow of the document.

https://html.spec.whatwg.org/multipage/grouping-content.html#the-figure-element

MDN Web Docs as more or less the same definition:

> The HTML <figure> (Figure With Optional Caption) element represents self-contained content, potentially with an optional caption, which is specified using the <figcaption> element. The figure, its caption, and its contents are referenced as a single unit.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure

Both put emphasis on the fact that content wrapped in a `<figure>` element can be moved to another part of the document or to an appendix without affecting the main flow.
Therefore, authors are encouraged to consider using labels to refer to figures, rather than using relative position, e.g., "in the photograph above" or "as the next figure shows", so that the page can easily be restyled without affecting the page's meaning.

I guess it's a general writing guideline that can apply when writing in AsciiDoc.

Listing block:

```html
<figure class="listing">
<pre>error: 1954 Forbidden search
absolutely fatal: operation lost in the dodecahedron of doom

Would you like to try again? y/n</pre>
</figure>
```

Listing block with title:

```html
<figure class="listing">
<figcaption><span class="label">Listing 1.</span> Gemfile.lock</figcaption>
<pre>GEM
  remote: https://rubygems.org/
  specs:
    asciidoctor (2.0.12)

PLATFORMS
  ruby

DEPENDENCIES
  asciidoctor (~> 2.0.12)</pre>
</figure>
```

Source block with title and roles:

```html
<figure id="ruby-code" class="listing ruby code">
<figcaption><span class="label">Listing 1.</span> Some Ruby code</figcaption>
<pre><code data-lang="ruby">require 'sinatra'

get '/hi' do
  "Hello World!"
end</code></pre>
</figure>
```

As you can see we are using a built-in role/class named "listing" to differentiate listing block and literal block:

```html
<figure class="literal">
<pre>error: 1954 Forbidden search
absolutely fatal: operation lost in the dodecahedron of doom

Would you like to try again? y/n</pre>
</figure>
```

We are also using a `<span>` element on the caption label to be able to style it.

#### Figure caption position

The `<figcaption` element can be declared as the first or last child of a `<figure>`:

##### before 

image

##### after

image

In theory we could use this capability to configure (via an attribute?) if we want to display the caption at the start or at the end of the listing block.

Having said that, we can also do that using CSS and I don't know if other non-HTML converters will be able to leverage this feature.

Mogztter:
resolves #4071

### DIFF
--- a/lib/asciidoctor/converter/semantic_html5.rb
+++ b/lib/asciidoctor/converter/semantic_html5.rb
@@ -67,6 +67,31 @@ class Converter::SemanticHtml5Converter < Converter::Base
     end
   end
 
+  def convert_listing node
+    caption = (caption = node.caption) ? %(<span class="label">#{caption}</span> ) : ''
+    title = node.title? ? %(<figcaption>#{caption}#{node.title}</figcaption>\n) : ''
+    attributes = common_html_attributes node.id, node.role, 'listing'
+    nowrap = (node.option? 'nowrap') || !(node.document.attr? 'prewrap')
+    if node.style == 'source'
+      lang = node.attr 'language'
+      if (syntax_hl = node.document.syntax_highlighter)
+        opts = syntax_hl.highlight? ? {
+          css_mode: ((doc_attrs = node.document.attributes)[%(#{syntax_hl.name}-css)] || :class).to_sym,
+          style: doc_attrs[%(#{syntax_hl.name}-style)],
+        } : {}
+        opts[:nowrap] = nowrap
+        content = syntax_hl.format node, lang, opts
+      else
+        content = %(<code#{lang ? %( data-lang="#{lang}") : ''}>#{node.content || ''}</code>)
+      end
+    else
+      content = node.content || ''
+    end
+    %(<figure#{attributes}>
+#{title}<pre#{nowrap ? ' class="nowrap"' : ''}>#{content}</pre>
+</figure>)
+  end
+
   def convert_thematic_break node
     '<hr>'
   end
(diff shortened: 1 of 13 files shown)


Mogztter:
Build is now failing because the code does not comply with the linter (RuboCop). I will update the code.

PR #4107, opened 2021-07-25
resolves #4106 use queue to iterate over lines in reader instead of stack

### LINKED ISSUE #4016: cross-reference links with __ are not rendered correct
Hi,

I try to write a cross-reference link to a file, that contains double underscores in its path. In the rendered result the __ are replaced with `<em>` and `</em>`.

```
<<./doc__with_double__underscore.adoc#, MyLink>>
```
is rendered to 
```
<a href="./doc<em>with_double</em>underscore.html">MyLink</a>
```

### LINKED ISSUE #4106: Update reader to use a queue to iterate over lines instead of a stack
The reader prepares the source lines and iterates over them using a stack, meaning it reads the next line by shifting it off the start of an array. The reader should instead iterate over the lines using a queue, meaning it should reverse the array of lines when preparing them and read the next line by popping it off the end of the array.

While there isn't a significant performance difference between these two methods in Ruby, what we've discovered is that there is an enormous difference in JavaScript, particularly when the number of lines exceeds a certain threshold (~50,000) (particularly Node.js 12 and above). Hence, this negatively affects the performance of Asciidoctor.js in a very noticeable way.

In general, if not optimized, shifting a line off the start of an array can be very costly. That's because the index of every other item in the array has to be updated. By popping a line off the end of the array, the only modification to the array is its length. Regardless of whether certain runtimes are optimized for stack operations, the fact remains it's a more efficient to use a queue.

This change can be made with no impact to the API or test suite. That's because the line iteration mechanism is hidden behind the Reader API ( ead_line, peek_line, unshift_line, and the like). By reversing the array in the constructor, and changing the shift/unshift calls to pop/push, we can make this change without breaking any code.

mojavelinux:


Mogztter:
I think it should be "resolves https://github.com/asciidoctor/asciidoctor/issues/4106" not #4016

Mogztter:
With this change, I can confirm that Asciidoctor.js is faster across all versions of Node.js.
Performance are now within an acceptable range between Node.js 10, 14 and 16:

**Node.js v10.22.1 (best run)**

```
Load scripts: 0.074s
Run #1: 0.602s
Run #2: 0.437s
Run #3: 0.41s
Run #4: 0.379s
avg: 457.00
stddev: 86.19
50th: 423.50
75th: 478.25
90th: 552.50
99th: 597.05
```

**Node.js 14.15.5 (best run)**

```
Load scripts: 0.053s
Run #1: 0.612s
Run #2: 0.482s
Run #3: 0.46s
Run #4: 0.431s
avg: 496.25
stddev: 69.23
50th: 471.00
75th: 514.50
90th: 573.00
99th: 608.10
```

**Node.js 16.5.0 (best run)**

```
Load scripts: 0.052s
Run #1: 0.611s
Run #2: 0.493s
Run #3: 0.451s
Run #4: 0.426s
avg: 495.25
stddev: 70.99
50th: 472.00
75th: 522.50
90th: 575.60
99th: 607.46
```

It's interesting to note that Asciidoctor.js with Node.js 10 is still slightly faster.
For reference, here are the results without this change:

**Node.js v10.22.1 (best run)**

```
Load scripts: 0.075s
Run #1: 0.872s
Run #2: 0.666s
Run #3: 0.641s
Run #4: 0.609s
avg: 697.00
stddev: 103.04
50th: 653.50
75th: 717.50
90th: 810.20
99th: 865.82
```

**Node.js 14.15.5 (best run)**

```
Load scripts: 0.054s
Run #1: 1.617s
Run #2: 1.475s
Run #3: 1.447s
Run #4: 1.424s
avg: 1490.75
stddev: 75.09
50th: 1461.00
75th: 1510.50
90th: 1574.40
99th: 1612.74
```

**Node.js 16.5.0 (best run)**

```
Load scripts: 0.053s
Run #1: 1.532s
Run #2: 1.421s
Run #3: 1.365s
Run #4: 1.341s
avg: 1414.75
stddev: 73.66
50th: 1393.00
75th: 1448.75
90th: 1498.70
99th: 1528.67
```

mojavelinux:
> I think it should be "resolves #4106" not #4016

Doh! Working too hard and too late ;)

mojavelinux:
> It's interesting to note that Asciidoctor.js with Node.js 10 is still slightly faster.

While there is still merit in investigating the different, I think we can live with it. It's also expected. I've observed that there are core logic operations in Node.js 16 that are slower than in Node.js 10. It's a tradeoff to curb memory. In effect, Node.js "cheats" by speeding up some logic operations by running them with the memory safeguards off. Running Asciidoctor.js on Node.js 16 gives you the guarantee the process will not run away with memory.

...with that said, we can pull up that flame graph again and see where the other hot spots are. But it's much, much less pressing.

mojavelinux:
After testing on very large docsets, I can now confirm that Asciidoctor.js 2 with this patch running on Node.js 16 runs faster than Asciidoctor.js 1.5.9 running on Node.js 10. As you observed, Asciidoctor.js 2 with this patch running on Node.js 10 still runs the fastest, but the difference is not substantial.

Here are the numbers from one particular docset with just over 9,000 AsciiDoc files:

```
.Node.js 10
3.0.0-alpha.2 (with Asciidoctor.js 1.5.9) :: 5m3s
3.0.0-alpha.7 (with Asciidoctor.js 2.2.4) :: 5m54s
3.0.0-alpha.7 (with patched Asciidoctor.js 2.2.4) :: 4m37s

.Node.js 16
3.0.0-alpha.2 (with Asciidoctor.js 1.5.9) :: 14m24s
3.0.0-alpha.7 (with Asciidoctor.js 2.2.4) :: 9m32s
3.0.0-alpha.7 (with patched Asciidoctor.js 2.2.4) :: 4m55s
```

It's worth noting that the worst performing combination is Asciidoctor.js 1.5.9 running on Node.js 16. That's because Asciidoctor.js 1.5.9 lacks the optimization for Array#$shift in Opal that Asciidoctor.js includes.

Mogztter:
> ...with that said, we can pull up that flame graph again and see where the other hot spots are. But it's much, much less pressing.

I did that but nothing really stands out.
`Parser#next_block` is definitely a hot path. And at a lower-level , `String#ends_with?` and `String#include?` are called quite often but the implementation is relatively fast. Opal is (somehow) still supporting IE so it does not use `String.endsWith` and `String.includes` but the performance gain is insignificant (if any).

Anyway, the performance boost compared to Asciidoctor.js 1.5.9 is amazing ⚡

mojavelinux:
> the performance boost compared to Asciidoctor.js 1.5.9 is amazing zap

Indeed it is. Thanks for working with me on this.

PR #4123, opened 2021-08-04
Create dependabot.yml

jwillemsen:
Will result in automated PRs when a used github action can be upgraded

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# Set update schedule for GitHub Actions
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thank you for offering this suggestion. However, we do not use dependabot to manage or monitor dependencies for this project. In general, I dislike dependabot and find it to be a disruption to development. I check dependencies as part of preparation for a release and do not want to be bothered with them before that. This project also doesn't declare any runtime dependencies, so it's not a risk to update the development dependencies when we see fit.

PR #4132, opened 2021-08-10
resolves #4128 prevent line numbers on source block in HTML output from being selected

### LINKED ISSUE #4128: Disable user select for inline line numbers
When the syntax highlighter is configured to use the inline linenums mode (e.g., `pygments-linenums-mode=inline`), the default stylesheet should prevent the line numbers from being selected (i.e., make the non-selectable). This allows the user to copy the code without the line numbers being caught up in the selection.

Add the following rule to the default stylesheet:

```css
pre.pygments span.linenos {
  user-select: none;
}
```

Or, to prevent accidentally grabbing the linenums column, add it to the td.linenos as well:

```css
table.linenotable td.linenos {
  user-select: none;
}
```

Also add the following rule to the default CodeRay stylesheet:

```css
.CodeRay span.line-numbers {
  user-select: none;
}
```

mojavelinux:
resolves #4128

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -26,6 +26,10 @@ Bug Fixes::
   * Remove unnamespaced selectors in Pygments stylesheet
   * Normalize output from Pygments to use `linenos` class for inline line numbering and trim space after number; update default stylesheet accordingly
 
+Improvements::
+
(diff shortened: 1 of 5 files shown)


mojavelinux:
This depends on #4129 being merged.

Issue #4070, opened 2021-05-27
Add visual regression testing to the stylesheet build


mojavelinux:
To detect regressions in the CSS (and, by extension, the HTML), integrate a visual regression testing tool into the stylesheet build. The tool of choice at the time of writing is [BackstopJS]().

BackstopJS offers visual regression testing for web pages. It does by using puppeteer to control the browser, take screenshots of the rendered page, and compare them against reference screenshots. It not only offers visual inspection against a reference sample, but also an interface to review those changes. The interface offers several diff views, including a before/after wipe. It also has a workflow for promoting acceptable changes back to the reference sample. As a bonus, BackstopJS is a very self-contained application, so it doesn't require much setup.

mojavelinux:
One thing that we'll need to get this going are a set of reference documents. Due to memory limitations, it's not possible to take a screenshot of the entire AsciiDoc.py user guide or Asciidoctor user manual. So we'll need shorter documents that cover the "kitchen sink" of AsciiDoc doctypes, syntax, and permutations of elements. Over time, we should break these out into smaller tests so that specific behaviors can be tested in isolation.

mojavelinux:
I'm getting pretty close to getting this working. I've got something running locally that I'm using to detect style changes, so I know it's going to work. I just need to sort out where the files go and get a few more fixtures to ensure we have sufficient visual coverage.

mojavelinux:
Interestingly, when comparing the stylesheet between 2.0.15 and 2.0.16, I was able to detect the fix for #2550.

PR #4133, opened 2021-08-15
resolves #4130 layout source with table-based linenums as table rows when using rouge as source highlighter

### LINKED ISSUE #4130: Layout lines of source block with line numbering as table rows when using Rouge as source highlighter
Layout the lines of a source block with table-based line numbering enabled as rows of a table when using Rouge as the source highlighter. This layout prevents wrapped lines from causing the line numbers to become misaligned from their respective lines (as reported in #1706). It also keeps the source code out of the line number gutter / column. This layout is the same strategy used to format source code on GitHub and GitLab.

While we can't solve this problem for all syntax highlighter libraries, Rouge gives us the hooks we need to solve it when using Rouge.

mojavelinux:


mojavelinux:
I did a visual test and verified that the output is exactly the same (except in the case when the line wraps, of course). We will want to make this a visual test though once #4070 is resolved.

PR #4144, opened 2021-08-24
resolves #4143 omit meta generator tag if reproducible document attribute is set

### LINKED ISSUE #4143: Feature request: Optionally remove asciidoctor meta tag
https://github.com/asciidoctor/asciidoctor/blob/eafb271a3afc9c331b9807168a8ecfd092ec700d/lib/asciidoctor/converter/html5.rb#L109

I'm trying to make some doc builds more easily reproducible. While I assume that the same version of AsciiDoctor will probably be required to make things work well, it would be helpful if I could disable that meta tag (the way I can the date ones) by defining or removing an attribute from the document.

rpavlik:
Fixes #4143.
The test seems to pass and appears to be right to me, but this was my first Asciidoctor test.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -20,6 +20,7 @@ This project utilizes semantic versioning.
 Enhancements::
 
   * Add support for inline linenums mode when using Rouge as source highlighter (`rouge-linenums-mode=inline`) (#3641)
+  * Omit meta generator tag in HTML output if `reproducible` document attribute is set (#4143) *@rpavlik*
 
 Bug Fixes::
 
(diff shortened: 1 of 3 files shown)


rpavlik:
Oh, I think I should have put this test in api_test.rb instead by the other reproducible test:

```ruby

    test 'should not output HTML meta generator tag if reproducible attribute is set in HTML 5' do
      doc = document_from_string 'text', backend: :html5, attributes: { 'reproducible' => '' }
      result = doc.convert
      assert doc.attr?('reproducible')
      assert_xpath '//meta[@name="generator"]', result, 0
    end
```

Unfortunately that test suite doesn't run for me locally, I presumably have too old of ruby or something. Let me know if you want that change instead.

mojavelinux:
Thanks for the PR!

mojavelinux:
To run the test suite, please follow these instructions: https://github.com/asciidoctor/asciidoctor/blob/master/CONTRIBUTING.adoc#submitting-a-pull-request If you use bundler, as the instructions advise, the test suite run fun. This is not a rails app, and therefore you shouldn't be using rails to run the test suite.

mojavelinux:
Thanks! I squashed and added an entry to the CHANGELOG. Otherwise, looks great!

PR #4146, opened 2021-08-25
Added example of pointing Pygments adapter to one's own stylesheet

rben01:
I think that in the future, a family of attributes like `:pygments-stylesheet:` would make sense. But for now, updating the docs on how to achieve this seems "the long way" seems best.


mojavelinux:
Merged as 40fc0fe353ebd45eea3751356c4760758ebeabcf. Thanks!

PR #4172, opened 2021-09-18
resolves #4136 document how to create a custom converter

### LINKED ISSUE #4136: Document how to create a new converter (or extend an existing one)
Although covered in the API docs, the formal documentation does not explain how to create a new converter or extend an existing one (including built-in converters).

We could consider putting this new page under Extensions. Though it might be better to add the page under a new category named Converters, and moving the Available Converters page under it too.

Here's the existing information on the topic: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter.rb#L4-L43

mojavelinux:


docs/modules/convert/pages/custom.adoc
```diff
@@ -0,0 +1,493 @@
...
+
+== Overview
+
+The converter in Asciidoctor is a specialized extension point.
+Even Asciidoctor's built-in converters use this facility.
+That means, in addition to being able to introduce a new converter, you can replace any of the existing ones.
+Since Asciidoctor is written in the Ruby programming language, you write custom converters in Ruby as well.
+
+TIP: You can also write a converter in Java using AsciidoctorJ or in JavaScript using Asciidoctor.js.
+The advantage to writing the converter in Ruby is that you can use the same code regardless of which Asciidoctor runtime you choose.
```
Mogztter, line 20: I would have written "The advantage of" but "The advantage to" might also be correct

docs/modules/convert/pages/custom.adoc
```diff
@@ -0,0 +1,493 @@
...
+That's the best strategy if you plan to share the converter with the community.
+
+When creating a custom converter, you can either write one from scratch or you can extend a built-in converter.
+You can then register that converter with a known backend to replace the previously registered converter, or you can register it with a new backend to create a new output target.
```
Mogztter, line 24: Maybe we should list the "known backend" (i.e. built-in) names?

docs/modules/convert/pages/custom.adoc
mojavelinux: You are absolutely correct. Good eye!

docs/modules/convert/pages/custom.adoc
mojavelinux: The issue there is that the known backends are any backends for which a converter has been registered. But I do agree we should at least link to https://docs.asciidoctor.org/asciidoctor/latest/converters/#built-in-converters

Mogztter (commented):
The step by step on how to write a custom converter is well written with a _gentle_ learning curve

Mogztter (approved):

PR #4180, opened 2021-10-05
resolves #3948 document how to use converter templates

### LINKED ISSUE #3948: Write documentation that explains how to customize the converter output using templates
Write documentation that explains how to create a custom backend. Information to be covered includes:
- What template languages can be used (answer: anything supported by Tilt)
- The names of each template and how the file is named
- What objects are available to the template
- Brief documenation about the structure of an Asciidoctor document object model
- Commonly used instance variables and methods in the template

Some starting material can be found here:

https://github.com/asciidoctor/asciidoctor-backends/issues/12#issuecomment-20255697

mojavelinux:


mojavelinux:
Indeed. I would say that in the end, I considered nearly all of them...even ones I initially pushed back on. If you see anything else that could use improving, feel free to click the "Edit this Page" link on the docs site and edit as you see fit.

PR #4198, opened 2021-10-30
resolves #3868 add --log-level option to CLI to control log level

### LINKED ISSUE #3868: Add option to CLI to control log level
Right now, the log level is coupled with the verbose option, so control over it is very course-grained. The log level is set to WARN by default. If the `-v` option is specified, the log level switched to DEBUG. As the logger starts to get used more across the ecosystem, it would be useful to control which log level is selected when using the CLI.

This issue proposes to add a `--log-level` option to the CLI that sets the log level explicitly. This setting overrides any implicit setting. If this option is not specified, the current behavior should be preserved.

mojavelinux:


test/invoker_test.rb
```diff
@@ -163,6 +163,38 @@
...
+    EOS
+    output = nil
+    redirect_streams do |_, err|
+      invoke_cli(%w(--log-level info), '-') { input }
+      output = err.string
+    end
+    assert_equal 'asciidoctor: INFO: possible invalid reference: install', output.chomp
+  end
+
+  test 'should not log when --log-level and -q are both specified' do
```
Mogztter, line 182: If `--log-level` and `-v` are both specified, which option should prevail?
In my opinion, `--log-level` should have a higher precedence over `-v` because `--log-level` is more explicit since you explicit state a log level (whereas `-v` is an alias for `--log-level debug`).

Alternatively, we could use the option order but not sure if this is supported (i.e., last or first declared wins).

mojavelinux: Right you are. I had programmed this logic, but forgot to add a test for it. `-q` wins out over `--log-level`, but `--log-level` wins out over `-v`.

PR #4200, opened 2021-11-05
resolves #3656 add support for link=self to image macros

### LINKED ISSUE #3656: Allow images to be linked to self without repetition
I often want to link to the image I display in my Asciidoc so the user can click to see it in it's own window at a larger size:

I might use something like this:

```asciidoc
image::autoreclose_logic_diagram.png[link="../logic_diagram/autoreclose_logic_diagram.pdf"]
```

whereas I would ideally like something along the lines of:

```asciidoc
image::autoreclose_logic_diagram.png[link=self]
```

This is unnecessarily verbose and quite a common use case (recently came up in discussion on Gitter and @mojavelinux indicated a willingness to reconsider this). I noticed the the custom converter [asciidoctor-html5s](https://github.com/jirutka/asciidoctor-html5s) offers some approaches to this (and documented them on the project README):

> Image Block
> The link attribute recognizes few special values:
> 
> `link=self`
> Make the image a link with URL of the image itself – to open it in full size.
> 
> `link=none` / `link=false`
> Suppress the effect of `:html5s-image-default-link: self`, i.e. remove the default image link.
> 
> Both block image and inline image supports additional attribute loading (see Lazy loading on MDN for more information).
> 

in particular:

> `html5s-image-default-link: self`
> Make every block image a link with the image’s source URL (i.e. user can click on the image to open it in full size), unless the link attribute is defined and is not none or false.
> 
> `html5s-image-self-link-label`
> The link title and ARIA label for the block image link that points to the image file (i.e. href equals the image’s src). Default is Open the image in full size.

Perhaps this could be considered as a non-breaking syntax extension as part of the Asciidoc specification process.

Whatever approach is taken, if this suggestion was looked into I think it should be harmonised with the approach to other language features, for instance perhaps the link macro could do with this option also (for the case where I want to show a URL which isn't an "easy" one and I want the hyperlink to be the description).

We also have other constructs with similar situations, see https://github.com/asciidoctor/asciidoctor/issues/2011 and https://github.com/asciidoctor/asciidoctor/issues/2717 

For completeness, another aspect is whether it should open in a new or existing window (see the [user manual](https://asciidoctor.org/docs/user-manual/#url)) and perhaps the caret syntax should be supported as well as the window attribute. 

I mention this because often when clicking on an image one is trying to focus on a detail or gain additional information and may not wish to lose context (so retain the existing window/tab).

So for instance:

```asciidoc
image::autoreclose_logic_diagram.png[link=self,window=_blank]
```

I can't think of a consistent approach with the link macro for the caret syntax but perhaps something like:

```asciidoc
image::autoreclose_logic_diagram.png[link=self^]
```

mojavelinux:


mojavelinux:
I've added a bunch more tests and did some prerequisite refactoring. I think this one is ready to go.

PR #4205, opened 2021-11-12
Hide autoload converter/template from Opal

Mogztter:
The latest version of Opal will resolve `autoload` module/class at compile time.
However, the built-in template converter (`converter/template.rb`) relies on Ruby template engines and, as a result, cannot be compiled by Opal.

It's no use to try to create a JavaScript compatible version of the built-in template converter because Asciidoctor.js already provides a custom implementation when running in a Node.js environment: https://github.com/asciidoctor/asciidoctor.js/blob/master/packages/core/lib/asciidoctor/js/asciidoctor_ext/node/template.rb

### DIFF
--- a/lib/asciidoctor/converter.rb
+++ b/lib/asciidoctor/converter.rb
@@ -43,7 +43,7 @@ module Asciidoctor
 #   puts Asciidoctor.convert_file 'sample.adoc', safe: :safe
 module Converter
   autoload :CompositeConverter, %(#{__dir__}/converter/composite)
-  autoload :TemplateConverter, %(#{__dir__}/converter/template)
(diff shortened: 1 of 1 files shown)


mojavelinux:
Just to clarify, you are saying that in the Asciidoctor.js environment, the TemplateConverter will already be loaded, so the code can just assume it is defined already. Is that correct?

Mogztter:
> Just to clarify, you are saying that in the Asciidoctor.js environment, the TemplateConverter will already be loaded, so the code can just assume it is defined already. Is that correct?

Yes that's correct.
The TemplateConverter is already loaded, we don't need the `autoload :TemplateConverter, %(#{__dir__}/converter/template)` statement (in an Asciidoctor.js environment).

mojavelinux:
Perfect! Thanks for clarifying.

I'll include this in the 2.0.17 release.

PR #4213, opened 2021-11-30
resolves #4212 invert FR "caution" &  "warning" translations

### LINKED ISSUE #4212: Localization: inversion in French translations
French translations for warning and caution seem inverted.

"A warning" is "un avertissement."
"Caution!" is "Attention !"

Right now, the attributes-fr.adoc contains the following:

```asciidoc
:caution-caption: Avertissement
// ...
:warning-caption: Attention
```

According to the French translations, the opposite would make more sense and be more consistent with other locales. I verified with the EN, DE and ES translations, which would all go the same way (Caution = Achtung = Attention VS Warning = Warnung = Avertissement).

cyChop:
This is just a suggestion for consistency across languages.

I'm a detail-oriented native French speaker and I'm available for exchanging on this topic if need be.
This is by no mean an urgent request and I'm sorry of how insignificant a contribution it is.

Thanks for the great tools and work!

### DIFF
--- a/data/locale/attributes-fr.adoc
+++ b/data/locale/attributes-fr.adoc
@@ -1,7 +1,7 @@
 // French translation, courtesy of Nicolas Comet <nicolas.comet@gmail.com> with updates from Maheva Bagard Laursen <mblaursen@gbif.org>
 :appendix-caption: Annexe
 :appendix-refsig: {appendix-caption}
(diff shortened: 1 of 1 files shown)


mojavelinux:
@Mogztter I'd love to get your input on this. If the two of you agree, I'm ready to proceed!

PR #4222, opened 2021-12-22
Fix typos

kianmeng:


mojavelinux:
Wow! Thanks for catching and fixing all of these! You're a :star2:!

kianmeng:
:+1: :tada:

PR #4231, opened 2021-12-29
Use recursive descent parser to implement ifeval

Light6533:
Hi, i found that your ifeval have some issue when i use plugin to build document, it will shift all the attributes at the phase preprocesser is parsing document, which makes the function Parser.parse() can't get correct attributes for ifevals to compare. So, i call Parser.process_attribute_entry() in Reader.process_line() to set attributes.
And I notice that your implement of ifeval seems incomplete, so i reference recursive descent parser of Python 3 to implement ifeval_parser.rb to support "and" "or" operator.

### DIFF
--- a/lib/asciidoctor.rb
+++ b/lib/asciidoctor.rb
@@ -570,6 +570,7 @@ require_relative 'asciidoctor/section'
 require_relative 'asciidoctor/stylesheets'
 require_relative 'asciidoctor/table'
 require_relative 'asciidoctor/writer'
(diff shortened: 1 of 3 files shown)


Light6533:
By the way, here is my sample code to verify:

:product: C

ifeval::["{product}"=="A" or "{product}"=="B"]
:XLEN: 32
endif::[]
ifeval::["{product}"!="A" and "{product}"!="B"]
:XLEN: 64
endif::[]

{product} is a {XLEN}-bit microprocessor.

PR #4235, opened 2022-01-03
html converter next low-hanging-fruit

redbow-kimee:


Mogztter:
Thanks @redbow-kimee!

It would be easier to review your contribution if each `convert_` method was in a separated pull request. Would you be able to do that?

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +224,63 @@ def convert_inline_anchor node
...
+  def convert_inline_kbd node
+    if (keys = node.attr 'keys').size == 1
+      %(<kbd>#{keys[0]}</kbd>)
+    else
+      %(<span class="keyseq"><kbd>#{keys.join '</kbd>+<kbd>'}</kbd></span>)
+    end
+  end
+
+  def convert_inline_button node
+    %(<samp class="button">#{node.text}</samp>)
```
Mogztter, line 266: I don't think `<samp>` is the right element to represent a button. MDN says:

> The [\<samp>](https://developer.mozilla.org/en-US/docs/Web/HTML) element is used to enclose inline text which represents sample (or quoted) output from a computer program.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp

Having said that, I don't have a good alternative.
Currently, we are using `<b>`... I guess we could use `<span>`?

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +224,63 @@ def convert_inline_anchor node
...
+    when :double
+      attributes = common_html_attributes node.id, node.role, 'double_quote'
+      %(<span#{attributes}>&#8220;#{node.text}&#8221;</span>)
+    else
+      %(<span#{attributes}>#{node.text}</span>)
+    end
+  end
+
+  def convert_inline_break node
+    %(#{node.text}<br />)
```
Mogztter, line 254: An open question is whether we want to support XHTML or not. And if we want to support XHTML, should we always use `/` on void elements or use an option. Currently, we have an option named `htmlsyntax` to control if we should add `/` or not.

May I ask with the extra space? `<br />` instead of `<br/>` ?

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -67,6 +67,150 @@ def convert_paragraph node
     end
   end
 
+  def convert_thematic_break node
+    "<hr />"
+  end
+
+  def convert_admonition node
```
Mogztter, line 74: I don't want to comment too much on this pull request since we need to extract each `convert_` method in a dedicated pull request.
However, I think we should open an issue where we can discuss how to implement an admonition. In https://github.com/asciidoctor/asciidoctor/issues/242 Dan said:

> Admonition blocks
Admonition blocks are rendered as a table, which makes styling them very difficult. They should be done instead as layered \<div> tags that can accommodate a wider range of layouts.

I don't necessarily agree but I think we should discuss it in an issue before jumping on the implementation.

I'm a bit unsure about `<aside>` as this element is described as: "The [\<aside>](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a portion of a document whose content is only *indirectly related to the document's main content*.".

In my opinion, admonitions are directly related to the content, so we might have to resort on using `<div>`. But let's discuss it in an issue.

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -67,6 +67,150 @@ def convert_paragraph node
...
+    end
+    node.items.each do |item|
+    sub_content = if item.blocks?
+      %(
+#{item.content}
+)
+    else
+      ""
+    end
+    ret << %(<li>#{item.text}#{sub_content}</li>)
```
Mogztter, line 102: I think we need a wrapping element if the content is compound.

redbow-kimee:
> It would be easier to review your contribution if each `convert_` method was in a separated pull request. Would you be able to do that?

Sorry for the delay. What I'll do for now is isolate each of the `convert_`ers on a branch, except for the ones with open comments. (then mark off which ones have been isolated)

lib/asciidoctor/converter/semantic_html5.rb
redbow-kimee: Yea, this one is tough, `samp` wasn't perfect, and I think `b` matches a bit less. `span class="button"` or something might be the "right" solution.

lib/asciidoctor/converter/semantic_html5.rb
redbow-kimee: Correct me if I'm wrong but XHTML5 is trivially HTML5, no? (maybe except some doctype nonsense). (I've been quite removed from web stuff for a while, so I'm probably rusty on this stuff).

As for `<br />` vs. `<br/>`, I have no opinion, other than I've always put a ` ` personally.

lib/asciidoctor/converter/semantic_html5.rb
redbow-kimee: I believe that `#{sub_content}` has a wrapper already. See [nested-unordered-list.html](https://github.com/asciidoctor/asciidoctor/pull/4235/files#diff-f75090d4025411eb5ce481ddbfa1bbe7889446c32418b2082b86280e6899a1f6). Maybe that isn't the case with all potential sub_contents?

redbow-kimee: I believe this also applys to `convert_olist` (L134) and `convert_dlist` (L167), so I'll hold off on moving them too.

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +224,63 @@ def convert_inline_anchor node
     end
   end
 
+  def convert_inline_quoted node
```
redbow-kimee, line 227: see https://github.com/asciidoctor/asciidoctor/pull/4285

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -159,6 +159,26 @@ def convert_dlist node
     ret.join LF
   end
 
+  def convert_image node
```
redbow-kimee, line 173: see https://github.com/asciidoctor/asciidoctor/pull/4287

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -67,6 +67,150 @@ def convert_paragraph node
...
+      %(<figure#{attributes}>
+#{link_start}<img src="#{target}" alt="#{encode_attribute_value node.alt}"#{size} />#{link_end}
+<figcaption>#{node.captioned_title}</figcaption>
+</figure>)
+    else
+      %(#{link_start}<img src="#{target}" alt="#{encode_attribute_value node.alt}"#{attributes}#{size} />#{link_end})
+    end
+  end
+
+  def convert_inline_image node
```
redbow-kimee, line 198: see https://github.com/asciidoctor/asciidoctor/pull/4287

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +224,63 @@ def convert_inline_anchor node
...
+    else
+      %(<span#{attributes}>#{node.text}</span>)
+    end
+  end
+
+  def convert_inline_break node
+    %(#{node.text}<br />)
+  end
+
+  def convert_inline_kbd node
```
redbow-kimee, line 257: see #4288

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +224,63 @@ def convert_inline_anchor node
...
+    else
+      %(<span class="keyseq"><kbd>#{keys.join '</kbd>+<kbd>'}</kbd></span>)
+    end
+  end
+
+  def convert_inline_button node
+    %(<samp class="button">#{node.text}</samp>)
+  end
+
+  def convert_inline_menu node
```
redbow-kimee, line 269: see #4289

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -207,5 +408,9 @@ def append_link_constraint_attrs node, attrs = []
     attrs << %( rel="#{link_types.join ' '}") unless link_types.empty?
     attrs
   end
+
+  def encode_attribute_value val
```
redbow-kimee, line 412: see #4287

lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -67,6 +67,150 @@ def convert_paragraph node
     end
   end
 
+  def convert_thematic_break node
```
redbow-kimee, line 70: see #4290

redbow-kimee:
@Mogztter I believe I've either successfully isolated each converter (in the case of images, pair of converters) to its own branch, Except for `convert_[d|o|u]list`, `convert_admonition`, `convert_inline_break`, and `convert_inline_button` which are all mentioned in threads on this MR.

lib/asciidoctor/converter/semantic_html5.rb
redbow-kimee: See #4291 for more on this.

lib/asciidoctor/converter/semantic_html5.rb
Mogztter: It's mostly about adding a slash to close tags. For instance: `<br>` is valid HTML5 but invalid XHTML.

Mogztter:
Thanks for your patience, I'm slowing merging your changes 🚋

lib/asciidoctor/converter/semantic_html5.rb
Mogztter: let's continue the discussion on https://github.com/asciidoctor/asciidoctor/pull/4321

Mogztter:
We can now close this pull request, I've extracted the remaining `convert_` methods into dedicated pull requests:

- thematic break: https://github.com/asciidoctor/asciidoctor/pull/4324
- button: https://github.com/asciidoctor/asciidoctor/pull/4323
- admonition: https://github.com/asciidoctor/asciidoctor/pull/4322
- lists (unordered lists, ordered list, description lists): https://github.com/asciidoctor/asciidoctor/pull/4321

PR #4260, opened 2022-02-26
Fix `:templates_dir` API option typo

tmazeika:
It's inconsistent with what's below and should be `:template_dirs`.


mojavelinux:
Absolutely correct. Thanks for the fix!

PR #4271, opened 2022-03-19
resolves #420 don't raise error if Asciidoctor::Extensions.unregister is called before groups are initialized

mojavelinux:


Mogztter (approved):
👍🏻

PR #4278, opened 2022-04-19
clarify the link name

awxiaoxian2020:
The “简体中文” is the common expression for "Simplified Chinese" in Chinese.

PR #4280, opened 2022-04-19
Update the link name in readme

awxiaoxian2020:
supplement for #4278. I found that in [README-fr.adoc](https://github.com/asciidoctor/asciidoctor/blob/main/README-fr.adoc), It does not display language names in the local language like other readme documents, but only in French. Is this a French convention or document style?

### DIFF
--- a/README-de.adoc
+++ b/README-de.adoc
@@ -64,7 +64,7 @@ Das Asciidoctor-Projekt wird {uri-repo}[auf GitHub] gehostet.
 
 ifndef::env-site[]
 Dieses Dokument ist auch in folgenden Sprachen erhältlich: +
-{uri-rel-file-base}README-zh_CN.adoc[汉语]
+{uri-rel-file-base}README-zh_CN.adoc[简体中文]
 |
 {uri-rel-file-base}README.adoc[English]
 |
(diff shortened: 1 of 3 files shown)


awxiaoxian2020:
@flashcode Could you give me a reply? Thank you.

mojavelinux:
It's probably just an inconsistency that crept in. I like the style that the language is its own language. That way, it is recognizable to someone looking for it.

awxiaoxian2020:
> It's probably just an inconsistency that crept in. I like the style that the language is its own language. That way, it is recognizable to someone looking for it.
I fully share your opinion. But I think it's better to be confirmed by contributors in French. I would like to keep it open until there are any reply.

flashcode:
Hi,

I agree with @mojavelinux, the name of each language should be translated in this language.

mojavelinux:
Thanks again!

awxiaoxian2020:
> Hi,
> 
> I agree with @mojavelinux, the name of each language should be translated in this language.

OK, thank you.

awxiaoxian2020:
@mojavelinux  sorry, I made it draft as I have not changed the French doc already, I think I will have a new PR.

mojavelinux:
I'm merging them separately because each change should be focused on a specific task. We have now updated the Simplified Chinese references, which was the first task. Changing French is a separate task. Thanks again for helping out!

awxiaoxian2020:
> I'm merging them separately because each change should be focused on a specific task. We have now updated the Simplified Chinese references, which was the first task. Changing French is a separate task. Thanks again for helping out!

Got it, thanks!

PR #4285, opened 2022-05-02
html-converter-next inline_quotes

redbow-kimee:
bold, italiic, quotes, superscripts, etc

### DIFF
--- a/lib/asciidoctor/converter/semantic_html5.rb
+++ b/lib/asciidoctor/converter/semantic_html5.rb
@@ -108,10 +108,10 @@ class Converter::SemanticHtml5Converter < Converter::Base
     when :subscript
       %(<sub#{attributes}>#{node.text}</sub>)
     when :single
-      attributes = common_html_attributes node.id, node.role, 'single_quote'
+      attributes = common_html_attributes node.id, node.role, 'singlequote'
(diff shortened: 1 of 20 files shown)


lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +80,32 @@ def convert_inline_anchor node
...
+    when :monospaced
+      %(<code#{attributes}>#{node.text}</code>)
+    when :mark
+      %(<mark#{attributes}>#{node.text}</mark>)
+    when :superscript
+      %(<sup#{attributes}>#{node.text}</sup>)
+    when :subscript
+      %(<sub#{attributes}>#{node.text}</sub>)
+    when :single
+      attributes = common_html_attributes node.id, node.role, 'single_quote'
```
Mogztter, line 99: Not sure about this default role. If we want to be consistent, then it should be `singlequote` and `doublequote`: https://github.com/asciidoctor/asciidoctor/blob/feature/html-converter-next/devdocs/modern-html-converter/built-in-roles.adoc

redbow-kimee: fixed with 21852ae

Mogztter (approved):
👍🏻

PR #4288, opened 2022-05-02
html-converter-next keyboard text

redbow-kimee:


Mogztter (approved):
I think this is fine, I don't have a counter proposal and can't see what else we could use.
I will let this pull request opens a few days before merging to give everyone a chance to review.

PR #4289, opened 2022-05-02
add inline menu method to the semantic HTML5 converter

redbow-kimee:


lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -80,6 +80,21 @@ def convert_inline_anchor node
     end
   end
 
+  def convert_inline_menu node
+    caret = (node.document.attr? 'icons', 'font') ? '&#160;<i class="fa fa-angle-right caret"></i> ' : '&#160;<b class="caret">&#8250;</b> '
```
Mogztter, line 84: This an interesting one!
In my opinion, we should not rely on the icons provider. We can still use an icon but we should not assume that  FontAwesome is the icon provider. In my mind, the icons provider should be used primarily for the `icon:` macro.

In this case, we could use unicode `›`?

PR #4290, opened 2022-05-02
implement thematic break in the semantic HTML5 converter

redbow-kimee:


lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -67,6 +67,10 @@ def convert_paragraph node
     end
   end
 
+  def convert_thematic_break node
+    "<hr />"
```
Mogztter, line 71: I think we should use `<hr>` (at least for now).
In my opinion, we should open an issue to discuss if we want to support XHTML in the new converter or not.

redbow-kimee: fixed with 0861630

Mogztter:
I've created https://github.com/asciidoctor/asciidoctor/issues/4309 as a follow-up issue.

test/fixtures/semantic-html5-scenarios/thematic-break.adoc
```diff
@@ -0,0 +1,5 @@
+this is a first paragraph.
```
Mogztter, line 1:
```suggestion
This is a first paragraph.
```

test/fixtures/semantic-html5-scenarios/thematic-break.html
```diff
@@ -0,0 +1,7 @@
+<p>
+this is a first paragraph.
```
Mogztter, line 2:
```suggestion
This is a first paragraph.
```

PR #4293, opened 2022-05-10
rouge: add AL install step

noraj:


mojavelinux:
Thanks for making this addition!

PR #4294, opened 2022-05-26
Add SciTE to the list of text editors

georgeraraujo:


mojavelinux:
I'm open to adding this editor to this directory. However, I don't want to include details about how to activate AsciiDoc support for older versions. That's just one more thing I have to maintain. As stated in this update:

> already bundled with SciTE 5.2.3 and later

So the entry can simply state that it requires using SciTE 5.2.3 and later. If you reduce it down to just that, I'll merge it.

georgeraraujo:
Done.

mojavelinux:
Looks good. I made a few minor edits, but the meaning is still the same.

PR #4314, opened 2022-07-18
resolves #2284 warn if include target is remote and allow-uri-read attribute is not set

mojavelinux:


Mogztter:
#2284 is a pull request

mojavelinux:
Oops. I was working in the wrong repository. Let's try this again.

PR #4321, opened 2022-07-30
add convert list methods to the semantic HTML 5 converter

Mogztter:
original work: @redbow-kimee


lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -96,6 +96,81 @@ def convert_thematic_break node
...
+    ret << %(<strong class="title">#{node.title}</strong>) if node.title?
+    node.items.each do |item|
+    sub_content = if item.blocks?
+      %(
+#{item.content}
+)
+    else
+      ""
+    end
+    ret << %(<li>#{item.text}#{sub_content}</li>)
```
Mogztter, line 111: Do we need a wrapping element around `item.text`?
Currently, we have:

```html
<ul>
<li>alice
<ul>
<li>subpoint 1</li>
<li>subpoint 2</li>
</ul>
</li>
</ul>
```

With the following structure, the text "alice" cannot be easily styled. So, maybe we should do:

```html
<ul>
<li><span class="listitem">alice</span>
<ul>
<li>subpoint 1</li>
<li>subpoint 2</li>
</ul>
</li>
</ul>
```

I believe it's unnecessary to add a class name on the nested `<ul>` since you can use the CSS selector `ul ul` to target nested lists.

mojavelinux: > Do we need a wrapping element around item.text?

Yes. It's absolutely essential. Otherwise, there is no way to target it with CSS, and that only gets us back into the dilemma we have with the current HTML of not being designer friendly.

mojavelinux: > I believe it's unnecessary to add a class name on the nested <ul> since you can use the CSS selector ul ul to target nested lists.

We need at least one role on any top-level element the converter creates to make it clear that the converter made that element. Otherwise, we can't make a distinction between lists created from an AsciiDoc list or a list created from some other source. It's not enough to assume that CSS can target it.

mojavelinux: > `<span class="listitem">alice</span>`

Since the whole element is already a list item, I prefer describing the part of it. That's why I like "principle" like in the EPUB 3 converter. I'd even consider "text" since that's what we call it in the Asciidoctor DOM.

Mogztter: > Otherwise, we can't make a distinction between lists created from an AsciiDoc list or a list created from some other source. It's not enough to assume that CSS can target it.

So `<ul class="list">` and `<span class="text">alice</span>`?
`principle` might be more semantic but less common (and difficult to remember).

Speaking of naming, I wonder if we should rename `menuitem` to `item` (on the `menu` macro) since the element is already in a "menu" element? Otherwise, someone could argue that we should have use `listtext` (instead of `text`) to be consistent.

mojavelinux: > So `<ul class="list">` and `<span class="text">alice</span>`?

I would prefer matching the AsciiDoc block context names, in which case it would be `ulist` on the `<ul>` tag.

There's one other way to look at it. We could say that `<ul>` already tells us the context in this case. Instead, we add the common/generic identifier `block`. The reason for this is to be able to style all block elements consistently, like we see here: https://gitlab.com/antora/antora-ui-default/-/blob/master/src/css/doc.css#L200-217 As you can see, having different names without a common class is annoying. So perhaps:

```
<ul class="block">
```

That would also mean for paragraphs, we use:

```
<p class="block">
```

> I wonder if we should rename menuitem to item (on the menu macro) since the element is already in a "menu" element?

That's a good point. There's really no right or wrong way here, just what feels most comfortable. What's most important is that we're consistent about it.

Mogztter: > There's one other way to look at it. We could say that `<ul>` already tells us the context in this case. Instead, we add the common/generic identifier block

I see, that's a good idea but what about `<figure>` used for image and source blocks? In this case, we will need to make a distinction using the block name.
As an aside, and speaking of source blocks, we are currently using `listing` even if it's a source block. I think we should use `source`. 

> As you can see, having different names without a common class is annoying

Why not...

```html
<ul class="block"></ul>
<ol class="block"></ol>
<dl class="block"></dl>
<dl class="qanda block"></dl>
<dl class="horizontal block"></dl>
<figure class="source block"></figure>
<figure class="image block"></figure>
<figure class="listing block"></figure>
```

The thing is `block` is a common name and if we really want to make a clear distinction between elements created from Asciidoctor and elements created from some other source then we should probably use a more unique name: `ablock`, `adocblock`, `asciidoctorblock`.
We could do the same for inline elements: `ainline`, `adocinline`, `asciidoctorinline`?

Mogztter: On a related note, we could also add a common/generic class on headings? https://gitlab.com/antora/antora-ui-default/-/blob/master/src/css/doc.css#L21-26

It's less annoying to declare, because h1, h2, h3... is known, but if you want to make sure that you are only apply a style on headings created by Asciidoctor then it might be useful.

mojavelinux: > As an aside, and speaking of source blocks, we are currently using listing even if it's a source block. I think we should use source.

That seems reasonable. The distinction in AsciiDoc has always been fuzzy, so I'm open to exploring making it clearer. (I'd even be happy with "code" instead of "source" like in Asciidoctor PDF).

> The thing is block is a common name

I think we need to be careful not to go overboard. The idea is that we are trying to be as concise as possible without making it hard to style. I think a generic name is the right approach. Once we start adding "AsciiDoc" as a prefix, it's just going to get ugly. We can explore other generic names, but right now "block" seems to be comfortable to me.

mojavelinux: > On a related note, we could also add a common/generic class on headings?

I'd be okay with that. I like just "heading" as a name here.

Btw, when in doubt, we should look at how some CSS frameworks are naming classes. That might lead us in the right direction when we get lost.

redbow-kimee: How about rather than `block` or `adocblock` as the catchall class, just `adoc`, concise and specific. to take the example from above

```
<ul class="adoc"></ul>
<ol class="adoc"></ol>
<dl class="adoc"></dl>
<dl class="qanda adoc"></dl>
<dl class="horizontal adoc"></dl>
<figure class="source adoc"></figure>
<figure class="image adoc"></figure>
<figure class="listing adoc"></figure>
```

mojavelinux: That's a very strong suggestion. I'm going to give that some consideration.

The only drawback is that it's not clear that it's a block rather than an inline element. And block elements almost always need to share styles.

We could consider "ablock" and "ainline" (or "aspan"). But even as I type that, remember we still have a root element we can use to scope styles like we do in the stylesheet for the default UI. That's where we could use the name "asciidoc".

Mogztter: > But even as I type that, remember we still have a root element we can use to scope styles like we do in the stylesheet for the default UI. That's where we could use the name "asciidoc" (e.g., ".asciidoc .block").

That does not address your initial concern since it's possible to create an element within the `.asciidoc` root element via JavaScript.

For reference, Bulma is using the class `block` and does not add any class on inline elements. As soon as an inline element is within a `content` element it gets styled accordingly.

[Materialize CSS](https://materializecss.com/getting-started.html), [Foundation ](https://get.foundation/sites/docs/typography-base.html), [Bootstrap](https://getbootstrap.com/docs/5.2/content/typography/) and [Skeleton](http://getskeleton.com/#intro) are not using any class on inline elements.

Semantic UI seems to be the only one to require a `ui` class on every elements: https://semantic-ui.com/introduction/getting-started.html

[UIKit](https://getuikit.com/docs/introduction) is using the `uk` prefix:  `uk-card`, `uk-container`... 

Brad frost recommends to use a global namespace:

> All classes associated with the design system are prefixed with a global namespace, which is the Company Name followed by a hyphen
> If you’re working on a CSS architecture that is only meant to be served to a single site or if you have a lot of control over your environment, including a global namespace is likely unnecessary. But if your design system is intermingling with other technologies, it might make sense to create an identifier for system-specific code. 

https://bradfrost.com/blog/post/css-architecture-for-design-systems/

A similar approach is to use namespaces/prefixes to communicate exactly what a class does, why it exists, where (else) it might already occur, whether or not we can reuse it elsewhere, and how safe it is to bind onto or modify. 

Here's an excerpt:

> `o`-: Signify that something is an Object, and that it may be used in any number of unrelated contexts to the one you can currently see it in. Making modifications to these types of class could potentially have knock-on effects in a lot of other unrelated places. Tread carefully.
> `c-`: Signify that something is a Component. This is a concrete, implementation-specific piece of UI. All of the changes you make to its styles should be detectable in the context you’re currently looking at. Modifying these styles should be safe and have no side effects.
> (...)

https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/

Since it exceeds the scope of this pull request, I think we should discuss that in a dedicated issue to decide which naming convention/CSS principles we want to use. What do you think?

Mogztter: follow-up issue: https://github.com/asciidoctor/asciidoctor/issues/4349

mojavelinux: In the AsciiDoc Language specification, we have decided to describe the main text of a list item as the principal text (or principal, for short). So I still think "principal" is the right word here. principal is defined as "the most important" and that's precisely what it is.

ggrossetie:
In summary:

```html
<ul class="list">
<li><span class="listitem">alice</span>
<ul>
<li>subpoint 1</li>
<li>subpoint 2</li>
</ul>
</li>
</ul>
```

Should we also add the `<span>` element when there's a nested list? (for consistency)

```html
<ul class="list">
<li><span class="listtext">alice</span>
<ul>
<li><span class="listtext">subpoint 1</span></li>
<li><span class="listtext">subpoint 2</span></li>
</ul>
</li>
</ul>
```

Regarding, description list should we use a `desctext` class name?

```html
<dl class="list qanda">
<dt>Question 1?</dt>
<dd><span class="desctext">Answer 1.</span></dd>
</dl>
```

redbow-kimee:
@ggrossetie  Regarding `<span>` within a nesting list. I'm not sure why we need `<span>`s at all? Unless, is it to target just the text of the li itself without going into the subpoint's text. In that case, if there were a sub-subpoint we should have a span on the subpoint text to target that text without the text of the sub-subpoint?. Maybe we should have spans all the way down? although somebody could write their CSS to understand nesting levels.

```css
/* For example if somebody wanted most fonts to be sans, but first level LIs to be serif, and subseqent nesting levels to go back to sans. Idk why anybody would want to change font family in particular, but I'm uncreative with my examples */

{
  font-family: sans;
}

li {
  font-family: serif;
}

li li {
  font-family: sans;
}
```

My personal preference would be to go with a system like this in CSS rather than adding extra spans and what not to facilitate a different system for writing CSS. But I understand that my opinions on CSS may differ from others, and that the spans don't prevent me from writing CSS in the shown manor. (They do cost extra characters on download, to the extent that actually matters with compression these days).

redbow-kimee:
Regarding `class='desctext'`, I don't see what that offers over already being inside a `dd` tag?

redbow-kimee:
How should we handle titles in lists. Currently we get

```
.List title
* List item 1
* List item 2
```

```
<ul>
<strong class="title">List title</strong>
<li>List item 1</li>
<li>List item 2</li>
</ul>
```

But that leaves us with <strong> on its own inside the <ul>. Most browsers display the <strong> as a block tag (although I believe strong is an inline tag). Although CSS can fix this (`ul>strong.title:first-child { display: block; }`) to sort of force our intended behavior in the case of a browser that doesn't do what we expect it to. Is this the way to go?

mojavelinux:
> Regarding `<span>` within a nesting list. I'm not sure why we need `<span>`s at all? 

From my experience styling HTML generated from AsciiDoc, I can say for certain that the `<span>` is crucial. Without it, there is no way to target the text of a list item in CSS without effecting other content. And without it, there would be no way to use the before and after pseudo-selectors, which are an important tool when styling list items. So this tag must remain.

Note that this tag is replacing a `<p>` in the output of the current HTML converter. There have been numerous complains about this because it adds extra space around each list item when no stylesheet is used, or no dedicated stylesheet is used (like on GitHub). The `<span>` is an ideal compromise.

Keep in mind that it's not poor semantics to add additional tags, especially div and span. That thread traces all the way back to the CSS Zen Garden days. If there an insufficient amount of tags, it makes the designer have to resort to using a lot of hacks, which is what we're trying to get away from.

What I'm not yet comfortable with is the name of the CSS class on it. I think that "listtext" is too specific when "text" would do. In CSS, you want to avoid redundant words in the hierarchy. If the class were just text, it could be targeted precisely using the `li > span.text:first-child` selector. I think we should try that until it's not sufficient.

mojavelinux:
I want to point out that in the spec, we are formalizing the name of the text of a list item, which precedes any attached blocks. The name is "principal text". That's why I still insist that the class be "principal" (as in span.principal) rather than "listtext" (or even "text" as I wrote above). Though since the formal term is "principal text" and the hierarchy suggests that it is principal based on its location, I would be willing to settle on "text".

mojavelinux:
> How should we handle titles in lists. 

After reading some other threads in which the block title came up, I'm becoming convinced that it needs to be wrapped in a `<div>`. The `<strong>` is there since we must accept the fact that HTML generated from AsciiDoc is going to be rendered in places that will never provide a proper stylesheet, such as on GitHub. The addition of `<strong>` both provides an important default and extra emphasis that it's a title. (I'd be willing to go with `<em>` as an alternative for this reason).

But the fact that we're using an inline tag, yet treating it as a block, I don't think is sufficient. I think it needs to be inside of a block tag like `<div>`. (I've also seen `<h3>` used for this purpose, but I've never been quite convinced about the validity of that choice). The other alternative is to wrap any block element with a title in `<figure>` with a `<figcaption>`. However, I can just hear people now complaining about that being too elaborate.

Again, keep in mind that we don't need to be austere with tags. What we're shooing for is something that looks readable with no stylesheet (or a deficient one like on GitHub) while at the same time being easy to customize. There are a lot of concerns that the current HTML is awkward to work with. If the structure we choose does not feel awkward, we will have achieved what we set out to achieve. There is no such thing as perfect HTML (even though many of probably yearn for such a thing, as I certainly do).

redbow-kimee:
I can give a go at working up something like this after work.

```
<ul>
<div role="heading"><strong class="title">List title</strong></div>
<li><span class="principal">List item 1</span></li>
<li><span class="principal">List item 2</span></li>
</ul>
```

I think [`role="heading"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role) might be the correct way to promote a non `<Hn>` to a heading of imprecise number. Although it has a friend tag `aria-level="n"` for the heading number which defaults to 2, so it might just default to being an `<h2>` tag.

redbow-kimee:
the `<header>` tag might actually be appropriate for titles too. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header.

mojavelinux:
I'd be open to exploring both ideas.

redbow-kimee:
I'm going to go for `<header>` for the time being, to avoid potential conflation with section titles.

redbow-kimee:
Okay, see work here https://github.com/ggrossetie/asciidoctor/pull/7

ValorNaram:
Mhh, a bit of present researching and my memories from my past researches about list titles for `<ul>`, `<ol>` and the `<dl>` one or two years ago lead me to using `aria-labelledby`. Reference:

- https://govtnz.github.io/web-a11y-guidance/wct/lists/make-a-list-accessible.html#aria-labelledby-when-a-visible-label-is-available
- http://kb.daisy.org/publishing/docs/html/lists.html#ex-02

Unfortunately there is no standard accessible design pattern for giving lists an accessible title. But the following solution derived from @redbow-kimee's [example](https://github.com/asciidoctor/asciidoctor/pull/4321#issuecomment-1927177575) with `aria-labelledby` gives the list an accessible name:

```html
<div id="ul-heading0"><strong class="title">List title</strong></div>
<ul aria-labelledby="ul-heading0">
<li><span class="principal">List item 1</span></li>
<li><span class="principal">List item 2</span></li>
</ul>
```

What I did:
- Moving `div role="heading"` out of the `<ul>` element: It is no longer a child of `<ul>`
- Removing the role from the `<div>` element
- Giving `div role="heading"` an id which can be computed by Asciidoctor.
- Assigning the attribute `aria-labelledby` to the `<ul>` list.
- Referencing the `id` given to  the `<div>` as value to `aria-labelledby` attribute of the `<ul>`

Problem: You need to generate an id which doesn't conflict. Maybe we can make this more unlikely if we append `asciidoctor-<title of document>` a prefix to each id.

PR #4322, opened 2022-07-30
add convert admonition method to the semantic HTML 5 converter

### LINKED ISSUE #4291: tag selection for admonitions
> I don't want to comment too much on this pull request since we need to extract each `convert_` method in a dedicated pull request. However, I think we should open an issue where we can discuss how to implement an admonition. In #242 Dan said:
> 
> > Admonition blocks
> > Admonition blocks are rendered as a table, which makes styling them very difficult. They should be done instead as layered <div> tags that can accommodate a wider range of layouts.
> 
> I don't necessarily agree but I think we should discuss it in an issue before jumping on the implementation.
> 
> I'm a bit unsure about `<aside>` as this element is described as: "The [<aside>](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents a portion of a document whose content is only _indirectly related to the document's main content_.".
> 
> In my opinion, admonitions are directly related to the content, so we might have to resort on using `<div>`. But let's discuss it in an issue.
>
>  -- @Mogztter (https://github.com/asciidoctor/asciidoctor/pull/4235/files#r839397957)

Personally, I have little opinion on `aside` or `div` or something else, so long as it isn't `table`.

(also, please tag this as `area/html-next`, I don't have permission to tag)

Mogztter:
original work: @redbow-kimee

resolves #4291


lib/asciidoctor/converter/semantic_html5.rb
```diff
@@ -96,6 +96,17 @@ def convert_thematic_break node
     '<hr>'
   end
 
+  def convert_admonition node
+    attributes = common_html_attributes node.id, node.role, %(admonition #{node.attr 'name'})
+    ret = []
+    ret << %(<aside#{attributes}>)
+    ret << %(<em class="label">#{node.attr 'textlabel'}</em>)
+    ret << %(<strong class="title">#{node.title}</strong>) if node.title?
```
ggrossetie, line 104: I do not completely agree with the use of `<em>` and `<strong>`. Why emphasize the label and give a strong importance to the title? Shouldn't we be more neutral and use `<span>`? If we consider that an admonition is a section, then we could possibly use `<h6>` for the title?

@mojavelinux any opinion?

mojavelinux: I'm not opposed to applying strong emphasis (i.e., strong) to the label as that is likely what most people we want to see anyway. But I think we may need some `<div>` elements here or this is going to be very difficult to style. The title should follow whatever we are doing for other block elements.

One idea would be to see if you can write some CSS for this markup and figure out where the pitfalls out. If it can't be styled easily, we haven't done any better than what we have in the existing converter.

ggrossetie: > One idea would be to see if you can write some CSS for this markup and figure out where the pitfalls out. If it can't be styled easily, we haven't done any better than what we have in the existing converter.

That's a good point. I think we need to add a `<div>` around the content.
Another interesting question is whether we want to produce a different HTML when we specify `:font: icons`.

Using a pseudo-element (`::before` or `::after`) makes it difficult to apply a different style between the element and the pseudo-element, especially if we want to use a different padding or combine a background-color with an icon declared as a background-image.

mojavelinux: > Another interesting question is whether we want to produce a different HTML when we specify :font: icons.

First, I would definitely mark the whole block when icons are enable so CSS can be written to handle the case when it is vs when it isn't.

I don't think there's any right or wrong HTML, but the HTML should feel like it is easy to style. If that means more elements are needed to avoid lots of hacks, I think that's a good thing.

ggrossetie: I'm thinking something like:

```html
<strong class="label"><span class="icon"></span><span class="text">Tip</span></strong>
```

Depending on the `icons` option we can add the corresponding block inside `<span class="icon"></span>`:

```html
<span class="icon"><i class="fa icon-tip" title="Tip"></i></span>
```

```html
<span class="icon"><img src="./images/icons/tip.png" alt="Tip"/></span>
```

> First, I would definitely mark the whole block when icons are enable so CSS can be written to handle the case when it is vs when it isn't.

👍🏻 

```html
<aside class="admonition tip has-icon">
```

mojavelinux: In my opinion, it's pretty much universally accepted that the `i` element has been repurposed to mean icon, so let's use it.

And remember, we are not going to use `<aside>` since that is technically not a legal use of the tag (see discussion in issue).

ggrossetie: > In my opinion, it's pretty much universally accepted that the i element has been repurposed to mean icon, so let's use it.

Do you mean removing the `<span>` wrapper element? I think we should keep it otherwise it's harder to style the icon independently from the text.

Having said that, I agree that "i" is universally accepted as "icon" hence:

```html
<span class="icon"><i class="fa icon-tip" title="Tip"></i></span>
```

ggrossetie: It also allows a common HTML structure that's ready for what's come next:

#### icon: none

```html
<span class="icon"></span>
```

#### icon: svg

```html
<span class="icon"><svg></svg></span>
```

mojavelinux: Yep, I'm good with that. I was just making a clarifying comment in support of the i tag.

Issue #4310, opened 2022-07-07
Upgrade nokogiri (development dependency) to 1.13.x


Mogztter:
We might be able to remove the `BUNDLE_BUILD__NOKOGIRI=--use-system-libraries` environment variable in:

https://github.com/asciidoctor/asciidoctor/blob/cd9b07c75323be97bb8364554f9a4ee095baec33/.github/workflows/ci.yml#L76

I believe we are using `--use-system-libraries` because of this issue: https://github.com/flavorjones/mini_portile/issues/101
Since this issue is now fixed in version 2.5.2 and Nokogiri 1.13.x is using version 2.8+ of mini_portile, I think it would make sense to upgrade.

Initial discussion: https://asciidoctor.zulipchat.com/#narrow/stream/279656-dev-.F0.9F.9B.A0.EF.B8.8F/topic/bundle.20install.20in.20asciidoctor.20with.20Ruby.203.2E1.20.28on.20Ubuntu.29

mojavelinux:
The main reason we're using `--use-system-libraries` is because it's takes an excruciatingly long time to rebuild nokogiri. By linking instead, it speeds up installed dramatically.

mojavelinux:
The problem with upgrading is that it requires Ruby >= 2.6 and we test back to Ruby 2.3.

We are reevaluating how we drop old versions of Ruby, but suffice to say, core has to be prudent with the minimum Ruby version to support all the Linux distributions and macOS. So I don't think we can make this jump yet.

What exactly is the reason for wanting to upgrade? All the tests are passing in CI right now.

Mogztter:
As an intermediate step we could upgrade to Nokogiri 1.12 which uses mini_portile 2.6.1 and requires Ruby 2.5.

The main reason is that it does not install on Ubuntu 20.4 LTS when building from source (i.e. without `--use-system-libraries`).
It's not a big deal but we might want to add a note in the contributing guide.

mojavelinux:
This was resolved by #4339.

PR #4339, opened 2022-08-23
upgrade Nokogiri to latest version and add overrides for older Ruby versions

flavorjones:
AsciiDoctor currently requires Nokogiri `~> 1.10.0`, but Nokogiri v1.10 has been out of support since January 2021, and this requirement may [prevent AsciiDoctor tests from running on Ruby 3.1](https://github.com/flavorjones/mini_portile/issues/101#issuecomment-1223072639).

This PR updates the gemspec to allow for modern versions of Nokogiri.

A side benefit is that Nokogiri >= 1.11 ships [precompiled native gems](https://nokogiri.org/index.html#native-gems-faster-more-reliable-installation) for most platforms.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,8 +16,10 @@ group :development do
   gem 'pygments.rb', ENV['PYGMENTS_VERSION'] if ENV.key? 'PYGMENTS_VERSION'
   # rouge is needed for testing source highlighting; Asciidoctor supports rouge >= 2
   gem 'rouge', (ENV.fetch 'ROUGE_VERSION', '~> 3.0')
-  if Gem.win_platform? && (Gem::Version.new RUBY_VERSION) >= (Gem::Version.new '2.6.0')
-    gem 'nokogiri', '~> 1.13.0'
+  if RUBY_ENGINE == 'truffleruby' || (Gem::Version.new RUBY_VERSION) < (Gem::Version.new '2.5.0')
+    gem 'nokogiri', '~> 1.10.0'
+  elsif (Gem::Version.new RUBY_VERSION) < (Gem::Version.new '2.6.0')
+    gem 'nokogiri', '~> 1.12.0'
   end
 end
 
--- a/asciidoctor.gemspec
+++ b/asciidoctor.gemspec
@@ -40,7 +40,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'erubi', '~> 1.10.0'
   s.add_development_dependency 'haml', '~> 5.2.0'
   s.add_development_dependency 'minitest', '~> 5.14.0'
-  s.add_development_dependency 'nokogiri', '~> 1.10.0'
+  s.add_development_dependency 'nokogiri', '~> 1.13.0'
   s.add_development_dependency 'rake', '~> 12.3.0'
   s.add_development_dependency 'slim', '~> 4.1.0'
   s.add_development_dependency 'tilt', '~> 2.0.0'


Mogztter:
Related issue: https://github.com/asciidoctor/asciidoctor/issues/4310

> The problem with upgrading is that it requires Ruby >= 2.6 and we test back to Ruby 2.3.
We are reevaluating how we drop old versions of Ruby, but suffice to say, core has to be prudent with the minimum Ruby version to support all the Linux distributions and macOS. So I don't think we can make this jump yet.

KartikSoneji:
> The problem with upgrading is that it requires Ruby >= 2.6 and we test back to Ruby 2.3.

Hmm my understanding was `nokogiri  ~> 1.10` automatically select the highest nokogiri  version above 1.10 that was compatible with the current ruby version?

Mogztter:
> ~> 2.0.3 is identical to >= 2.0.3 and < 2.1. ~> 2.1 is identical to >= 2.1 and < 3.0

https://guides.rubygems.org/patterns/#pessimistic-version-constraint

Currently, we are using `~> 1.10.0` which is equivalent to `>= 1.10.0` and `< 1.11` (i.e., take the latest patch version available)

flavorjones:
Ah, I see, I wasn't aware of the testing matrix. I think we can loosen it without breaking compatibility by updating to `~> 1.10`, I'll update this PR.

mojavelinux:
Since this is effectively the 2.1.x branch, we can revisit the minimum version of Ruby we support. Currently, that is Ruby 2.5 based on what is still being used by supported versions of Linux. That would allow us to set the version of Nokogiri as follows:

```
~> 1.11.0
```

(This will select 1.11.7, which requires Ruby >= 2.5)

We can also change the minimum version of Ruby in .github/workflows/ci.yml

I'm baffled why nokogiri is being so overly-specific about versions of Ruby that they support. It's making using the library very difficult. Fortunately, it's only for tests, so we could set per-Ruby version overrides in Gemfile.

Btw, we know Asciidoctor works fine with Ruby 3.1 because we test on it in CI. See https://github.com/asciidoctor/asciidoctor/blob/main/.github/workflows/ci.yml#L28

flavorjones:
> I'm baffled why nokogiri is being so overly-specific about versions of Ruby that they support

Nokogiri drops support for versions of Ruby that are more than a year past end-of-life. I don't think this is an unreasonable policy. If you feel differently, please open an issue and we can discuss it!

flavorjones:
I'm happy to update this PR however you want -- however changing the CI matrix feels like a policy decision that maybe could be visited in a separate PR by the project maintainers?

Also feel free to close this without merging if it's not something you're interested in, no worries! Thanks for your time.

KartikSoneji:
> Btw, we know Asciidoctor works fine with Ruby 3.1 because we test on it in CI. See https://github.com/asciidoctor/asciidoctor/blob/main/.github/workflows/ci.yml#L28

Hmm. I'm getting an error on Windows, `ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x64-mingw-ucrt]`.

Also, [the changelog](https://github.com/ruby/ruby/blob/fb4df44d1670e9d25aef6b235a7281199a177edb/NEWS.md#stdlib-updates) specifies `net/ftp` was moved to a bundled gem.
How does that work with the current Gemfile?
image

mojavelinux:
I find it very hostile to Linux packaging to use hard version ranges, especially ones that are so tight. The reason we are in the situation we are in is because I got to great lengths to make it easy on Linux packages to create packages for Asciidoctor gems. The biggest problem I ran into is the premature dropping of older versions of Ruby. Just because the Ruby project marks a version of Ruby as EOL doesn't mean it is not supported by a Linux vendor. I have to play nice with them or else my gem won't be packaged. And they aren't going to move any faster.

I get saying that you need Ruby >= 2.5 or Ruby >= 2.7, but having the version range be Ruby >= 3.1 <= 3.2.dev is overkill.

mojavelinux:
> A side benefit is that Nokogiri >= 1.11 ships [precompiled native gems](https://nokogiri.org/index.html#native-gems-faster-more-reliable-installation) for most platforms.

This has bee a feature well before 1.11.

flavorjones:
> having the version range be Ruby >= 3.1 <= 3.2.dev is overkill

Ah, this pattern is used ONLY in  the precompiled native libraries (and the ranges are broader than you're representing, generally the last 4 ruby minors). The "ruby" platform library for Nokogiri will only give a minimum Ruby version. And modern bundler is smart enough to sort this all out for you.

flavorjones:
e.g., https://rubygems.org/gems/nokogiri/versions/1.13.8 supports ruby `>= 2.6.0` but https://rubygems.org/gems/nokogiri/versions/1.13.8-x86-linux supports ruby `>= 2.6, < 3.2.DEV` because those are the versions that were precompiled.

mojavelinux:
> e.g., rubygems.org/gems/nokogiri/versions/1.13.8 supports ruby >= 2.6.0 but rubygems.org/gems/nokogiri/versions/1.13.8-x86-linux supports ruby >= 2.6, < 3.2.DEV because those are the versions that were precompiled.

My mistake. I misread the metadata on rubygems.org. I now see it is Ruby >= 2.6 for the latest release.

flavorjones:
> > A side benefit is that Nokogiri >= 1.11 ships [precompiled native gems](https://nokogiri.org/index.html#native-gems-faster-more-reliable-installation) for most platforms.
>
> This has bee a feature well before 1.11.

Nokogiri < 1.11 only shipped precompiled for Windows. 1.11 and later also ships precompiled for linux and macos of various architectures.

mojavelinux:
We have been planning to up the minimum version of Ruby to 2.5 in this branch. I'll apply that change separately. I'll also reenable the window-latest 3.1 job (and make any updates necessary to ensure that works).

We can then proceed with the change in this PR to set the version of nokogiri to ~> 1.11.0. If you'd like, you can wait until I've made those other updates to update the PR.

mojavelinux:
> Nokogiri < 1.11 only shipped precompiled for Windows. 1.11 and later also ships precompiled for linux and macos of various architectures.

Good to know. And definitely a welcomed improvement for users.

mojavelinux:
It's important to keep in mind that nokogiri is only used in the test suite for this gem. None of this affects users of Asciidoctor. We only need to consider how it impacts what versions of Ruby we can test.

I have used exceptions for old versions of Ruby in the Gemfile in the past. If possible, I like to avoid doing that as it requires more maintenance. But ultimately I'll do what is necessary.

flavorjones:
Sure, hey, if it's easier to keep the gemspec as-is I won't feel bad if y'all close this without merging. I apologize if this has been an unwanted distraction.

KartikSoneji:
> However, I'm confused why it would not work on Windows but would work on Linux. The unbundling of gems hasn't been a smooth transition, but we'll get through it.

Maybe it has something to do with this?
https://github.com/asciidoctor/asciidoctor/blob/d72fe54346aa6963ae835f48681fa1ed917fb3fe/.github/workflows/ci.yml#L75-L76
`require mini_portile2` might not be called at all while building nokogiri.

mojavelinux:
I bumped the minimum version for Ruby to 2.5 and reenabled the Ruby 3.1 on Windows job. I'm now attempting to bump the version of Nokogiri, but there seems to be some unrelated issues. Whenever we touch the CI jobs, it's always a bit of a wrestling match, so just bear with me ;)

mojavelinux:
For context, one of the reasons I'm always tentative to upgrade Nokogiri is because it usually ends up taking me a couple of hours to figure out how to get it to run on our full matrix. I just discovered that the precompiled binaries are not available for TruffleRuby, which is one of the Ruby implementations we have to test on. I appreciate how hard it is to support the full spectrum of Ruby implementations since I have to go through it as well. But the fact remains that it takes a lot of time and attention to make this upgrade.

mojavelinux:
It looks like we have something that's going to work. We need some overrides, but those should only ever be used in CI.

mojavelinux:
Thanks again!

PR #4346, opened 2022-08-28
resolves #4345 fix position on positional attributes

### LINKED ISSUE #4345: The position on positional attributes should ignore named attributes
According to the documentation:

> Entries in an attribute list that only consist of a value are referred to as positional attributes. The position is the 1-based index of the entry once all named attributes have been removed (so they may be interspersed).

https://docs.asciidoctor.org/asciidoc/latest/attributes/positional-and-named-attributes/#positional

For instance:
```
[first,second=value,Sherlock Holmes]
```

The above example should produce:
```rb
expected = { 1 => 'first', 'second' => 'value', 2 => 'Sherlock Holmes' }
```

Currently, it produces:

```rb
expected = { 1 => 'first', 'second' => 'value', 3 => 'Sherlock Holmes' }
```

ggrossetie:
resolves #4345

### DIFF
--- a/lib/asciidoctor/attribute_list.rb
+++ b/lib/asciidoctor/attribute_list.rb
@@ -66,12 +66,10 @@ class AttributeList
     return @attributes if @attributes
 
     @attributes = {}
-    index = 0
 
-    while parse_attribute index, positional_attrs
+    while parse_attribute positional_attrs
       break if @scanner.eos?
       skip_delimiter
-      index += 1
     end
 
     @attributes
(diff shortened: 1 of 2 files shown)


lib/asciidoctor/attribute_list.rb
```diff
@@ -170,11 +169,11 @@ def parse_attribute index, positional_attrs
         end
       end
     else
+      index = @attributes.select {|key| key.is_a? Numeric }.size
```
Mogztter, line 172: Not the most efficient implementation but `@attributes` is unlikely to exceed 10 items and it removes the `index` parameter from the `parse_attribute` method.

Alternatively, we could return an extra boolean to indicate that we've added a positional attribute and in that case increment the index.

mojavelinux:
See comment on #4345.

PR #4348, opened 2022-08-30
resolves #3961 add loading attribute to image macro

### LINKED ISSUE #3961: Add lazy load attribute for images.
The `loading="lazy"` attribute for lazy loading images should be supported by Asciidoctor. ([MDN](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes))

Eg:

```
image::file_name.ext[loading="lazy", ...]
```

Lazy loading images increases the performance of the site.

The browser will handle the lazy loading of images. [The attribute is supported in the major browsers as well.](https://caniuse.com/?search=loading%3D%22lazy%22) The support will only increase, so I think it is a good feature that can be added.

KartikSoneji:
Adds `image::[loading=eager | loading=lazy]` attributes to the image macro.
Also adds global `image-loading` attribute that can used to set the loading policy document-wide.

resolves: #3961

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -627,17 +627,25 @@ Your browser does not support the audio tag.
     target = node.attr 'target'
     width_attr = (node.attr? 'width') ? %( width="#{node.attr 'width'}") : ''
     height_attr = (node.attr? 'height') ? %( height="#{node.attr 'height'}") : ''
+
+    # https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading
+    loading_attr_value = (node.attr 'loading') || (node.document.attr 'image-loading')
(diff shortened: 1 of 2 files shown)


KartikSoneji:
Hi @mojavelinux is there anything pending for this pull request?
I'm not sure where the docs for this would go, the `docs` folder doesn't seem to have a section on the image macro.
Please let me know what the next steps should be.

mojavelinux:
I'm extremely busy right now and won't be able to look at least until next week. But I'll do my best to prioritize it so we can get it wrapped up.

KartikSoneji:
Appreciate you looking into it.

slonopotamus:
@mojavelinux Maybe busy week has ended?)

mojavelinux:
My busy week never ends. I'm currently focused on the specification. Once that's well in motion, only then will I get back to moving this project forward again towards 2.1.x and beyond.

Issue #4362, opened 2022-10-06
Indicate that the Atom editor is being sunset


stoobie:
It would be helpful to update [IDEs and text editors](https://docs.asciidoctor.org/asciidoctor/latest/tooling/#ides-and-text-editors) so that new users don't waste their time installing and learning to use Atom. :(

See the official statement on the atom site: https://github.blog/2022-06-08-sunsetting-atom/

Mogztter:
Fixed in https://github.com/asciidoctor/asciidoctor/pull/4363
I guess we should cherry-pick this commit onto the main branch.

mojavelinux:
I've already ported it. See ebd63ad96000108fdaa2e97010bf0e66c888fe9a. Thanks for pointing out that this didn't get closed.

PR #4363, opened 2022-10-06
Added IMPORTANT: Atom is being sunset

stoobie:
This PR addresses issue #4362.

Added to the section on text editors:

```
[IMPORTANT]
====
The Atom editor's official site has announced that Atom is being sunset at the end of 2022.

See link:https://github.blog/2022-06-08-sunsetting-atom/[Sunsetting Atom] 
===
```

### DIFF
--- a/docs/modules/tooling/pages/index.adoc
+++ b/docs/modules/tooling/pages/index.adoc
@@ -58,6 +58,13 @@ See its documentation for {url-asciidocfx-docs}[download and installation instru
 
 === Atom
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Great suggestion! I think it's important that users are notified of this change.

FYI, you don't need the `link:` prefix when the target begins with `https:`. See https://docs.asciidoctor.org/asciidoc/latest/macros/link-macro/#when-to-use-the-link-macro

PR #4374, opened 2022-10-25
Fix sentence structure

kensanata:
The alternative would be to simple delete "since" and leave two separate sentences.


mojavelinux:
Right you are! Thanks!

PR #4394, opened 2022-12-21
resolves #4393 avoid matching numeric character references when searching for # in xref target

### LINKED ISSUE #4393: Substitutors matches numeric character references while searching for hash index
Hi, 

in 2.0.18,

```adoc
see <<Cub => Tiger>>

== Cub => Tiger
```

yields:

```html
<div class="paragraph">
<p>see <a href="Cub &.html#8658; Tiger">Cub &.html</a></p>
</div>
<div class="sect1">
<h2 id="_cub_tiger">Cub &#8658; Tiger</h2>
<div class="sectionbody">

</div>
</div>
```

The result should be:

```html
<div class="paragraph">
<p>see <a href="#_cub_tiger">Cub &#8658; Tiger</a></p>
</div>
<div class="sect1">
<h2 id="_cub_tiger">Cub &#8658; Tiger</h2>
<div class="sectionbody">

</div>
</div>
```

I have prepared a patch to fix it.

simonmysun:
Test provided. 

As mentioned in #4393, the substitutor handles link with numeric character references incorrectly. This fixes it by avoiding matches of numeric character references with a regex.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -62,6 +62,7 @@ Improvements::
   * Set linenums option on source block when line numbering is enabled (#3313)
   * Warn if include target is remote and `allow-uri-read` attribute is not set (#2284)
   * Return empty string instead of nil if raw or verbatim block has no lines
+  * Avoid numeric character reference when looking for fragment in target of xref (#4393)
   * Don't split value of `-r` CLI option; treat as single path (#4425)
 
 Bug Fixes::
--- a/lib/asciidoctor/substitutors.rb
+++ b/lib/asciidoctor/substitutors.rb
@@ -756,7 +756,7 @@ module Substitutors
 
         if doc.compat_mode
           fragment = refid
-        elsif (hash_idx = refid.index '#')
+        elsif (hash_idx = refid.index '#') && refid[hash_idx - 1] != '&'
           if hash_idx > 0
             if (fragment_len = refid.length - 1 - hash_idx) > 0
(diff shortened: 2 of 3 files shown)


simonmysun:
I wonder is there anything blocking the process? How may I help?

mojavelinux:
My attention is elsewhere at the moment. When I have a spare moment, I will review this change and make a determination about how to proceed. However, I can't make any promises about when that will be.

simonmysun:
> My attention is elsewhere at the moment. When I have a spare moment, I will review this change and make a determination about how to proceed. However, I can't make any promises about when that will be.

I see. Thx for your time then.

mojavelinux:
Thanks for this PR.

There's actually a simpler way to solve this. Since the text is already XML-escaped at this point, we only have to check if the `#` is preceded by `&` (e.g., `&#169;`). Under normal circumstances, the only time the `&#` sequence of characters would exist at this point is if it's a numeric character reference. The `&` would already be escaped as `&amp;` otherwise. I've updated the PR accordingly.

mojavelinux:
If a strong case arises where we have to look more thoroughly, we will address it at that time. I think this covers most, if not all, cases.

simonmysun:
Thanks for the update and the reviewing!

> There's actually a simpler way to solve this. Since the text is already XML-escaped at this point, we only have to check if the `#` is preceded by `&` (e.g., `&#169;`). Under normal circumstances, the only time the `&#` sequence of characters would exist at this point is if it's a numeric character reference. The `&` would already be escaped as `&amp;` otherwise.

Yes, that makes sense. I haven't come up with any uncovered edge cases yet.

Issue #4382, opened 2022-10-31
Tests not working properly 


Segaja:
```
/usr/bin/ruby -w -I"lib:test" /usr/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader.rb "test/api_test.rb" "test/attribute_list_test.rb" "test/attributes_test.rb" "test/blocks_test.rb" "test/converter_test.rb" "test/document_test.rb" "test/extensions_test.rb" "test/helpers_test.rb" "test/invoker_test.rb" "test/links_test.rb" "test/lists_test.rb" "test/logger_test.rb" "test/manpage_test.rb" "test/options_test.rb" "test/paragraphs_test.rb" "test/parser_test.rb" "test/paths_test.rb" "test/preamble_test.rb" "test/reader_test.rb" "test/sections_test.rb" "test/substitutions_test.rb" "test/syntax_highlighter_test.rb" "test/tables_test.rb" "test/text_test.rb"
RUBY_GC_HEAP_FREE_SLOTS=750000 (default value: 4096)
RUBY_GC_HEAP_INIT_SLOTS=750000 (default value: 10000)
RUBY_GC_HEAP_GROWTH_FACTOR=2.000000 (default value: 1.800000)
RUBY_GC_HEAP_GROWTH_MAX_SLOTS=50000 (default value: 0)
RUBY_GC_MALLOC_LIMIT=128000000 (default value: 16777216)
RUBY_GC_OLDMALLOC_LIMIT=128000000 (default value: 16777216)
Run options: --seed 64652

# Running:

............................................................................................................................................................................................................................................................................................Haml::Engine: Option :attr_wrapper is invalid
Haml::Engine: Option :attr_wrapper is invalid
Haml::Engine: Option :attr_wrapper is invalid
Haml::Engine: Option :attr_wrapper is invalid
F................F.....................................................................................................................................................................F..F.FFF....F...............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................E............F................E..........................................................................................................................................................................................................................................................F..F.................................................................................................................................................................................................................................................................................................

Finished in 2.011873s, 1326.6245 runs/s, 4717.9916 assertions/s.

  1) Failure:
[...]

PR #4395, opened 2022-12-26
Port tests to haml 6

terceiro:
Fixes: #4382

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ group :development do
   gem 'asciimath', (ENV.fetch 'ASCIIMATH_VERSION', '~> 2.0')
   # coderay is needed for testing source highlighting
   gem 'coderay', '~> 1.1.0'
-  gem 'haml', '~> 4.0' if RUBY_ENGINE == 'truffleruby'
   gem 'open-uri-cached', '~> 1.0.0'
   # pygments.rb is needed for testing source highlighting; Asciidoctor supports pygments.rb >= 1.2.0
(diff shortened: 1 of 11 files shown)


mojavelinux:
Thanks for taking this on!

We will still need to test again Haml 5 for compatibility reasons. Would you be able to add conditionals in such a way that the tests work with either version? I'll then make sure the CI matrix is set up to run on both major versions.

terceiro:
> Thanks for taking this on!
> 
> We will still need to test again Haml 5 for compatibility reasons. Would you be able to add conditionals in such a way that the tests work with either version? I'll then make sure the CI matrix is set up to run on both major versions.

There.

mojavelinux:
Thanks! I'll review as soon as you think it's ready to go. Just give me the signal.

terceiro:
this should be ready. I tested against haml 5 and 6 by hacking the Gemfile, and it works

thesamesam:
Thanks for this, I was just trying to figure out why the tests were failing for us :)

@mojavelinux able to take a look?

mojavelinux:
Superseded by #4431.

I don't want to introduce an intermediary class to add support for Haml 6, and I don't think it's necessary. Instead, I want to use the existing infrastructure in the template converter to deal with option variance. However, this PR did point me to what needed to be changed, so I very much appreciate the contribution.

>  I tested against haml 5 and 6 by hacking the Gemfile, and it works

We can't just rely on local testing. I added a profile to the CI workflow to run the test suite using Haml 5 in additional to Haml 6.

PR #4407, opened 2023-02-21
resolves #4406 fix test assertion for fallback Rouge stylesheet to be compatible with Rouge 4.1

### LINKED ISSUE #4406: test suite failures with rouge 4.1.0
A change in rouge 4.1.0 causes the `should gracefully fallback to default style if specified style not recognized` test to fail.  The CSS produced by rouge no longer matches the regex.  This looks to be due to [6f5aecd2 (Update GitHub theme, add dark mode (#1918), 2023-01-24)](https://github.com/rouge-ruby/rouge/commit/6f5aecd2) in rouge.

This was noticed in Fedora ([bz#2171711](https://bugzilla.redhat.com/2171711)). Here is the [build.log](https://github.com/asciidoctor/asciidoctor/files/10786328/build.log).

tmzullinger:
Adjust the pattern used by the `should gracefully fallback to default style if specified style not recognized` test to support rouge-4.1.0.

Upstream rouge commit 6f5aecd2 (Update GitHub theme, add dark mode (#1918), 2023-01-24) changed the CSS output.  Work with both old and new CSS.

resolves #4406

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -42,6 +42,7 @@ Compliance::
   * Disallow the use of dot (`.`) in the name of a named element attribute (#4147)
   * Disallow the use of the left square bracket in an attribute list on formatted text (#4306)
   * Fix call order so use of an include file with invalid encoding continues to raise error when using Ruby >= 3.2.0
+  * Fix test assertion for fallback Rouge stylesheet to be compatible with Rouge 4.1 (#4406) (*@tmzullinger*)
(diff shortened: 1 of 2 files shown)


tmzullinger:
I don't know if making the regex more complicated is better than comparing the old and new rouge CSS and finding a pattern which remains consistent.

I'll understand if this is too ugly to merge. :)

mojavelinux:
Thanks! This should get me started.

PR #4411, opened 2023-03-09
Add closing parenthesis to capture group to properly catch enclosed links

r0ckarong:
I am working on an extension that extracts more detailed info about documents. I was trying to get the links and their source code location. I noticed that some of the links would have broken match data and wouldn't work. After some digging I noticed that this only happens when you have links enclosed in parentheses.

The regex for `InlineLinkRx` incorrectly matches the closing parenthesis into the third capture group. This usually is not a problem because `sub_macros` will perform some additional sanitizing and correctly strip the parenthesis for the internal processing. However, if you try to just filter the unprocessed lines with the regex you'll end up with false matches.

```ruby
require 'asciidoctor'

include Asciidoctor

testlinks = [
  'ABCD https://www.google.com',
  'ABCD (https://www.google.com)',
  'link:https://www.google.com',
  '(link:https://www.google.com)',
  'link:https://www.google.com[]',
  '(link:https://www.google.com[])',
]

puts
puts 'InlineLinkRx'
puts InlineLinkRx

testlinks.each do |link|
  bar = link.match(InlineLinkRx)
  pp bar
end

puts
patched = %r((^|link:|#{CG_BLANK}|&lt;|[>\(\)\[\];"'])(\\?(?:https?|file|ftp|irc)://[^\s\[\]<]*([^\s.,\[\]<\)]))(?:\[(|#{CC_ALL}*?[^\\])\])?)m
puts 'Patched InlineLinkRx'
puts patched

testlinks.each do |link|
  bar = link.match(patched)
  pp bar
end
```

Note how MatchData result 2 the has these parentheses included in the current version. When you use an empty attriblist `[]` it matches correctly. This only occurs if you have a link without macro or macro without attriblist enclosed in parenthesis.

```
InlineLinkRx
(?m-ix:(^|link:|\p{Blank}|&lt;|[>\(\)\[\];"'])(\\?(?:https?|file|ftp|irc):\/\/[^\s\[\]<]*([^\s.,\[\]<]))(?:\[(|.*?[^\\])\])?)
#<MatchData " https://www.google.com" 1:" " 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "(https://www.google.com)" 1:"(" 2:"https://www.google.com)" 3:")" 4:nil>
#<MatchData "link:https://www.google.com" 1:"link:" 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "link:https://www.google.com)" 1:"link:" 2:"https://www.google.com)" 3:")" 4:nil>
#<MatchData "link:https://www.google.com[]" 1:"link:" 2:"https://www.google.com" 3:"m" 4:"">
#<MatchData "link:https://www.google.com[]" 1:"link:" 2:"https://www.google.com" 3:"m" 4:"">

Patched InlineLinkRx
(?m-ix:(^|link:|\p{Blank}|&lt;|[>\(\)\[\];"'])(\\?(?:https?|file|ftp|irc):\/\/[^\s\[\]<]*([^\s.,\[\]<\)]))(?:\[(|.*?[^\\])\])?)
#<MatchData " https://www.google.com" 1:" " 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "(https://www.google.com" 1:"(" 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "link:https://www.google.com" 1:"link:" 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "link:https://www.google.com" 1:"link:" 2:"https://www.google.com" 3:"m" 4:nil>
#<MatchData "link:https://www.google.com[]" 1:"link:" 2:"https://www.google.com" 3:"m" 4:"">
#<MatchData "link:https://www.google.com[]" 1:"link:" 2:"https://www.google.com" 3:"m" 4:"">
```

### DIFF
--- a/lib/asciidoctor/rx.rb
+++ b/lib/asciidoctor/rx.rb
@@ -517,7 +517,7 @@ module Asciidoctor
   #   "https://github.com[]"
   #
   # FIXME revisit! the main issue is we need different rules for implicit vs explicit
-  InlineLinkRx = %r((^|link:|#{CG_BLANK}|&lt;|[>\(\)\[\];"'])(\\?(?:https?|file|ftp|irc)://[^\s\[\]<]*([^\s.,\[\]<]))(?:\[(|#{CC_ALL}*?[^\\])\])?)m
+  InlineLinkRx = %r((^|link:|#{CG_BLANK}|&lt;|[>\(\)\[\];"'])(\\?(?:https?|file|ftp|irc)://[^\s\[\]<]*([^\s.,\[\]<\)]))(?:\[(|#{CC_ALL}*?[^\\])\])?)m
 
   # Match a link or e-mail inline macro.
   #


r0ckarong:
Probably relates to this: #3376

mojavelinux:
Since the current parser has no aware of context, this is an extremely tricky situation. The trailing characters are often consumed then sanitized as a way to provide some context sensitivity. So it was done the way it is done for good reason. But that doesn't imply that it is without problems.

If we add a character to this regular expression, then we need an additional test case to verify that it works properly and doesn't break any other cases. The fact that the tests are passing is already a good sign. But clearly there must be a boundary condition to exercise here.

mojavelinux:
> The fact that the tests are passing is already a good sign.

Aha. The CI workflow never ran. If it did, tests would have failed.

r0ckarong:
> To reiterate, these regular expressions are not something you should be relying on in external code. They are purely internal.

My intention was to use as much "internal" functionality to identify document aspects so I wouldn't implement anything that would deviate from how asciidoctor does it. Thanks for looking into this and all the hard work.

PR #4412, opened 2023-03-13
Add support for table striping in docbook converter

snaggen:
This patch adds support to have tables striped in the docbook converter, on par with the html converter.


mojavelinux:
I do not want to hard code colors into the DocBook output. If DocBook provides an attribute that instructs the backend to add striping, then I would consider it. Otherwise, this is not an feature I am willing to consider.

If it is something you need, you can extend the converter to add it (see https://docs.asciidoctor.org/asciidoctor/latest/convert/custom/#extend-and-replace-a-registered-converter).

snaggen:
Closing this. Will submit a combined patch for both header style and table

PR #4413, opened 2023-03-13
Add option to shade the header background

snaggen:
This patch adds an option to shade the table header background in the docbook converter.


mojavelinux:
For the same reason as #4412, I'm not willing to consider hard coding colors into the DocBook output.

PR #4416, opened 2023-03-16
Clarify that source_location is not tracked by default

ShabbyX:
... in the API documentation. The docs/ explain this, but a hint at the sourcemap option in the API docs would allow the reader to find that information.


ShabbyX:
Added a bit to the doc about how sourcemap can be enabled by an extension.

mojavelinux:
Thanks!

PR #4417, opened 2023-03-16
Add --sourcemap to v2.0

ShabbyX:


mojavelinux:
On [this page](https://docs.asciidoctor.org/asciidoctor/latest/api/sourcemap/#set-sourcemap-option), it might be good to show how an extension can ensure the sourcemap is enabled (to build on your open PR about the sourcemap option).

Issue #761, opened 2013-10-31
Introduce attribute for controlling the location of the MathJax script


mojavelinux:
Introduce attributes for controlling the location of the MathJax script. There are two cases:
- CDN location
- relative path location

This issue is dependent on #659.

mojavelinux:
It also allows the document to set the version of MathJax being used (just like we do for Font Awesome).

jxxcarlson:
Good!

akosma:
Just hit this issue using asciidoctor-reveal.js and later found this issue here :)

akosma:
I submitted a solution for this issue in the context of asciidoctor-reveal.js that could be useful here as well: https://github.com/asciidoctor/asciidoctor-reveal.js/pull/350

createyourpersonalaccount:
Here's the fix I use (post-process the AST after parsing):

1. Include a `localmathjax.rb` file with contents:

```ruby
require 'asciidoctor'
Asciidoctor::Extensions.register do
  postprocessor do
    process do |document, output|
      mathjaxjs = document.attr('mathjaxjs', 'mathjax/tex-chtml.js')
      output.gsub(
        %r{<script\b[^>]*\bsrc="[^"]*MathJax\.js[^"]*"[^>]*></script>},
        %(<script defer src="#{mathjaxjs}"></script>)
      )
    end
  end
end
```
2. In your asciidoc files specify `:stem: latexmath` and `:mathjaxjs: mathjax/tex-chtml.js`.
3. In your makefile use `asciidoctor -r localmathjax.rb -o article.html article.adoc`

Issue #659, opened 2013-10-01
Use a consistent attribute naming scheme for resolving assets


mojavelinux:
We need to define a consistent attribute naming scheme for resolving assets such as Font Awesome, prettify and highlight.js.

Here's how Font Awesome is currently resolved:
- `iconfont-remote` - enables or disables use of remote asset
- `iconfont-cdn` - full URL to CDN (remote) asset
- `iconfont-name` - name of local CSS file (without `.css` extension)

Here's how highlight.js is current resolved:
- `highlightjsdir` - base directory (or remote URL) of asset
- `highlightjs-theme` - name of CSS file (without `.min.css` extension)

Here's how prettify is currently resolved:
- `prettifydir` - base directory (or remote URL) of asset
- `prettify-theme` - name of CSS file (without `.min.css` extension)

In the last two cases, I don't think we have fine enough granularity. I'm leaning more towards the way the icon font is handled, which is either a full path or attributes for various path segments.

We need to have sufficient flexibility and consistency so that it's easy for the user to remember the attributes and apply the naming pattern regardless of which asset they are using.

nh2:
+1 for this. I would like to view docs without access to the Internet. This would help a lot.

mojavelinux:
I want to clarify that this is currently possible. The names of the attributes just aren't consistent.

The following line should reveal what you need to set to refer to local files:

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L73

Nearby lines reveal which attributes to set to control the path to other resources.

nh2:
@mojavelinux Maybe I'm missing something, but it seems that I can't change where [MathJax is located](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L130) since `cdn_base` doesn't seem to be overridable by an attribute. Or is it?

mojavelinux:
Ooops, MathJax is one of the few libraries that you can't currently control :/ Sorry for misleading you. That's part of why we need to address this.

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L130

You could hack the template using the template-based converter. Here's the default set of templates to hack on. You only need to override what you need to change (in this case, document.html.slim). See https://github.com/asciidoctor/asciidoctor-backends/tree/master/slim/html5

jgibson:
[...]

PR #4420, opened 2023-03-17
Factor out cdn_base_url.

expeehaa:
These changes are intended as a kind of intermediate solution to #761 and, to a small degree, also #659.

I use AsciiMath with the html5 backend and want the math part to be rendered even if I have no network connection.
For that I can currently either patch the gem sources to use a different `cdn_base_url` in `Asciidoctor::Converter::Html5Converter#convert_document`, monkey-patch the entire 200-line method or replace the URL after the conversion. The first 2 are too much maintenance in case I want to update the gem, while the last one always requires maintenance should `cdn_base_url` change.

This pull request does not change any behavior of the converter methods and instead factors out the quasi-constant variable `cdn_base_url` (and additionally its dependency `asset_uri_scheme`) into a method that can be monkey-patched with much less maintenance. (I realize that this might not have less maintenance than simply replacing the URL afterwards since the method could be removed, but at least it would be possible.)
Obviously this does not solve #761 and neither should it be a recommended solution until #761 is solved, but it at least reduces the burden of making a solution for oneself.

Since I’m not familiar with the development style here, this request is a bit of a shot in the dark. If I miss, I would also be happy to help with fully solving #659 and then #761 properly, although it seems like there are still some design decisions required.

### DIFF
--- a/lib/asciidoctor/converter/html5.rb
+++ b/lib/asciidoctor/converter/html5.rb
@@ -93,7 +93,6 @@ class Converter::Html5Converter < Converter::Base
 
   def convert_document node
     br = %(<br#{slash = @void_element_slash}>)
-    cdn_base_url = %(#{asset_uri_scheme_from_node node}//cdnjs.cloudflare.com/ajax/libs)
     linkcss = node.attr? 'linkcss'
     max_width_attr = (node.attr? 'max-width') ? %( style="max-width: #{node.attr 'max-width'};") : ''
     result = ['<!DOCTYPE html>']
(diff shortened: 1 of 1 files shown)


mojavelinux:
I'd rather keep asset_uri_scheme constant and just have a helper for cdn_base_url. I'd even consider allowing cdn_base_url to be overridden by a document attribute in the default implementation. I don't see any reason why asset_uri_scheme should not remain constant as mixing schemes in a web page is considered an unsecure practice.

mojavelinux:
> This pull request does not change any behavior of the converter methods

Actually it does. It allows the `asset-uri-scheme` to be set per block element (that uses it). That is a fundamental change in behavior and not one that we want to permit.

mojavelinux:
As I said, I'm find with extracting the cdn_base_url to a method. That method should accept the URL scheme as a parameter (since that should be fixed throughout the document). Alternately, the method could return the URL without the scheme (starting from the host), though that seems less clean.

PR #4421, opened 2023-03-20
Add styling for table header and striping in docbook converter

snaggen:
Adds a tab style attribute with striping and header background value if they are present. The actual table striping is now handled downstream in docbook xsl. 

An example of the table.row.properties needed to use these features
```xml
<xsl:template name="table.row.properties">
    <xsl:choose>
        <xsl:when test="$bgcolor != ''">
            <xsl:attribute name="background-color">
                <xsl:value-of select="$bgcolor" />
            </xsl:attribute>
        </xsl:when>
        <xsl:when test="contains($tabstyle, 'striped-odd')">
            <xsl:if test="$rownum mod 2 != 0 and not(ancestor::thead)">
                <xsl:attribute name="background-color">
                    <xsl:value-of select="$table.row.background-color" />
                </xsl:attribute>
            </xsl:if>
        </xsl:when>
        <xsl:when test="contains($tabstyle, 'striped-even')">
            <xsl:if test="$rownum mod 2 = 0">
                <xsl:attribute name="background-color">
                    <xsl:value-of select="$table.row.background-color" />
                </xsl:attribute>
            </xsl:if>
        </xsl:when>
    </xsl:choose>
    <xsl:if test="ancestor::thead">
        <xsl:if test="contains($tabstyle, 'headerbg')">
            <xsl:attribute name="background-color">
                <xsl:value-of select="$table.heading.background-color" />
            </xsl:attribute>
        </xsl:if>
    </xsl:if>
</xsl:template>
```

### DIFF
--- a/lib/asciidoctor/converter/docbook5.rb
+++ b/lib/asciidoctor/converter/docbook5.rb
@@ -379,7 +379,11 @@ class Converter::DocBook5Converter < Converter::Base
     pgwide_attribute = (node.option? 'pgwide') ? ' pgwide="1"' : ''
     frame = 'topbot' if (frame = node.attr 'frame', 'all', 'table-frame') == 'ends'
     grid = node.attr 'grid', nil, 'table-grid'
-    result << %(<#{tag_name = node.title? ? 'table' : 'informaltable'}#{common_attributes node.id, node.role, node.reftext}#{pgwide_attribute} frame="#{frame}" rowsep="#{(%w(none cols).include? grid) ? 0 : 1}" colsep="#{(%w(none rows).include? grid) ? 0 : 1}"#{(node.attr? 'orientation', 'landscape', 'table-orientation') ? ' orient="land"' : ''}>)
+    stripes = node.attr 'stripes', nil, 'table-stripes'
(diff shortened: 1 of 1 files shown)


mojavelinux:
[tabstyle](https://tdg.docbook.org/tdg/5.1/html.table.html) does seem to be a standard attribute in DocBook, so we're now heading in the right direction. Are these values standardized anywhere?

snaggen:
There doesn't seem to be any standardization of these values according to our research.

mojavelinux (changes requested):
I'm fine with adding `tabstyle` for table stripes. However, I'd like to stick with the same names that are used by the HTML converter. Those are as follows:

* stripes-all
* stripes-even
* stripes-odd

I don't like the idea of a headerbg option. It's not part of AsciiDoc (i.e., not compliant) and I don't see why this can't be controlled using a role (e.g., `.header-bg`).

I can see the benefit of mapping the stripes attribute to tabstyle (or a role) since that way you don't need to define it twice.

Also, please add a test.

mojavelinux:
There needs to be an associated issue in order for this to be merged. Unless it's a documentation changes, pull requests are not accepted without an issue since there is then no reference for why the change was made.

snaggen:
Added bug #4593

PR #4446, opened 2023-05-03
Add Ruby 3.1 to CI

stanhu:


mojavelinux:
This is unnecessary. We only test using the oldest and newest minor for any major we support, limited of course to the oldest minor version we support (in this case 2.5). There's no need to test all the in between minor versions as there should be nothing that creates an incompatibility assuming Ruby follows SemVer correctly. Adding more versions is just wasteful and slows down the CI workflow.

PR #4447, opened 2023-05-03
Fix mantitle parsing (fixes #4445)

### LINKED ISSUE #4445: Escaped replacement in doctitle (mantitle) creates invalid output in DocBook output
Some manpages require double dashes (`--`), but there's no consistent way to write them:

    = foo--bar(1)

This generates a proper output for the `manpage` backend, but not `html` or `docbook5`, they get replaced with `&#8212;&#8203;`.

If they are escaped:

    = foo\--bar(1)

Then the `manpage` and `html` backends work fine, but `asciidoctor -b docbook5 -d manpage` doesn't, as it generates this:

    <refentrytitle>foo\--bar</refentrytitle>

And docbook stylesheets doesn't generate a proper manpage.

It seems no combination works, except using an attribute, and then do something like `foo{dd}bar`, but that seems far from ideal.

If there was a way to globally disable this replacement (as suggested in #1061) then I suppose things would just work.

It seems `mantitle` is not being parsed correctly, but it can't be specified either.

felipec:
If the document title is unescaped for `<title>`, it should be unescaped for `<refentrytitle>` as well.

    = foo\--bar\--roo(1)
    <title>foo--bar--roo(1)</title>
    <refentrytitle>foo--bar--roo</refentrytitle>

### DIFF
--- a/lib/asciidoctor/parser.rb
+++ b/lib/asciidoctor/parser.rb
@@ -206,7 +206,7 @@ class Parser
   def self.parse_manpage_header reader, document, block_attributes, header_only = false
     if ManpageTitleVolnumRx =~ (doc_attrs = document.attributes)['doctitle']
       doc_attrs['manvolnum'] = manvolnum = $2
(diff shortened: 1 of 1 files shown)


mojavelinux:
Superseded by #4449.

PR #4466, opened 2023-06-06
Patch 1: Added Infos about Android/Markor

k3b:


mojavelinux:
Thanks for bringing this project to our attention. However, the README for Asciidoctor is not the right place to be promoting integrations/editors. You are welcome to announce it in the project chat at https://chat.antora.org. If we get good feedback about it, then we might decide to add it to https://docs.asciidoctor.org/asciidoctor/latest/tooling/. But I want to see testimonials about it before we do that.

PR #4469, opened 2023-07-11
Update templates.adoc with new URL for Slim

kadang:
The original value for  `:url-slim:` no longer points to a page about the Slim template language.


mojavelinux:
Thanks!

PR #4485, opened 2023-09-07
[Docs] Add command for generating default Gemfile

djl101:
To help new users, I thought the Bundler instructions could include the `bundle init` command. This generates a Gemfile with the default rubygems.org source [1].

[1] https://bundler.io/guides/gemfile.html#gemfiles


mojavelinux:
This seems reasonable to me. Thanks for suggesting it.

PR #4488, opened 2023-09-15
resolves #4492 add Wistia support to the video macro

### LINKED ISSUE #4492: Add support for the Wistia video service
As mentioned in [this comment](https://github.com/asciidoctor/asciidoctor/issues/1690#issuecomment-192434666), it would be relatively straightforward to add Wistia support to Asciidoctor. The goal would be to include basic support - we don't need all the bells and whistles, but the core functions that you see with the Youtube and Vimeo implementations would be a solid start.

Nadreck:
This PR adds basic support for Wistia to the `video::` macro, following a similar approach to how Youtube and Vimeo are currently supported. It supports some basic options (controlling looping behavior, autoplay, start time, dimensions, and muting), and uses an `iframe` embedding method.

It does include referencing an external script they recommend including with their embeds - while recommended, we don't strictly have to for things to work, so your call on whether you'd prefer we remove it.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -35,6 +35,7 @@ Enhancements::
   * Honor `level` and `formatter` keyword arguments passed to Logger constructor (#4250)
   * Set logdev to $stderr if no arguments are passed to Logger constructor (#4250)
   * Add support for skipping TOML front matter (Hugo) when `skip-front-matter` attribute is set (#4300) (*@abhinav*)
+  * Add support for Wistia using the video macro
 
 Compliance::
 
--- a/data/reference/syntax.adoc
+++ b/data/reference/syntax.adoc
@@ -287,6 +287,8 @@ video::aHjpOzsQ9YI[youtube]
 
 video::300817511[vimeo]
 
+video::e4a27b971d[wistia]
+
 == Breaks
 
 // thematic break (aka horizontal rule)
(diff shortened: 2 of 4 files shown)


mojavelinux:
This seems like a reasonable addition.

Could you file an issue so that this change can be tracked?

> It does include referencing an external script they recommend including with their embeds

I would prefer not to include a link to a script in the HTML output, especially if it can be avoided. First, it's going to be repeated for each video on the page...and thus more work would need to be done to extract it to the top level. I think it's better if we just go without it. If it becomes a problem, we can revisit it.

Nadreck:
More info on the external script is [here](https://wistia.com/support/developers/e-v1) - including it with every embed is what they actually recommend, unless there's a particular reason to cut down on bloat.

According to their documentation, it sounds like we lose a few things by axing the script, but none sound like blockers for this implementation:

> However, there are still some key limitations that apply when E-v1.js is missing from your webpage, even with a Fallback/iframe embed. They pertain to features that rely on E-v1 being there for various on-page interactions or page-specific functionality.
>
>Without E-v1 on your webpage, these features will not work as expected:
>
>Google Analytics
>Marketing Automation Integrations (Hubspot, Marketo, Pardot)
>JavaScript Player API
>Domain Restrictions

Considering implementers planning to use the feature can always insert the script in their UI template to regain those additional features, I don't see removing the script as a bad tradeoff, but I did want to note what those tradeoffs are. :thumbsup:

Updated the PR with the script removed.

mojavelinux:
Thanks for filing the issue and updating the PR.

It seems like those are all reasonable tradeoffs. We don't really want Wistia to be putting all that marketing cruft into the page anyway. Just the video, please ;)

mojavelinux:
Looks good to me! Thanks!

PR #4495, opened 2023-10-07
resolves #3466 update stylesheet to use break-after in place of page-break-before/after/inside

### LINKED ISSUE #3466: Use break-after property in the HTML 5 converter
> `page-break-after` has been replaced by the `break-after` property

https://developer.mozilla.org/en-US/docs/Web/CSS/page-break-after

As a result, we should use `break-after` in the `convert_page_break` method:

https://github.com/asciidoctor/asciidoctor/blob/d1aee25c12556cd11d813b8964944c8679c61c41/lib/asciidoctor/converter/html5.rb#L776

mojavelinux:


data/stylesheets/asciidoctor-default.css
```diff
@@ -295,7 +295,7 @@ sup.footnote a:active,sup.footnoteref a:active{text-decoration:underline}
 #footnotes .footnote:last-of-type{margin-bottom:0}
 #content #footnotes{margin-top:-.625em;margin-bottom:0;padding:.75em 0}
 div.page-break{display:none}
-div.unbreakable{page-break-inside:avoid}
```
ggrossetie, line 298: According to MDN, Firefox supports `break-inside:avoid`. Do we really need `-moz-column-break-inside`?

https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside

data/stylesheets/asciidoctor-default.css
```diff
@@ -295,7 +295,7 @@ sup.footnote a:active,sup.footnoteref a:active{text-decoration:underline}
...
+div.unbreakable{-moz-column-break-inside:avoid;break-inside:avoid}
```
ggrossetie, line 298: According to MDN, Firefox supports `break-inside:avoid`. Do we really need `-moz-column-break-inside`?

https://developer.mozilla.org/en-US/docs/Web/CSS/break-inside

mojavelinux: That's being added by autoprefixer. It seemed odd to me too, but I couldn't see how to make it stop without changing our setting from "default". See https://github.com/asciidoctor/asciidoctor/blob/main/src/stylesheets/package.json#L16-L18

PR #4496, opened 2023-10-09
resolves #3526 return nil if name passed to Asciidoctor::SafeMode.value_for_name is not recognized

### LINKED ISSUE #3526: SafeMode#value_for_name throws an exception if name is not found
But `SafeMode#name_for_value` returns `nil` if the value is not found. 
I think we should use the same behavior between the two methods (ie. either throws an exception or returns nil if the SafeMode is not found).

Since the Asciidoctor API does not extensively use exceptions, I think we should return `nil` if the name is not found:

```ruby
SafeMode.value_for_name 'server'   # returns 10 (SafeMode::SERVER)
SafeMode.value_for_name 'paranoid' # returns nil
```

mojavelinux:


ggrossetie (approved):
Looks good 👍🏻

PR #4501, opened 2023-10-11
Changed file name attributes-sv.adoc to attributes-se.adoc

docudoc:
Changed file name from attributes-sv.adoc to attributes-se.adoc, since SV is  El Salvador, and SE is Sweden.


docudoc:
It seems I misunderstood which tag goes into the file name.

mojavelinux:
I believe we are following 639-1 lang codes. See https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

PR #4538, opened 2024-01-17
RFC: Use YARD to document API

slonopotamus:
Motivation:

1. While rubydoc.org understands some of the current documentation format, there are other tools that do not. Most notable example is RubyMine.
2. Tools (including rubydoc.org) cannot infer nullability from current documentation (should I say "nillability" in Ruby?).
3. Tools (including rubydoc.org) do not properly understand variant types.
4. Things like visibility are already described by code itself and feel redundant.

### DIFF
--- a/lib/asciidoctor/abstract_node.rb
+++ b/lib/asciidoctor/abstract_node.rb
@@ -1,31 +1,46 @@
 # frozen_string_literal: true
 
 module Asciidoctor
-# Public: An abstract base class that provides state and methods for managing a
+# An abstract base class that provides state and methods for managing a
 # node of AsciiDoc content. The state and methods on this class are common to
 # all content segments in an AsciiDoc document.
 class AbstractNode
   include Logging
   include Substitutors
 
-  # Public: Get the Hash of attributes for this node
+  # Get the Hash of attributes for this node
+  #
+  # @return [Hash<String, String>]
   attr_reader :attributes
 
-  # Public: Get the Symbol context for this node
+  # Get the Symbol context for this node
+  #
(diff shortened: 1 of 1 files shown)


mojavelinux:
I came to similar conclusions when working on Asciidoctor core...though I've yet to fully apply the change there.

slonopotamus:
I don't think converting the whole codebase in one go is manageable, because at least nullability requires conscious thoughts. In total, I spent ~20mins on `AbstractBlock` in current PR and not fully sure I got everything right. So, my suggestion is to do it gradually.

ggrossetie:
Speaking of YARD it would be great if we could publish the API documentation on https://docs.asciidoctor.org/ (HTML pages) instead of relying on https://www.rubydoc.info/

@slonopotamus If you are interested it might be a good idea to set it up on the Asciidoctor EPUB3 project first?

slonopotamus:
Sure, we can use other project for testing.

Though, I want to add a couple of thoughts:

1. I'm not sure publishing API on a website has much value. We use API when writing code, so we're in IDE, which we use to navigare through code. And why argument/return types are important is because they allow IDE to offer autocomplete and provide warnings when wrong types are used.
2. Asciidoctor Core is much more important API-wise because it is used by extensions. I haven't heard about anyone writing code that would extend Asciidoctor EPUB3. Type info can be useful _within_ a program, but my main point is about externally-visible stuff.

mojavelinux:
Please discuss hosting the API in https://github.com/asciidoctor/asciidoctor.org/issues/1011 (or in the dev stream in the project chat).

PR #4541, opened 2024-01-31
Add warning about Google Fonts to documentation

### LINKED ISSUE #4540: Warn users about the use of Google Fonts
The use of Google Fonts in Asciidoctor-generated HTML (that uses the default style and doesn't go out of its way to add a consent form) violates the GDPR - see [here for example](https://www.cookieyes.com/documentation/google-fonts-and-gdpr/).

As stated [here](https://github.com/asciidoctor/asciidoctor/issues/2515#issuecomment-1915402879):

> A majority of users prefer and rely on this feature, so the default will not be changed.

I'll give you the benefit of the doubt for that, but either way it doesn't help users who don't even *know* about the use of Google Fonts. I was certainly very surprised to see that I was *accidentally* violating the GDPR. It was surprising that Asciidoctor by default uses Google Fonts.

I think you should at least warn users about this. It is mentioned [here](https://docs.asciidoctor.org/asciidoctor/latest/html-backend/default-stylesheet/#web-fonts) but that is pretty deep in the docs and not highlighted. I would suggest a note on the command line like

```
NOTE: Using Google Fonts.
```

Or failing that, a warning box on [this page](https://docs.asciidoctor.org/asciidoctor/latest/html-backend/).

Alternatively, you could inline the fonts. WOFF2 fonts are very small (a few 10s of kB). That may be a better solution actually. Would you accept a PR for that?

Timmmm:
Fixes #4540

### DIFF
--- a/docs/modules/html-backend/pages/index.adoc
+++ b/docs/modules/html-backend/pages/index.adoc
@@ -7,6 +7,8 @@ Asciidoctor's default output format is HTML.
 
 `html`:: The HTML 5 converter (`html` or `html5`) generates HTML 5 styled with CSS3.
 
+WARNING: The <<Default stylesheet>> uses Google Fonts which is https://cookie-script.com/blog/google-fonts-and-gdpr[not GDPR compliant]. To avoid this issue you can <<Apply a Custom Stylesheet>>.
+
 == Generate HTML using the html5 converter
 
 In this section, we'll create a sample document, then process and convert it with Asciidoctor's built-in HTML converter.


mojavelinux:
> To avoid this issue you can <<Apply a Custom Stylesheet>>.

This recommendation is too drastic. You only have to unset the `webfonts` attribute and the Google Fonts will not be loaded. It's a completely reasonable compromise.

Timmmm:
Yeah that makes sense. Only thing I would change is to include the actual code for how to unset the attribute so people don't have to look it up. I'll update the PR.

mojavelinux:
I'd rather link to the section about webfonts (https://docs.asciidoctor.org/asciidoctor/latest/html-backend/default-stylesheet/#disable-or-modify-the-web-fonts). There's no need to duplicate the information. Then it just has to be maintained in multiple places.

mojavelinux:
Were you going to make this update, or would you like me to proceed with the text I proposed?

Timmmm:
I was going to do it but I'd you'd like to just get it done please go ahead! I am very short on free time unfortunately :/

mojavelinux:
Sure thing, I can take care of it. I just didn't want to step on any toes.

mojavelinux:
This information has been added in 5cfd8adfdae60ca1168315a78999c80610c91568. See https://docs.asciidoctor.org/asciidoctor/latest/html-backend/

Timmmm:
Thanks! :tada:

PR #4545, opened 2024-02-05
Sidebars for semantic HTML 5 converter

redbow-kimee:
Code and tests for making [asciidoc sidebars](https://docs.asciidoctor.org/asciidoc/latest/blocks/sidebars/).


redbow-kimee:
A notable question in this feature is the handling of titles. In most other blocks the translation to `strong class="title"` works, In a sidebar, I think the title should be a block element,  so I've done `<p><strong class="title">Title...`. But this may lead to the awkward syntax difference between single-paragraph and multi-paragraph sidebars. Notice in the single-paragraph example the title is in a `<p>` but the content is not in a `<p>`. But for the multi-paragraph, both the title and content have `<p>` containers.

```
<aside>
<p role="heading"><strong class="title">Sidebar Title</strong></p>
This is a sidebar.
</aside>
```

```
<aside>
<p role="heading"><strong class="title">Sidebar Title</strong></p>
<p>
This is a sidebar.
</p>
<p>
This is another paragraph in the sidebar.
</p>
</aside>
```

Another heading is if it is appropriate to give the title's `<p>` the[ ARIA `role="heading"` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role). It seems appropriate to me.

The alternative to both of these would be to use an `<Hn>` tag. Although it could be complicated to determine the necessary number, and because it is an `<aside>` its title probably is irrelevant to the document's structure, and shouldn't be an `<Hn>` tag.

PR #4549, opened 2024-02-11
Block quotes for the semantic HTML 5 Converter

redbow-kimee:
This MR adds functionality to convert block-quotes in AsciiDoctor for the semantic-html5 converter/backend. I've added the examples from [the asciidoctor](https://docs.asciidoctor.org/asciidoc/latest/blocks/blockquotes/) manual as tests. The `blockquotes-6` test requires unordered lists, which is blocked by MR #4321.

This is an example of the old and new output.

OLD:

```
<div class="quoteblock">
<div class="title">After landing the cloaked Klingon bird of prey in Golden Gate park:</div>
<blockquote>
Everybody remember where we parked.
</blockquote>
<div class="attribution">
&#8212; Captain James T. Kirk<br>
<cite>Star Trek IV: The Voyage Home</cite>
</div>
</div>
```

NEW:

```
<blockquote>
<header><strong class="title">After landing the cloaked Klingon bird of prey in Golden Gate park:</strong></header>
Everybody remember where we parked.
<footer>
<span class="blockquote-attribution">Captain James T. Kirk</span>
<cite class="blockquote-citation">Star Trek IV: The Voyage Home</cite>
<footer>
</blockquote>
```

The use of `<header>` for the title matches my latest [suggestions for admonitions]. The use of `<footer>` matches MDNs examples https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote.

### DIFF
--- a/lib/asciidoctor/converter/semantic_html5.rb
+++ b/lib/asciidoctor/converter/semantic_html5.rb
@@ -92,6 +92,31 @@ class Converter::SemanticHtml5Converter < Converter::Base
 </figure>)
   end
 
+  def convert_quote node
+    attributes = common_html_attributes node.id, node.role
+    title = node.title ? %(<header><strong class="title">#{node.title}</strong></header>) : nil
+    attribution = (node.attr? 'attribution') ? (node.attr 'attribution') : nil
+    citation = (node.attr? 'citetitle') ? (node.attr 'citetitle') : nil
+    ret = []
+    ret << "<blockquote>"
+    if title
+      ret << title
+    end
+    ret << node.content
+    if attribution or citation
+      ret << "<footer>"
+      if attribution
+        ret << %(<span class="blockquote-attribution">#{attribution}</span>)
+      end
+      if citation
+        ret << %(<cite class="blockquote-citation">#{citation}</cite>)
+      end
+      ret << "<footer>"
+    end
+    ret << "</blockquote>"
+    ret.join LF
+  end
+
   def convert_thematic_break node
     '<hr>'
   end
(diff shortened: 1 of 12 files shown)


redbow-kimee:
I added `convert_verse` too, because it's basically the same as block quote.

mojavelinux:
Glad to see this taking shape.

The paragraph content needs to be wrapped in a paragraph. We should never end up with text adjacent to a block-oriented element. It's simply too difficult to style it using CSS. For a styled paragraph, it should be implicitly wrapped in `<p>` tag. We see this used in Asciidoctor EPUB3. See https://github.com/asciidoctor/asciidoctor-epub3/blob/main/lib/asciidoctor-epub3/converter.rb#L1326-L1328

redbow-kimee:
I've added a condition so that content that does not start with a `<` will be surrounded by a `<p>`, to avoid situations where multi-paragraph quotes would be doubly surrounded by `<p>`, although, unless non-html `<` symbols are always replaced with `&lt;` then there may be edge-cases where this is performed incorrectly.

redbow-kimee:
Okay. I've tested this a bit. I can imagine someone might start with a leading `pass:[<]`, but I think if they're doing that, then they're on their own.

mojavelinux:
> I've added a condition so that content that does not start with a < will be surrounded by a <p>, to avoid situations where multi-paragraph quotes would be doubly surrounded by <p>, although, unless non-html < symbols are always replaced with &lt; then there may be edge-cases where this is performed incorrectly.

This is not the correct approach. A converter should never do analysis on the generated HTML itself. The information needed here is already available by consulting the content model of the node. If the content model is simple, there will be no `<p>` tag in the generated HTML. Otherwise, there will be a block tag. That's why I pointed to this helper: https://github.com/asciidoctor/asciidoctor-epub3/blob/main/lib/asciidoctor-epub3/converter.rb#L1326-L1328

Babbz86 (approved):

PR #4569, opened 2024-03-26
remove reference to skins project

geoffcline:
it seems the domain for hosting the previews has gone down, and the repo hasn't been updated in 4 years


mojavelinux:
The preview site may be down for that project, but the project itself is still useful. The default stylesheet has been stable for a long time, so I expect that most of these themes will still work. So I don't see any reason to remove this reference at this time.

PR #4600, opened 2024-07-01
resolves #3545 warn if closing endif directive is not found

### LINKED ISSUE #3545: Detecting improperly terminated conditionals (feature request)
We have some very heavily conditionalized documents, and it's not uncommon for contributors to improperly terminate a conditional in a way that makes the rest of the document  build incorrectly as a result (for example, using ifdef::condition[] where endif::condition[] is needed, or just leaving out the endif:: completely). This seems like something straightforward for asciidoctor to detect at end of processing - presumably it's got some sort of stack of conditions and if the stack isn't empty at document end, something is wrong in the markup. Is there some way to do this already, or is it a feature request there would be any appetite for? Ideally asciidoctor -v would emit an ERROR, or at least a WARN, about this. Ideally it could point back to where in the input the outstanding conditionals began, too.

Originally raised in https://discuss.asciidoctor.org/Detecting-improperly-terminated-conditionals-td7510.html , turned into a feature request her per @mojavelinux .

ianw:
It's quite easy to leave out an endif and then have the rest of the document accidentally be skipped.  This implements a small check when skipping lines in the pre-processor; if we are currently in skipping mode, and we can see the end of the line stack, and it doesn't include an endif, give the user a warning that they've forgotten to close their conditional.

Closes: https://github.com/asciidoctor/asciidoctor/issues/3545

### DIFF
--- a/test/reader_test.rb
+++ b/test/reader_test.rb
@@ -2902,6 +2902,27 @@ class ReaderTest < Minitest::Test
         assert_equal 'before', doc.blocks[0].source
         assert_equal 'after', doc.blocks[1].source
       end
+
+      test 'should warn if unclosed if' do
+        input = <<~EOS
+
+        ifdef::flag[]
+        closed ifdef
+        endif::[]
+
+        ifdef::flag[]
+        unclosed ifdef
+
+        EOS
+
+        using_memory_logger do |logger|
+          doc = Asciidoctor::Document.new input, attribues: {'flag' => 'true'}
+          reader = doc.reader
+          lines = []
+          lines << reader.read_line while reader.has_more_lines?
+          assert_message logger, :WARN, '<stdin>: line 7: Unclosed conditional at end of parsing', Hash
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


mojavelinux:
I had completely forgotten about that issue, but it looks like I had agreed it would be a reasonable check to add. Thanks for submitting an implementation. I've gone ahead and scheduled it for 2.0.x since it does detect a misconfiguration in the document that people probably want to know about. I'll get back to you if any changes are needed.

mojavelinux:
Altering the behavior of the reader must be done extremely carefully because there are numerous entry points to consider and any changes must be carefully studied and made to minimize any impact on performance.

I have identified that the peek_line method is the more appropriate place for this logic since it is always consulted for any unprocessed line and thus has already precomputed the end of the document. This check should only run at the top of the include stack.

With that said, this PR has been superseded by #4602

ianw:
Thank you, I was quite sure there was a better way to do it lurking there :)

test/reader_test.rb
Patches66, line 1: json

PR #4636, opened 2024-11-06
Add attribute "manvoldir"

hansu:
This adds the attribute `manvoldir` to be used when generating man pages.

Currently if you have multiples names in the NAMES section, there will be generated alias files like
`.so manpage.1`

If you specify `manvoldir` they will look like:
`.so <manvoldir>/manpage.1`

This is especially helpful when migrating from a2x to asciidoctor.

See also https://asciidoctor.zulipchat.com/#narrow/channel/279642-users/topic/Man.20page.20generation.3A.20Setting.20base.20path.20for.20alias.20files

This is more a template as I don't really know what is all needed to add a new attribute.


hansu:
This should resolve #4637

PR #4639, opened 2024-12-04
[HTML5] Add icon set support when icons=font

### LINKED ISSUE #3694: Add support for set attribute to icon macro

ViToni:
Resolves #3694.


ViToni:
@mojavelinux Could you take a look at it?

PR #4644, opened 2025-01-07
Convert `nil_or_empty?` core_ext to a refinement

HParker:
This limits the scope of the method addition to this gem preventing it effecting code outside of the gem. This is hopefully less suprising for gem adopters. Refinements are available since ruby 2.7 which is the minimum required Ruby version.

### DIFF
--- a/lib/asciidoctor/abstract_block.rb
+++ b/lib/asciidoctor/abstract_block.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module Asciidoctor
+using NilOrEmptyRefinement
 class AbstractBlock < AbstractNode
   # Public: Get the Array of {AbstractBlock} child blocks for this block. Only applies if content model is :compound.
   attr_reader :blocks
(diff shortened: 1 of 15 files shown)


mojavelinux:
While I agree refinements could be used for this, the `.nil_or_empty?` method is used through the Asciidoctor ecosystem of gems, and I'm not going to try to figure out what the migration path is for all those gems to work together having changed this in core. The support cost for me is simply too high.

And, frankly, I think Ruby should provide this method. It's extremely annoying that it does not. To me, it's part of Ruby, and I want to keep this patch as it stands.

HParker:
I did a little looking around. It seems there was a proposal of a similar idea 5 years ago here: https://bugs.ruby-lang.org/issues/17330

I am not sure I see a new justification good enough to try to reopen the discussion, but it does seem like it is common to want a method that does something like this. I'll keep thinking about it. It is commonly implemented on gems on GitHub: https://github.com/search?q=%22nil_or_empty%3F%22+language%3ARuby&type=code I hope we don't have two different definitions of `nil_or_empty?` from different gems 😬.

I also updated my comment about YJIT. I am wrong about how this works with the JIT. I do still think it would preferred to not have this method globally without asking for it, but appreciate that it is a readability improvement in this gem and in other places.

PR #4647, opened 2025-01-10
Add `logger` as a runtime dependency

Earlopain:
On Ruby 3.4, requiring `logger` without it being in the bundle shows the following warning:

```rb
require "asciidoctor"
/home/earlopain/Documents/asciidoctor/lib/asciidoctor/logging.rb:3: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
=> true
irb(main):002> RUBY_VERSION
=> "3.4.1"
```

### DIFF
--- a/asciidoctor.gemspec
+++ b/asciidoctor.gemspec
@@ -34,6 +34,8 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
   #s.test_files = files.grep %r/^(?:features|test)\/.+$/
 
+  s.add_dependency "logger"
+
   # concurrent-ruby, haml, slim, and tilt are needed for testing custom templates
   s.add_development_dependency 'concurrent-ruby', '~> 1.1.0'
   s.add_development_dependency 'cucumber', '~> 3.1.0'


mojavelinux:
We have prided ourselves on having 0 dependencies in Asciidoctor. It infuriates me that the Ruby project is forcing this situation on existing projects by extracting stdlib into RubyGem dependencies.

I DO NOT want to add a mandatory dependency to Asciidoctor. It adds too many complications. The only way I see this going is replacing the logging facility in Asciidoctor with an internal implementation. Given how the logger is used, I don't think it would be terribly difficult to do so. We can just implement the most basic functionality of logging to a stream or file.

Earlopain:
Infuriating is one way to feel about the bundled gems, yes. I'm certain there will be more of this with the next versions as well.

Since I'm not at all familiar with the project, please feel free to close it and make an issue instead. For reference, `yard` also has 0-dependency as a goal and this is how they solved it: https://github.com/lsegal/yard/commit/a589b9a9af04398b977e328bbd37694672dd3ef7. Does not seem too complicated.

mojavelinux:
> I'm certain there will be more of this with the next versions as well.

And I will continue to fight it by removing use of any library that gets labeled as a dependency. Ruby is destroying its user base by doing this and I hope they are content with that.

mojavelinux:
> For reference, yard also has 0-dependency as a goal and this is how they solved it: lsegal/yard@a589b9a. Does not seem too complicated.

Thanks for the reference. That is indeed useful. We will likely do the same.

Yedfree (approved):

Earlopain:
Since this will not be the solution, I'm just going to close this

mojavelinux:
Indeed, we will resolve this, just not by adding a dependency. We have a long-standing policy to not add any required dependencies to Asciidoctor.

fliiiix:
Is there any progress on this? I just played around with ruby 3.5.0 where logger is gone and stumbled over the same issue

mojavelinux:
Progress will be tracked in #4684, not here. This issue is closed. I don't have time to work on this at the moment, but someone is welcome to contribute a solution to help it move forward.

PR #4667, opened 2025-03-30
feat: semantic HTML 5 table generator

ValorNaram:
My contribution to a new modern semantic HTML 5 table generator for the Asciidoctor and the AsciiDoc community. This is my first time contributing to this project.

ToDos before marking this PR as ready:

* [ ] writing tests
* [ ] writing code for the new table generator
* [ ] testing code using the written tests
* [ ] tweaking it
* [ ] Optional: Putting an html attribute `data-header` to each non-header cell to make [Responsive Data Tables - CSS-Tricks](https://css-tricks.com/responsive-data-tables/) with their example [Responsive Data Tables Example](https://codepen.io/team/css-tricks/pen/wXgJww) possible. Only with some modifications to remove the need to insert row content via CSS. So instead of `content: "First Name";` it will be something like `content: attr(data-header);` for all CSS `td` blocks.

### DIFF
--- a/test/fixtures/semantic-html5-scenarios/table-basic.adoc
+++ b/test/fixtures/semantic-html5-scenarios/table-basic.adoc
@@ -1,0 +1,5 @@
+|===
+|Cell in column 1, row 1|Cell in column 2, row 1
+|Cell in column 1, row 2|Cell in column 2, row 2
+|Cell in column 1, row 3|Cell in column 2, row 3
+|===
--- a/test/fixtures/semantic-html5-scenarios/table-basic_implicit-header-row.adoc
+++ b/test/fixtures/semantic-html5-scenarios/table-basic_implicit-header-row.adoc
@@ -1,0 +1,7 @@
+|===
+|Cell in column 1, row 1|Cell in column 2, row 1
+
+|Cell in column 1, row 1|Cell in column 2, row 1
+|Cell in column 1, row 2|Cell in column 2, row 2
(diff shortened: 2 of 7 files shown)


ValorNaram:
I uploaded the hopefully final HTML table generation code and are open for review comments by the maintainers.

What the `convert_table` code does:

1. It generates an HTML table out of AsciiDoc table. Here I used many codelines from the current html5 backend with some minor tweaks towards accessibility as the old code was not bad in this regard.
2. At generation time it calculates and bakes some HTML attribute and in case of `rowspan` and `colspan` also some hidden cells into the output. This allows CSS or JS to change the layout of the table at users preference.

## The `data-header` algorithm

This algorithm bakes some HTML attributes and in some cases extra cells into the output to make table simplification for complex tables via CSS or JS possible. See [responsive tables from CSS-Tricks](https://css-tricks.com/responsive-data-tables/) for the overall idea.

image

This image shows a complex data table with nested header rows, cells spawning multiple rows and columns and many columns. With a tick of an checkbox it gets translated to the following via JS:

image

[responsive-table-algorithm-demo-.zip](https://github.com/user-attachments/files/21005534/responsive-table-algorithm-demo-.zip) is a zip archive consisting of more examples. These are possible because the `data-header` algorithm baked into `convert_table` does two things:

1. Calculating the row- and column level headers for each cell and assigning to `data-column-header` and `data-row-header` attributes of each non-spawning and non-header cell.
2. Creating shadow cells for each `rowspan` and `colspan`.

PR #4672, opened 2025-04-15
Fix typo in preprocessor extension docs

r0ckarong:
There is a small typo in the preprocessor extension docs.


mojavelinux:
Thanks!

PR #4677, opened 2025-05-02
doc: Improve understanding with using long-format switches

buhtz:


mojavelinux:
Using long-form switches is not the behavior I want to encourage. It's also a lot more that a user has to type. The short options are the preferred form and have been for over 2 decades.

I'm fine with adding a single example that shows the long-form switches for illustration, but I don't want this changed to be made throughout.

I also disagree that it improves the understanding. It just makes the command look more overwhelming. The short form switches are documented in the usage text alongside the long-form switches, and it should not be hard for users to quickly make the association.

mojavelinux:
> That is IMHO a good compromise. Reading long form in the beginning of a help document would be fine for me and had saved me time reading the man page.

Great. This seems like the path forward.

> But where? In the man page. 

Just like almost every CLI command, the asciidoctor command has a help switch:

```
asciidoctor --help
```

See https://docs.asciidoctor.org/asciidoctor/latest/cli/#help-topics

>  I have to open a documentation object (the man page) to understand another documentation object (the tutorial).

The man page is also in the documentation. See https://docs.asciidoctor.org/asciidoctor/latest/cli/man1/asciidoctor/#options

> Not quickly, because the information of interest is not contained in the primary document (the tutorial).

Yes it is. See previous answer.

mojavelinux:
I have updated the following sections:

* https://docs.asciidoctor.org/asciidoctor/latest/cli/#version-and-runtime
* https://docs.asciidoctor.org/asciidoctor/latest/cli/#convert-an-asciidoc-file
* https://docs.asciidoctor.org/asciidoctor/latest/cli/#help-topics
* https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/#generate

PR #4681, opened 2025-05-06
fix front matter in preprocessor.adoc

vvavrychuk:
Front matter must start with `---`, see https://docs.asciidoctor.org/asciidoctor/latest/html-backend/skip-front-matter/.


mojavelinux:
Thanks. That line got inadvertently dropped when the docs were migrated.

PR #4795, opened 2025-10-19
resolves #4568 space out dots that are not meant to be an ellipsis in the man output

### LINKED ISSUE #4568: manpage.rb: space ellipses more pleasantly
https://github.com/asciidoctor/asciidoctor/blob/fd0e92a78b18e974a567b0ef798d99bbbb8c6acc/lib/asciidoctor/converter/manpage.rb#L707

_groff_man_style_(7):
```
   Portability
...
     \|        Thin space (one‐sixth em on typesetters, zero‐width on
               terminals); a non‐breaking space.  Used primarily in
               ellipses (“.\|.\|.”) to space the dots more pleasantly on
               typesetters like dvi, pdf, and ps.
```

The `\|` escape sequence is universally portable to _roff_ implementations; it appeared in Joe Ossanna's CSTR #54 in 1976.

mojavelinux:


g-branden-robinson:
The change log entry would be more comprehensible to me (and perhaps other man page- or *roff-oriented minds) if it said:

```
Put thin space escape sequences amid sequences of three dots that are not intended to become U+2026 in the man output
```

...or similar.  The distinction between an ellipsis's _encoding_ and its _rendering_ is significant.

PR #4833, opened 2025-12-31
Docs: add example for :sectnums: attribute

### LINKED ISSUE #4832: Improve keyboard accessibility for PDF viewer
The current documentation explains attributes such as :sectnums: and :toc:, but for new users it is difficult to understand their effect without concrete examples.

I propose adding short before/after examples showing how these attributes change the rendered output. This would improve accessibility for new users and reduce onboarding friction.

rubenmdsp:
This adds a small example showing how the :sectnums: attribute works.
Closes #4832

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -501,3 +501,17 @@ endif::[]
 
 Refer to the {url-changelog}[CHANGELOG] for a complete list of changes in older releases.
 endif::[]
+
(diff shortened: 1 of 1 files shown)


mojavelinux:
This is not where we add documentation. If you want to add documentation, please use the Edit this Page link in the documentation to find the source associated with the given page.

PR #4834, opened 2026-01-06
Add the logger Gem dependency for Ruby 4.0

igaiga:
In Ruby 4.0, logger became a bundled gem, so please add it to dependencies and install it together.

I'm submitting an example change in this PR.

logger became a bundled gem in Ruby 4.0: https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/#stdlib-updates


igaiga:
I understand. Thank you for your reply. 🙂

PR #4848, opened 2026-04-03
fix BlockAttributeLineRx incorrectly matching paragraph lines ending with a footnote

jpritikin:
…with ]

A paragraph starting with an inline role (e.g. [.smallcaps]#text#) and ending with a footnote macro (e.g. footnote:[...]) was falsely matched by BlockAttributeLineRx because the regex used CC_ANY* which allows ] inside the match. This caused the parser to treat the paragraph line as a block attribute line and break out of paragraph reading with no content, silently dropping the paragraph body in the output (notably in DocBook5).

Fix by replacing CC_ANY* with [^\]]* in both BlockAttributeListRx and BlockAttributeLineRx so that a bare ] cannot appear inside the matched attribute list content.

### DIFF
--- a/lib/asciidoctor/rx.rb
+++ b/lib/asciidoctor/rx.rb
@@ -177,12 +177,12 @@ module Asciidoctor
   #   # as attribute reference
   #   [{lead}]
   #
-  BlockAttributeListRx = /^\[(|[#{CC_WORD}.#%{,"']#{CC_ANY}*)\]$/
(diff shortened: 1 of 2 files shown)


mojavelinux:
The behavior is too the specification of AsciiDoc. Until the language is defined (and this matter is addressed) we are not changing Asciidoctor. If you want to discuss further, you can do so in the project chat.

PR #4852, opened 2026-04-23
Distinguish between extension types & examples

henrebotha:


mojavelinux:
That makes sense. Thanks!

Issue #3748, opened 2020-09-05
Support for in-document SVG content


Oblomov:
Hello all,

my understanding is that it's not currently possible to have an inline SVG in an `.adoc`; at best, under appropriate safety conditions, one may inline an _external_ SVG. What I'm interested in instead would be a way to do something like the following (pure speculation on the syntax, of course):

```
[[someid]]
.caption content
[image]
+++
<svg>blah blah </svg>
+++
```

which would be equivalent to

```
[[someid]]
.caption content
[%inline]
image::someexternal.svg[]
```
with the `<svg>blah blah </svg>`  in `someexternal.svg`.

I've found a partial workaround with custom blocks that do raw passthrough, but the obvious issue is that I then need to recreate also all of the wrapping stuff (caption, div etc), which basically requires me to recreate `convert_image`.

Suggestions?

mojavelinux:
You are correct this is not built-in functionality. This would make a great block extension. You can define a block named `svg`, then in the extension generate an image block that passes the SVG data through via the target (perhaps base 64 encoded)...or the extension could generate a passthrough block, but that's not as ideal for reasons you mentioned.

mojavelinux:
I almost wonder if this would fit with asciidoctor-diagram. You could think of the raw SVG as the source of a "diagram", and it would handle all the necessary details of translating it to an image block.

Oblomov:
Hello @mojavelinux and thanks for your replies.

I did try creating an image block with a data-uri: of the image content, but I wasn't successful. With the default settings, it would complain about `allow-uri-read` being disabled, and after enabling it still failed with the generic warning: `asciidoctor: WARNING: could not retrieve contents of SVG at URI: <mybase64data>` (without an explanation or hint about the reason why). The relevant part of the block processing extension is
```
    attrs['target'] = 'data:image/svg+xml;base64,' + Base64::strict_encode64(svg)
    attrs['format'] = 'svg'
    attrs['inline-option'] = true
    create_image_block parent, attrs
```
but maybe I'm  doing something wrong (still, the need to enable `allow-uri-read` is a bit annoying, although I'm sure there's security reasons for this).

As an aside, to make it work provisionally I'm doing something horrible: live-patching the `convert_image` method to change the first `if` into `if node.content? ; img = node.content ; elsif`, and then using
```
    block = create_block parent, :image, [svg], attrs, content_model: :raw
    if title
      block.title = title
[...]

Issue #3698, opened 2020-07-03
It should be possible to set the contents of an image (bytes) on an Image node


Mogztter:
When using the API it should be possible to set the content of an image on an Image node. It would be useful when the target of the Image node is not directly accessible on the file system or when the image cannot be fetched using a GET request.

For instance, if I'm using an extension that get an image from a remote server using an authenticated POST request, I want to set the contents on the Image node.
In this case, Asciidoctor should use the contents available on the Image node and not try to fetch the content (again) from the target.

LucCappellaro:
IncludeProcessors allow you to implement you own way to fetch the resource anywhere you want.
But this concept is strangely missing on image resources.
The ability to register custom ImageIncludeProcessors would help a lot.

However I find the resource handling in general quite strange.
Why not remove all the filesystem support, an rely solely on URLs (absolute, root-relative and relative)?
Because the filesystem itself is also an URL (file:///).
In order to extend the resource handling, you could allow to register custom protocol handlers.
(see [URLStreamHandler](https://docs.oracle.com/javase/8/docs/api/java/net/URLStreamHandler.html))

Mogztter:
> IncludeProcessors allow you to implement you own way to fetch the resource anywhere you want.
But this concept is strangely missing on image resources.
The ability to register custom ImageIncludeProcessors would help a lot.

That's related but not entirely.
In fact, the image can be generated (for instance, using a diagram library) and not referenced by a URI. The idea here is to create an Image node with its contents using the Asciidoctor API.

dsisnero:
+1
would be good for all the extensions that generate png files

Issue #3657, opened 2020-05-16
Generate the image uri from the mimetype and bindata attributes on the image Node (if present)


Mogztter:
If the `mimetype` and `bindata` attributes are present on an image Node, Asciidoctor should not try to read the image data from the target of the image Node.
Instead, we should use these attributes to generate the image URI:

```rb
if (mimetype = attr 'mimetype') && (bindata = attr 'bindata')
  %(data:#{mimetype};base64,#{::Base64.strict_encode64 bindata})
else
  # ...
end
```

mojavelinux:
Wouldn't it be better if you just precomputed the data URI and assigned it to the target? Then core would just need to make sure it doesn't mess with the target if it's a data URI (which I think it already does).

Mogztter:
If it does that's even better!
I read the code and didn't find any indication that it was the case, I will add a test to check.

Mogztter:
It's working but it's a bit unexpected! That's why I missed it.
The code will actually produce an exception and a warn message:
```
could not retrieve image data from URI: data:image/svg+xml;base64,...
```

The exception is: 
```
undefined method `buffer_open' for #<URI::Generic:0x000055e0ca3a6e88>
```
The reason is that we are trying to "open" the data URI using `OpenURI`:

https://github.com/asciidoctor/asciidoctor/blob/8aadc1412267bf161f911e6c4800d08fa3d9263e/lib/asciidoctor/abstract_node.rb#L401-L412

I don't think we should rely on an exception from `open_uri`. We are also doing unnecessary work.
I believe that `data:` as a scheme is a clear signal that the target is a data URI and as a result we should not even try to "open" it but return as-is.

I can make that change if you want?

mojavelinux:
I'm pretty sure image_uri already handles the data URI without relying on allow-uri-read to be set or open_uri. We even have a test for it.

```
image::data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=[Dot]
```

The problem seems to be that if both data-uri and allow-uri-read are set, then the logic breaks.

We may need the logic from Asciidoctor PDF. See https://github.com/asciidoctor/asciidoctor-pdf/blob/master/lib/asciidoctor/pdf/ext/asciidoctor/image.rb#L13-L19 But that might be too heavy for what we need. Perhaps just fix the logic bug to pass through the data URI in this case.

mojavelinux:
[...]

PR #4865, opened 2026-06-25
resolves #3791 add support for data URI image targets

### LINKED ISSUE #3791: Add support for block image target as data URI
There are various code paths for which a block image macro with a data URI target is not supported. One case is when both allow-uri-read and data-uri are set. Another one #3764. When the target of a block image macro is a data URI, that URI should be preserved, processed if necessary, and passed through to the output. (We should also consider whether this support should be extended to inline image macros).

Example:

```
image::data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAwAAAAMCAIAAADZF8uwAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAB3RJTUUH4wUbCB4vYn/PXAAAAB1pVFh0Q29tbWVudAAAAAAAQ3JlYXRlZCB3aXRoIEdJTVBkLmUHAAAAE0lEQVQY02NcxUAYMDGMKhqEigA8yQDCfg5F4QAAAABJRU5ErkJggg==[red square]
```

To support an SVG as a data URI, it may be necessary to specify the format explicitly.

```
image::data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHZpZXdCb3g9IjAgMCAyMDAgMjAwIj48cmVjdCB3aWR0aD0iMjAwIiBoZWlnaHQ9IjIwMCIgZmlsbD0iI2ZmMDAwMCI+PC9yZWN0Pjwvc3ZnPg==[red square,format=svg]
```

This requirement could be avoided if we look for and parse the data URI up front when looking for the image format.

For most images, the processor can assume the data is base64 encoded. However, for SVG images, the data can be passed through in raw XML. In this case, the data should be base64 encoded so it's properly escaped in the output.

When the format is SVG and opts=inline, the data should be read from the data URI, base64 decoded if necessary, and emitted as an embedded SVG.

ggrossetie:
Toolchains that produce embedded diagrams (e.g., when `data-uri` is set, or extensions that emit a data-URI target) end up with the image carried in the target itself. Raster images already worked because they pass straight through to ``, but an SVG that needs to be embedded inline went through `read_contents`, which only understands local files and remote `http(s)` URIs, so it always degraded to alt text. This rounds out `data:` URI support for image targets.

Add support for a `data:` URI as the target of an image:

1. **Inline SVG embedding** — when the `inline` option is set, the SVG carried in the target is decoded and emitted as inline `<svg>` instead of falling back to the alt text. Both Base64 (`data:image/svg+xml;base64,...`) and percent-encoded (`data:image/svg+xml,...`) payloads are supported.
2. **Format inference** — the SVG format is inferred from the `image/svg+xml` media type, so `format=svg` is no longer required on a `data:` URI target.
3. **Pass-through without a spurious warning** — a non-SVG `data:` URI target (e.g., `data:image/png;base64,...`) is passed through unchanged to the ``; previously, with `data-uri` and `allow-uri-read` both set, Asciidoctor would attempt to fetch the `data:` URI via OpenURI and log a "could not retrieve image data from URI" warning.

#### Example

```asciidoc
image::data:image/svg+xml;base64,PHN2ZyB4bWxucz0i...[Diagram,opts=inline]
```

renders the `<svg>…</svg>` inline instead of an `` fallback.

#### Implementation details

- **Inline SVG decoding** — `Html5Converter#read_svg_contents` branches on `target.start_with? 'data:'` and decodes the payload via a new private `decode_data_uri` helper instead of calling `read_contents`. `decode_data_uri` decodes Base64 with `unpack1 'm'` (the decode counterpart of the `pack 'm0'` already used by `generate_data_uri`, so no dependency on the `base64` gem) and forces UTF-8 on the result; percent-encoded payloads go through a new `Helpers.decode_uri_component`.
- **`Helpers.decode_uri_component`** — added as the counterpart to `encode_uri_component`, with the same `RUBY_ENGINE == 'opal'` / `CGI.unescapeURIComponent` / fallback structure. The fallback preserves a literal `+` rather than turning it into a space, matching JavaScript's `decodeURIComponent` (and `encodeURIComponent` round-trips through it).
- **Format inference** — the SVG-detection condition in `convert_image` and `convert_inline_image` also matches a target that starts with `data:image/svg+xml`.
- **Pass-through guard** — `AbstractNode#image_uri` returns a `data:` URI target as-is up front, so it is never read as a file nor fetched via OpenURI.

#### Security

Decoding a `data:` URI reads neither the filesystem nor the network — the bytes are already part of the document — so it does not bypass the safe-mode guarantees that `read_contents` / `normalize_system_path` enforce for local files. The tests run under `safe: :safe`.

#### Tests

- `test/blocks_test.rb`:
  - inline SVG from a Base64 data URI and from a percent-encoded data URI (asserts the `<svg>` is embedded with the expected width and no `` fallback);
  - SVG format is inferred from the media type without an explicit `format` attribute;
  - a `data:image/png` target is passed through to `` unchanged and logs no warning even with `data-uri` + `allow-uri-read`.
- `test/helpers_test.rb`: `decode_uri_component` decodes percent-encoded octets, preserves a literal `+`, and round-trips with `encode_uri_component`.

Resolves #3791

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -33,6 +33,7 @@ Enhancements::
   * Don't add empty line before block content if principal text of dlist item is not specified when converting to man page (#4182)
   * Add support for role on thematic break in HTML converter (#4101)
   * Add support for link=self on image macros (#3656)
+  * Support a `data:` URI as the target of a block or inline image; embed inline SVG (Base64 or percent-encoded) when the `inline` option is set and infer the SVG format from the `image/svg+xml` media type (#3791)
   * Alias Inline#content to Inline#text so Inline node quacks like other nodes (#3220)
   * Add support for skipping TOML front matter (Hugo) when `skip-front-matter` attribute is set (#4300) (*@abhinav*)
   * Add support for Wistia using the video macro
@@ -76,6 +77,7 @@ Improvements::
 Bug Fixes::
 
   * When skip-front-matter document attribute is set, only skip front matter in primary document, not included content (#3437)
+  * Don't attempt to read a `data:` URI image target as a file or remote URI when `data-uri` and `allow-uri-read` are set (avoids a spurious "could not retrieve image data from URI" warning) (#3791)
   * When normalizing toclevels, detect part even when book starts with special section (#4814)
   * Add parts at correct level in TOC even when book starts with special section (#2262)
 
--- a/lib/asciidoctor/abstract_node.rb
+++ b/lib/asciidoctor/abstract_node.rb
@@ -302,7 +302,8 @@ class AbstractNode
 
   # Public: Construct a URI reference or data URI to the target image.
   #
-  # If the target image is a URI reference, then leave it untouched.
+  # If the target image is a URI reference, then leave it untouched. If the target image is a data
+  # URI, then it is already an embedded image, so it is returned as-is.
   #
   # The target image is resolved relative to the directory retrieved from the
   # specified attribute key, if provided.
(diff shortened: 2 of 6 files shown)


ggrossetie:
This also addresses #3657, #3748, and #3698, which were circling around the same underlying problem (handling data: URIs as image targets/content).
From the discussion:

- On #3657, @mojavelinux noted that the right solution is to always support a `data:` URI as the target, and linked it directly to #3791.
- On #3748, @Oblomov concluded that fixing `data:` URI handling would essentially solve their inline-SVG case, and possibly #3698 too.
- On #3698, I made a similar observation, that supporting a `data:` URI target would let you create an image block with the actual SVG content without re-fetching it.

Fixing it here should close them out as well 🎉

